PR #6, opened 2021-11-17
TNO-52 Add Syndication Ingest Service

Fosol:
There are now two shared java libraries, `dal-db` and `core`. 

There are now up to three new containers to host the ATOM, RSS, and NLP services.

## Warning

Many of the `.env` files have been moved and will need to be regenerated.

The database has been refreshed, which will require you to drop the objects and rerun the `make db-update` within the DAL DB container.  At some point I need to make a more intuitive process.

If you want to clean up your local environment you should delete all your existing `.env` files and then regenerate them.  Just be aware that this will create a new password, which won't align with any existing containers.  So you can either remove all containers and volumes, or make sure all the new `.env` files have the correct password.

## Summary

- Fix NLP service
- Rename folders
- Update documentation
- Fix docker compose
- Add data source config
- Update make
- Fix DAL DB package
- Fix setup scripts
- Reduce dependencies
- Add core library package


db/kafka/docker-compose.yml
```diff
@@ -4,6 +4,7 @@ version: '3.5'
 services:
   zookeeper:
     image: tno:kafka-zookeeper
+    profiles: ["kafka"]
```
Fosol, line 7: Using profiles now to separate different related container groups.

PR #9, opened 2021-11-18
Update Makefile

ckayfish:
Specified all profiles for docker-compose "up:". When a profile is specified for a service in the compose YAML, it will not start unless that profile is specified. Would prefer a solution that starts all profiles without specifying them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ build: ## Builds all containers or the one specified (n=service name)
 
 up: ## Starts all containers or the one specified (n=service name)
 	@echo "$(P) Starting all containers..."
-	@docker-compose --env-file .env -f docker-compose.yml -f docker-compose.override.yml -f ./db/kafka/docker-compose.yml up -d $(n)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -55,7 +55,7 @@ build: ## Builds all containers or the one specified (n=service name)
 
 up: ## Starts all containers or the one specified (n=service name)
 	@echo "$(P) Starting all containers..."
-	@docker-compose --env-file .env -f docker-compose.yml -f docker-compose.override.yml -f ./db/kafka/docker-compose.yml up -d $(n)
+	@docker-compose --env-file .env -f docker-compose.yml -f docker-compose.override.yml -f ./db/kafka/docker-compose.yml --profile service --profile core --profile kafka --profile utility up -d $(n)
```
Fosol, line 58: I'm still sorting this part out.  Found out about profiles this week.  The main issue I was attempting to solve is that some of the containers will fail if they start too early.  The `depends_on` isn't the only part that needs to be checked.  So I'm leaning towards only including the `core` and `kafka` services as part of the default `make up`.  Then at least the developer can wait a minute before running `make service-up`.  Or at least that's my thoughts presently.

Fosol (approved):

PR #12, opened 2021-11-23
TNO-5 Database build/deploy templates

Fosol:
## Summary

Add build and deployment configuration templates to provide a way to get the PostgreSQL database running in Openshift.

The official Docker Hub images are not supported by Redhat and can't run within Openshift due to security requirements.  The official images provided by Redhat regrettably have both differences and issues, which complicate their use locally.  For now I've made it possible to use either locally, but it requires additional steps.  This has been documented.

The database deployment config also includes a persistent volume claim (PVC) that is hosted on a file store that does backup, but the database itself is not stored on this PVC.  

## TODO (another PR)

- Database backup configuration
- Database migration process


Makefile
```diff
@@ -51,19 +51,19 @@ setup: ## Setup local environment for development, generate configuration files.
 
 build: ## Builds all containers or the one specified (n=service name)
 	@echo "$(P) Building images..."
-	@docker-compose -f docker-compose.yml -f docker-compose.override.yml -f ./db/kafka/docker-compose.yml build --no-cache $(n)
+	@docker-compose -f docker-compose.yml -f docker-compose.override.yml -f ./db/kafka/docker-compose.yml --profile core --profile kafka build --no-cache $(n)
```
Fosol, line 54: We only want the core and Kafka services to startup by default.  The others are not critical for development, and some of the utilities will fail startup if they are not done manually after everything else is running.

Makefile
```diff
@@ -158,13 +158,21 @@ npm-refresh: ## Removes and rebuilds node app containers, images, volumes.
...
 ##############################################################################
-# Flyway Database Migration Commands
+# Database Commands
 ##############################################################################
 
 db-update: ## Run the flyway database migration update (requires maven to be installed).
 	@echo "$(P) Run the flyway database migration update"
 	@cd libs/java/dal/db; make db-update;
 
+db-refresh: ## Stop and delete the database container and volume, then rebuilds and starts.
```
Fosol, line 168: Makes it quicker to refresh the database.  Keycloak also uses this database and must be regenerated.

docker-compose.yml
```diff
@@ -18,21 +18,21 @@ services:
...
+
+  ## Redhat Image
   database:
     image: tno:database
     profiles: ["core", "data"]
     restart: on-failure:1
     container_name: tno-database
     build:
-      context: db/postgres
+      context: ${DB_CONTEXT:-db/postgres/docker}
```
Fosol, line 29: Environment variables to enable quickly switching back and forth.

PR #14, opened 2021-11-26
TNO-165 Add DB migration

Fosol:
To improve developer local environment setup, scripts have been updated and a `make db-update` command has been added.  Developers can now initialize their TNO database and run future migrations.

DevOps process isn't complete, but the same process can be applied within Openshift through building the `dal-db-migration` image and running it as a Job.  I will look into other CI/CD options to perform this.

There is a new `make init` command that will now perform initializing a new environment.

## Warning

A number of updates were made to `.env` file locations.  The `make setup` script will generate them, however it would be best to reset your environment entirely by deleting all of your containers and images and existing `.env` files.  This makes me think I may need to start providing migration scripts for this, but I really don't want to spend extra time on it presently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,12 @@ setup: ## Setup local environment for development, generate configuration files.
 	@mkdir -p ./app/editor/node_modules
 	@mkdir -p ./app/subscriber/node_modules
 
+init: ## Initialize your local environment and start the core solution.
+	@echo "$(P) Initialize your local environment and start the core solution"
+	@make setup
+	@make up
+	@make db-update
+
 ##############################################################################
 # Docker Management
 ##############################################################################
@@ -163,7 +169,7 @@ npm-refresh: ## Removes and rebuilds node app containers, images, volumes.
 
 db-update: ## Run the flyway database migration update (requires maven to be installed).
 	@echo "$(P) Run the flyway database migration update"
-	@cd libs/java/dal/db; make db-update;
+	@./tools/scripts/db-update.sh
 
 db-refresh: ## Stop and delete the database container and volume, then rebuilds and starts.
 	@echo "$(P) Stop and delete the database container and volume, then rebuilds and starts"
--- a/libs/java/dal/db/dal-db-migration/Dockerfile
+++ b/libs/java/dal/db/dal-db-migration/Dockerfile
@@ -1,0 +1,19 @@
+FROM maven:3.6.3-openjdk-17 AS base
+
+WORKDIR /project
+COPY . .
+RUN mvn package
+
+FROM openjdk:17.0.1-slim AS final
+
+WORKDIR /app
+COPY --from=base /project/target .
+
+RUN apt-get update && apt-get install curl -y
+
+EXPOSE 8080
+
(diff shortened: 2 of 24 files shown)


Makefile
```diff
@@ -45,6 +45,12 @@ setup: ## Setup local environment for development, generate configuration files.
 	@mkdir -p ./app/editor/node_modules
 	@mkdir -p ./app/subscriber/node_modules
 
+init: ## Initialize your local environment and start the core solution.
```
Fosol, line 48: The new "let's get started developing" command.

libs/java/dal/db/dal-db-migration/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM maven:3.6.3-openjdk-17 AS base
```
Fosol, line 1: This image is so we can run the database migration as a docker container.  This provides us a way to run it in Openshift and locally.

libs/java/dal/db/dal-db-migration/src/main/java/ca/bc/gov/tno/dal/db/MigrationApp.java
```diff
@@ -0,0 +1,26 @@
...
+
+  public static void main(String[] args) {
+    var app = new SpringApplication(MigrationApp.class);
+    app.setWebApplicationType(WebApplicationType.NONE);
+    app.run(args);
+    System.exit(0);
+  }
+
+  @Override
+  public void run(String... args) throws Exception {
```
Fosol, line 23: Doesn't do much be we need a main class to run for the migration to work.

libs/java/dal/db/dal-db/src/main/java/ca/bc/gov/tno/dal/db/services/ClaimService.java
```diff
@@ -18,8 +18,18 @@
...
+
+  /**
+   * Creates a new instance of a ClaimService object, initializes with specified
+   * parameters.
+   * 
+   * @param repository The claim repository.
+   */
   @Autowired
-  private IClaimRepository repository;
+  public ClaimService(final IClaimRepository repository) {
```
Fosol, line 30: Switch dependency injection to use constructors instead of properties.

openshift/templates/jobs/db/build.yaml
```diff
@@ -0,0 +1,116 @@
+kind: Template
```
Fosol, line 1: Build the db migration image

openshift/templates/jobs/db/deploy.yaml
```diff
@@ -0,0 +1,127 @@
+apiVersion: template.openshift.io/v1
```
Fosol, line 1: Run the db migration container as a job

Makefile
```diff
@@ -163,7 +169,7 @@ npm-refresh: ## Removes and rebuilds node app containers, images, volumes.
 
 db-update: ## Run the flyway database migration update (requires maven to be installed).
```
Fosol, line 170: You can now run the migration from the root of the repo without opening additional containers.

sturufous (approved):

PR #22, opened 2021-12-09
nuke.sh updated

ckayfish:
- Added "*.bak" to .gitignore. All other changes in nuke.sh
- Only proceed to nuke if there are 1 or more existing environment/config files
- Ask user if they want to delete existing backups
- Move all environment/config files using pattern \<oldname\>.\<timestamp\>.bak


Fosol (approved):
Now that's a `nuke`!  Way better than mine.

PR #23, opened 2021-12-09
TNO-141 Add database backup

Fosol:
We now have nightly backups of the database running in DEV!

## Summary

- Fixed port number configuration
- Add build and deploy templates for the database backup pod
- Ability to schedule backups based on configuration
- Ability to create backups, or restores ad-hoc through scripts

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,6 @@
 {
-  "git.ignoreLimitWarning": true
+  "git.ignoreLimitWarning": true,
+  "yaml.schemas": {
+    " https://raw.githubusercontent.com/garethr/openshift-json-schema/master/v4.1.0/_definitions.json": ["/*"],
+  }
 }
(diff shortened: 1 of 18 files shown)


.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-  "git.ignoreLimitWarning": true
+  "git.ignoreLimitWarning": true,
+  "yaml.schemas": {
+    " https://raw.githubusercontent.com/garethr/openshift-json-schema/master/v4.1.0/_definitions.json": ["/*"],
```
Fosol, line 4: This fixed the YAML plugin if you have it installed so that it doesn't display errors for yaml files.  Still don't think the plugin works well for intelli-sense however.

openshift/templates/backup/README.md
```diff
@@ -0,0 +1,42 @@
...
+Example of a restore.
+
+```bash
+./backup.sh -r postgres=tno-database:5432/tno -f tno-database-tno_2021
+```
+
+If you want a single command to SSH and perform a backup use the following syntax.
+
+```bash
+oc rsh $(oc get pod -l name=tno-backup --no-headers -o custom-columns=POD:.metadata.name) ./backup.sh -l
```
Fosol, line 41: This is a nice one-liner, which we can use to automate future scripts

sturufous (approved):

PR #24, opened 2021-12-09
updated scripts

ckayfish:
- Suppressed grep errors in `variables.sh` for files not expected to exist
- Suppressed grep errors in `gen-maven-files.sh`, and fixed bug where looking for postgress .env file in wrong location
- Now using backup file extension *.nkbak* in `nuke.sh` & `.gitignore` so as not to interfere with others
- **Note:** Run `find . -type f -name "*.bak" -delete` to remove backup files from previous version
- Cleaned up nuke.sh

### DIFF
--- a/tools/scripts/gen-maven-files.sh
+++ b/tools/scripts/gen-maven-files.sh
@@ -7,7 +7,7 @@ echo "*********************************"
 
 . ./tools/scripts/variables.sh
 
-ossrhUsername=$(grep -Po '<username>\K[^\<]+' ./libs/java/dal/db/.devcontainer/maven-settings.xml)
+ossrhUsername=$(grep -Po '<username>\K[^\<]+' ./libs/java/dal/db/.devcontainer/maven-settings.xml 2>/dev/null)
 if [ -z "$ossrhUsername" ]
 then
     echo 'Enter your OSSRH Maven Central username'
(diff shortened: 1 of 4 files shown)


tools/scripts/gen-maven-files.sh
```diff
@@ -7,7 +7,7 @@ echo "*********************************"
 
 . ./tools/scripts/variables.sh
 
-ossrhUsername=$(grep -Po '<username>\K[^\<]+' ./libs/java/dal/db/.devcontainer/maven-settings.xml)
+ossrhUsername=$(grep -Po '<username>\K[^\<]+' ./libs/java/dal/db/.devcontainer/maven-settings.xml 2>/dev/null)
```
Fosol, line 10: What does this do?

tools/scripts/gen-maven-files.sh
```diff
@@ -25,9 +25,9 @@ else
     echo "OSSRH Maven Central password: $ossrhPassword"
 fi
 
-dbUser=$(grep -Po 'POSTGRES_USER=\K.*$' ./db/postgres/.env)
-password=$(grep -Po 'POSTGRES_PASSWORD=\K.*$' ./db/postgres/.env)
-dbName=$(grep -Po 'POSTGRES_DB=\K.*$' ./db/postgres/.env)
+dbUser=$(grep -Po 'POSTGRES_USER=\K.*$' ./db/postgres/docker/.env)
```
Fosol, line 28: Thanks for fixing

Fosol (approved):

tools/scripts/gen-maven-files.sh
ckayfish: Sends stderr (2) to /dev/null (rather than screen) so error doesn't appear on the terminal saying grep cant find the file that it's trying to read the variable from. We know the file/variable doesn't exist when it prompts for it.

PR #29, opened 2022-01-05
TNO-126 Add Tekton pipeline

Fosol:
These templates provide a CI/CD pipeline that uses Tekton and native Openshift BuildConfig and DeployConfig objects.  GitHub webhooks are configured on `push` and `pull-request`.  

A `push` event on the `dev` branch will trigger a PipelineRun when the `app/editor` context has been modified.

## Future Enhancements

Currently it doesn't handle versioning.  The templates will need to be updated to listen for a Release to be created which will then trigger a build and deploy.

I was unable to get Buildah to work with the Redhat image registry.  Much of this we can probably simply use GitHub Actions, which would most likely provide us feedback during Pull Requests (which would be better in most cases).

At some point unit-tests, code scanning, and image scanning will need to be added to either GitHub Actions or Tekton pipelines.

A pipeline will need to be created for the other services and components.  Which would also require database migration automation too.


.vscode/settings.json
```diff
@@ -1,8 +1,10 @@
 {
   "git.ignoreLimitWarning": true,
   "yaml.schemas": {
-    " https://raw.githubusercontent.com/garethr/openshift-json-schema/master/v4.1.0/_definitions.json": [
-      "/*"
-    ]
-  }
+    "kubernetes": "kubernetes.yml",
```
Fosol, line 4: Fix the vscode yaml plugin

sturufous (approved):

PR #35, opened 2022-01-21
Fix maven gpg signing

Fosol:
The fix to run with `sudo` won't work with containers in Windows VM.  The default behaviour is now for Windows VM, but there is now an override that can be used through parameters.

`make setup u=sudo`
`make deploy-release u=sudo`

### DIFF
--- a/libs/java/dal/db/scripts/setup.sh
+++ b/libs/java/dal/db/scripts/setup.sh
@@ -1,10 +1,21 @@
 #!/bin/bash
 
-# This fixes a bug with GPG when signing the package.
-# export GPG_TTY=$(tty)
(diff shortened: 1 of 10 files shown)


libs/java/dal/db/scripts/setup.sh
```diff
@@ -1,10 +1,21 @@
...
 # This is required to sign and verify packages.
-# sudo env "PATH=$PATH" gpg --import private.key
-sudo gpg --import private.key
\ No newline at end of file
+# sudo env "PATH=$PATH" gpg --import private.key]
+FILE=private.key
+if [ -f "$FILE" ]; then
+  if [ "$1" = "sudo" ]; then
+    sudo gpg --import "$FILE"
+    sudo cp .devcontainer/gpg.conf ~/.gnupg/gpg.conf
```
Fosol, line 10: I'm not 100% sure this works yet.

sturufous (approved):

PR #48, opened 2022-02-15
TNO-136 Add DevOps for Syndication

Fosol:
I've bumped the DAL-DB to `0.0.19-SNAPSHOT` to deal with the prior snapshot from being overwritten.


api/editor/api-editor/pom.xml
```diff
@@ -55,7 +55,7 @@
     <dependency>
       <groupId>ca.bc.gov.tno</groupId>
       <artifactId>dal-db</artifactId>
-      <version>0.0.18-SNAPSHOT</version>
+      <version>0.0.19-SNAPSHOT</version>
```
Fosol, line 58: Bumped to avoid collusion

libs/java/service/src/main/java/ca/bc/gov/tno/services/data/BaseScheduleService.java
```diff
@@ -143,6 +143,13 @@ public void onApplicationEvent(ServiceStartEvent event) {
           if (index == sourceConfigs.getSources().size())
             index = 0;
 
+          if (sourceConfigs.getSources().size() == 0) {
```
Fosol, line 146: Handle issue when database has no data sources for the specified media type.

PR #52, opened 2022-02-18
TNO-210 Fix JWT Refresh Bug

Fosol:
Published tno-core version 0.0.92

The JWT wasn't correctly being refreshed.  Had to add `state` as a dependency to `useSummon` hook.  This however has a negative impact on other components, as they now will want to refresh when `useSummon` does.  This requires not including `useSummon` related components as state dependency values, which results in a linting warning.

I'm not certain how to work around this kind of issue.  `useSummon` is a hook because it needs access to state and it creates the `axios` instance.  However, this means that any time state changes related to JWT it will create a new instance of `axios`.  Which then trickles down to other components using the hook...  Very annoying issues with React state management.


libs/npm/core/src/hooks/summon/useSummon.tsx
```diff
@@ -45,7 +45,8 @@ export const useSummon = ({
           'Access-Control-Allow-Origin': '*',
         },
       }),
-    [],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [baseURL, state],
```
Fosol, line 49: Create a new axios instance when JWT is refreshed.

openshift/templates/tekton/triggers/git-eventlistener.yaml
```diff
@@ -52,7 +52,7 @@ objects:
                   - push
             - cel:
                 filter: >-
-                  body.ref = 'refs/heads/dev'
+                  body.ref == 'refs/heads/dev'
```
Fosol, line 55: Fix template bug

app/editor/src/features/content/list-view/ContentListView.tsx
```diff
@@ -117,14 +121,13 @@ export const ContentListView: React.FC = () => {
...
+    [api, filterAdvanced],
   );
 
   React.useEffect(() => {
-    fetch({ ...filter, ...filterAdvanced, sortBy });
-    // We don't want a render when the advanced filter changes.
+    fetch(filter, sortBy);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [fetch, filter, sortBy]);
+  }, [filter, sortBy]);
```
Fosol, line 130: Don't refresh when `fetch` changes, or it'll make a call when the token is refreshed...  Not certain how to work around this.

PR #56, opened 2022-03-03
TNO-208 Fix content actions

Fosol:
We can now save `content.contentActions` with the API.

Note that the API simplifies the schema.  I have also updated the Postman tests.

> Will need to run `make refresh n=api-editor`


api/editor/api-editor/src/main/java/ca/bc/gov/tno/areas/editor/models/ContentModel.java
```diff
@@ -148,6 +149,10 @@ public Content ToContent() {
             .map((file) -> new FileReference(file.getId(), content, file.getPath(), file.getMimeType(), file.getSize(),
                 file.getRunningTime()))
             .toList());
+    content.getContentActions()
```
Fosol, line 152: Note that the JSON from the API is simplified to be `content.actions: [{id:  1, value: "true"}]`

jdtoombs (approved):

PR #61, opened 2022-03-07
Fix content list view filter

Fosol:
The content list view filter wasn't correctly displaying/updating when returning to the page.  I've provided a way to set the default filter and removed one of the `React.useEffect(...)` calls.  Seems to be working now.

> Requires an `make npm-refresh`

I've noticed a few other minor issues with the page that will need to be resolved in the future.

- Table formatting when it has rows
- Paging button styling
- Table doesn't extend to end of div properly
- Need to keep the active filter when returning to the page instead of starting over with default filter
- Need to place filter parameters into URL so it can be shared or returned to from expired session.  This will need a common hook to handle.


jdtoombs (approved):
To note, I have been working on the paging button style - just so there is no crossover!

PR #63, opened 2022-03-11
TNO-247 Add data sources

Fosol:
This is a partial PR for the import ETL of data from the existing TNO solution.  The database will now contain all data sources (they are not fully configured yet, as they don't include connection settings or schedules).  We cannot store secrets in the scripts, so they will need to be added separately.

The shared **service** package will now fetch all data sources each full iteration.  This will allow dynamically add/remove functionality.  I have also started the process to enable fail limits per data source (not complete however).

I've upgraded library package versions.

- dal-db: `0.0.21-SNAPSHOT`
- service: `0.0.10-SNAPSHOT`

> Requires `make db-refresh`
> Requires `make refresh n=api-editor`


api/editor/api-editor/src/main/java/ca/bc/gov/tno/areas/admin/controllers/DataLocationController.java
```diff
@@ -0,0 +1,93 @@
...
+import ca.bc.gov.tno.dal.db.services.interfaces.IDataLocationService;
+import ca.bc.gov.tno.dal.db.entities.DataLocation;
+
+/**
+ * Endpoints to communicate with the TNO DB data data locations.
+ */
+@RolesAllowed("administrator")
+@RestController("AdminDataLocationController")
+@RequestMapping({ "/admin/data/locations", "/api/admin/data/locations" })
+public class DataLocationController {
```
Fosol, line 26: Data locations allow us to filter out data locations services don't currently support (i.e. NAS).

libs/java/service/src/main/java/ca/bc/gov/tno/services/data/BaseScheduleService.java
```diff
@@ -139,9 +139,12 @@ public void onApplicationEvent(ServiceStartEvent event) {
         }
 
         if (state.getStatus() == ServiceStatus.running) {
-          // Reset index to start over.
-          if (index == sourceConfigs.getSources().size())
+          if (index == sourceConfigs.getSources().size()) {
+            // Reset index and reinitialize configuration to start over.
+            // This enables the ability to add and remove data sources dynamically.
             index = 0;
+            initConfigs();
```
Fosol, line 146: Get all the data sources again.

libs/java/service/src/main/java/ca/bc/gov/tno/services/data/BaseScheduleService.java
```diff
@@ -151,9 +154,16 @@ public void onApplicationEvent(ServiceStartEvent event) {
...
 
           var dataSource = (C) sourceConfigs.getSources().get(index);
-          // Make request to TNO DB for data source configuration settings.
+          // Make request to TNO DB for most recent version of the data source
+          // configuration settings.
           dataSource = fetchDataSource(dataSource);
 
+          // This data source has failed too many times, ignore it and move on.
+          if (dataSource.getFailedAttempts() >= dataSource.getMaxFailedAttempts()) {
+            logger.debug(String.format("Data source has failed too many times: %s", dataSource.getId()));
```
Fosol, line 163: Don't keep trying a data source that has failed too many times.

sturufous (approved):

PR #64, opened 2022-03-15
TNO-250: Reusable Modal

jdtoombs:
Splitting off some work just so the PR isn't one big clump at the end. 

Made a reusable modal, ended up rewriting from what I had originally worked on after reading about [portals](https://reactjs.org/docs/portals.html) (and the help of a few blogs) which were very handy for setting opacity when the modal appears. Allows us to append outside the DOM node we are in.

There will be a few more styling/customization adjustments made as I finish the **View Log** modal, and match up with the modal mockup.

modal

Usage:
_Call the following within any component you wish, the portal will take care of where it is rendered_
```
const [isShowing, setIsShowing] = useState(false);

<Modal
  headerText="Header"
  onConfirm={() => alert('working')}
  body="Example body text. Are you sure you want to do the thing?"
  isShowing={isShowing}
  setIsShowing={setIsShowing}
/>

/** Then on wherever you wish to trigger the modal */
<Button onClick={()=>setIsShowing(!isShowing)} />
```


app/editor/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+  /** the text to show along with the cancel button to close the modal */
+  cancelText?: string;
+  /** the text to show along with the confirm button on the modal */
+  confirmText?: string;
+  /** function to be called when user clicks continue/confirm button */
+  onConfirm?: () => void;
+  /** the text for the header */
+  headerText?: string;
+  /** the text for the body of the modal */
+  bodyText?: string | React.ReactFragment;
```
jdtoombs, line 22: should rename this prop, will do in the a.m

app/editor/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/** Modal used to display informative and confirmation messages through the application.
+ * Portal is used to allow the modal to be appended outside of inner elements */
```
jdtoombs, line 26: Will clean this up in a.m

app/editor/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isShowing,
+  hide,
+  cancelText,
+  confirmText,
+  onConfirm,
+  headerText,
+  bodyText,
+}) =>
+  isShowing
+    ? ReactDOM.createPortal(
```
Fosol, line 37: I'm not familiar with this?

app/editor/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <>
```
Fosol, line 38: Why wrap with fragment?

app/editor/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+                </div>
+              </div>
+            </div>
+          </styled.Modal>
+        </>,
+        document.body,
+      )
+    : null;
+
+export default Modal;
```
Fosol, line 66: Don't export defaults anymore.

app/editor/src/components/row/Row.tsx
```diff
@@ -9,6 +9,14 @@ import * as styled from './styled';
...
  */
-export const Row: React.FC<React.HTMLAttributes<HTMLDivElement>> = ({ children, style }) => {
-  return <styled.Row style={style}>{children}</styled.Row>;
+export const Row: React.FC<React.HTMLAttributes<HTMLDivElement>> = ({
+  children,
+  style,
+  className,
+}) => {
+  return (
+    <styled.Row className={className} style={style}>
```
Fosol, line 18: Since you're not doing anything with the props don't extract them.
`<styled.Row {...rest}>`

app/editor/src/hooks/modal/useModal.ts
```diff
@@ -0,0 +1,15 @@
...
+    setIsShowing(!isShowing);
+  };
+
+  return {
+    isShowing,
+    toggle,
+  };
+};
+
+export default useModal;
```
Fosol, line 15: Don't export defaults anymore.

app/editor/src/hooks/modal/useModal.ts
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+const useModal = () => {
```
Fosol, line 3: Export the component here.

Fosol, line 3: Not certain of the value of this hook.  It's not currently connected to the Modal in anyway.

jdtoombs: Was just to cut down some time as it will be the same on every modal, can remove though

Fosol (approved):

PR #66, opened 2022-03-15
TNO-245 Add Audio Capture

Fosol:
Adding Build/Deploy Infrastructure as Code templates for the Audio Capture Service.

Presently the service still has the bootstrap issue so it can't start,  That issue is outside of this PR's scope.

### DIFF
--- a/openshift/templates/services/audio/capture/deploy.yaml
+++ b/openshift/templates/services/audio/capture/deploy.yaml
@@ -1,11 +1,11 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: syndication-service-deploy
+  name: audio-capture-service-deploy
   annotations:
-    openshift.io/display-name: Java Syndication Ingestion Service Image
-    description: Build an image containing the Java Syndication Ingestion Service.
-    tags: java,maven,service,syndication,kafka
+    openshift.io/display-name: Java Audio Capture Ingestion Service Image
+    description: Build an image containing the Java Audio Capture Ingestion Service.
+    tags: java,maven,service,audio,capture,kafka
 
 parameters:
   - name: SOLUTION_NAME
@@ -17,7 +17,7 @@ parameters:
     displayName: Application Name
     description: The name of the application.
     required: true
-    value: syndication
+    value: audio-capture
 
   - name: PROJECT_NAMESPACE
     displayName: OpenShift Project Namespace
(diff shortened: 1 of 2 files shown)


openshift/templates/services/audio/capture/deploy.yaml
```diff
@@ -0,0 +1,338 @@
...
+  - name: MAX_FAILED_ATTEMPTS
+    displayName: Maximum Failed Attempts
+    description: The maximum number of failed attempts before stopping the service.
+    required: true
+    value: "5"
+  - name: DATA_SOURCE_MEDIA_TYPE
+    displayName: Data Source Media Type
+    description: The name of the media type that this service will ingest.
+    required: true
+    value: News Radio
```
Fosol, line 67: Need to double check this is the correct default value.

openshift/templates/services/audio/capture/deploy.yaml
```diff
@@ -0,0 +1,338 @@
...
+      name: ${SOLUTION_NAME}-${APP_NAME}
+      namespace: ${PROJECT_NAMESPACE}-${ENV_NAME}
+      labels:
+        name: ${SOLUTION_NAME}-${APP_NAME}
+        solution: ${SOLUTION_NAME}
+        app: ${APP_NAME}
+    spec:
+      storageClassName: ${VOLUME_CLASS}
+      accessModes:
+        - ReadWriteOnce
```
Fosol, line 141: Not certain if other pods will be able to connect to this PVC.  Will test once we have a working service.

openshift/templates/services/audio/capture/deploy.yaml
```diff
@@ -0,0 +1,338 @@
...
+                  valueFrom:
+                    configMapKeyRef:
+                      name: ${SOLUTION_NAME}-${APP_NAME}
+                      key: MAX_FAILED_ATTEMPTS
+                - name: DATA_SOURCE_MEDIA_TYPE
+                  valueFrom:
+                    configMapKeyRef:
+                      name: ${SOLUTION_NAME}-${APP_NAME}
+                      key: DATA_SOURCE_MEDIA_TYPE
+              # livenessProbe:
```
Fosol, line 303: These technically work but it takes so long for a Java Spring service to start that they will fail and cause the pod to not start.  I think I may configure to wait 5+ minutes just so they are running, but it somewhat defeats the purpose.

jdtoombs (approved):

PR #71, opened 2022-03-17
Content List View Bug Fix

jdtoombs:
When clicking a row, the wrong id was being provided. `row.original` contains the values we need.

`row.id` is the id of the row itself not the content

Ex)
Row 1 would have row.id of 0
Row 2 would have row.id of 1


Fosol (approved):
That makes sense

PR #73, opened 2022-03-18
Adding the start-stop timer to the capture service.

sturufous:
Removing spurious import statements

First round of PR#62 changes

Second round of changes for PR#62

Added the start/stop scheduler to the capture service

Final tweaks prior to PR

### DIFF
--- a/tools/scripts/db-update.sh
+++ b/tools/scripts/db-update.sh
@@ -2,5 +2,5 @@
 
 cd libs/java/dal/db
 docker build -t tno:db-migration dal-db-migration --no-cache --force-rm
-docker run -it --env-file=.env --name tno-db-migration tno:db-migration
+docker run -i --env-file=.env --name tno-db-migration tno:db-migration
 docker rm tno-db-migration


services/capture/Dockerfile
```diff
@@ -7,7 +7,7 @@ FROM openjdk:17.0.1-slim AS final
 WORKDIR /app
 COPY --from=base /project/target .
 
-RUN apt-get update && apt-get install curl -y
+RUN apt-get update && apt-get install curl -y && apt-get install procps -y
```
Fosol, line 10: I think you can do this in less steps.
`RUN apt-get update && apt-get install curl procps -y`

services/capture/pom.xml
```diff
@@ -71,14 +72,14 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>2.6.3</version>
+    <version>2.5.6</version>
```
Fosol, line 75: why rolling back spring boot?

services/capture/src/main/java/ca/bc/gov/tno/services/capture/handlers/ScheduledService.java
```diff
@@ -67,14 +65,15 @@ public ScheduledService(final ServiceState state,
    * @param ranAt  The date and time the transaction ran at.
    */
   @Override
-  protected void updateDataSource(DataSourceConfig dataSource, ScheduleConfig schedule, ZonedDateTime ranOn) {
+  protected synchronized void updateDataSource(DataSourceConfig dataSource, ScheduleConfig schedule, ZonedDateTime ranOn) {
```
Fosol, line 68: What does the synchronized keywork do?

tools/scripts/db-drop.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-docker exec -it tno-database bash -c "psql -U \$POSTGRES_USER \$POSTGRES_DB -c 'DROP SCHEMA public CASCADE; CREATE SCHEMA public;'"
+docker exec -i tno-database bash -c "psql -U \$POSTGRES_USER \$POSTGRES_DB -c 'DROP SCHEMA public CASCADE; CREATE SCHEMA public;'"
```
Fosol, line 3: Is the `-it` required?

tools/scripts/db-update.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
 cd libs/java/dal/db
-docker build -t tno:db-migration dal-db-migration --no-cache --force-rm
+docker build tno:db-migration dal-db-migration --no-cache --force-rm
```
Fosol, line 4: This won't work.  Need the `-t`

tools/scripts/gen-env-files.sh
```diff
@@ -445,6 +445,46 @@ DB_URL=jdbc:postgresql://host.docker.internal:$portDatabase/$dbName
...
+
+KAFKA_BOOTSTRAP_SERVERS=host.docker.internal:$portKafkaBorkerAdvertisedExternal
+KAFKA_CLIENT_ID=audio-capture-01
+
+MAX_FAILED_ATTEMPTS=5" >> ./services/audio/.env
+    echo "./services/audio/.env created"
+fi
+
+## Audio - Clip Producer
+if test -f "./services/audio/.env"; then
```
Fosol, line 478: Why are there two configs for the same file?

Fosol (changes requested):

services/capture/Dockerfile
sturufous: Fixed it.

services/capture/pom.xml
sturufous: That was a merge error.

services/capture/src/main/java/ca/bc/gov/tno/services/capture/handlers/ScheduledService.java
sturufous: Another merge error. This was to temporarily prevent more than one thread from executing this method. It was throwing Hibernate errors because two TransactionCompleteEvents were being generated instead of one (an issue I fixed before the demo yesterday). I had removed it before the git issue yesterday afternoon.

tools/scripts/db-drop.sh
sturufous: I believe this is the "t" I had to put back in to get db-refresh to work. This is another one of those last-minute fixes that was lost late yesterday afternoon.

sturufous: The "t" causes a teletype error when I run db-drop.sh.

tools/scripts/db-update.sh
sturufous: Added it back in again. This is another last-minute change that wasn't saved.

tools/scripts/gen-env-files.sh
sturufous: Must be another merge error. I've ensured there is only one.

tools/scripts/db-update.sh
Fosol: I don't see it?

tools/scripts/db-update.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
 cd libs/java/dal/db
-docker build -t tno:db-migration dal-db-migration --no-cache --force-rm
-docker run -i --env-file=.env --name tno-db-migration tno:db-migration
+docker build tno:db-migration dal-db-migration --no-cache --force-rm
+docker run -it --env-file=.env --name tno-db-migration tno:db-migration
```
Fosol, line 5: Wrong place.

Fosol (changes requested):

tools/scripts/db-update.sh
sturufous: It was -i now it's -it?

tools/scripts/db-update.sh
Fosol: Looks like you added an unnecessary `-t` here.

Fosol (changes requested):

PR #80, opened 2022-03-30
First version of the maintenance page

sturufous:
Creating a first draft of the maintenance page. This was based on the MMIA login page and contains the same text in the centre column, even though this text isn't representative of a maintenance page. The maintenance page centre column text should likely be static, whereas the text on the login page is "message of the day" type content.


Fosol (approved):
Thanks!

jdtoombs (approved):

PR #84, opened 2022-04-04
Tno 273

sturufous:
This PR contains the following changes:

1, Change the object passed as the caller in the first argument to the ErrorEvent constructor. This allows the ErrorHandler to notify the BaseScheduler instance so that it can continue processing after an error.
2, Add a third argument (config) to the call to ErrorEvent's constructor to allow ErrorHandler to update the retry_attempts column for the data source each time an error occurs.
3, Rename the dataSource variable to config where this improves the sematics.
4, Update exception handling in the audio service for improved flow.
5, Fixed a bug in ErrorHandler where the instance variable "config" was being replaced by a local variable with the same name.


Fosol (commented):
I'm curious how the exceptions are bubbling up correctly?  If I read the code correct, you removed the catch so that the exception would be thrown.  However, since they are spring events, does this then somehow get handled correctly?

sturufous:
executeClipCmd() and clipPath() now throw all their exceptions up to line 113 (where executeClipCmd() is called). These and any other exceptions that might occur inside that try-catch block, are now caught on line 120. The way it was coded before, the failure of verifyClipSchedule() would publish an ErrorEvent, but execution would continue to the ProducerSendEvent code meaning stuff would be sent to Kafka even if verifyClipSchedule() failed.

sturufous:
Also, while executeClipCmd() and clipPath() were logging their respective exceptions they were not publishing an ErrorEvent. I hope these changes address all three of these issues without introducing any others.

Fosol (approved):

PR #86, opened 2022-04-12
TNO-281 Fixing API

Fosol:
Openshift doesn't appear to like the container using port 80, so it is now defaulted to 8080.

Fixed user registration bug.  New users will automatically get registered when making a request for `/auth/userinfo`.  This handles the basic requirement to resolve users added to Keycloak.  Work still needs to be completed on this to support adding the user to groups too.

> Requires you to refresh your api `make refresh n=api` and `make refresh n=nginx`.


api/net/Controllers/AuthController.cs
```diff
@@ -40,14 +40,15 @@ public AuthController(IUserService userService)
     /// Return environment information.
     /// </summary>
     /// <returns></returns>
-    [HttpGet("userinfo")]
+    [HttpPost("userinfo")]
```
Fosol, line 43: POST ensures it's understood it can make state changes.

tools/scripts/gen-env-files.sh
```diff
@@ -192,6 +192,7 @@ if test -f "./api/net/.env"; then
 else
 echo \
 "ASPNETCORE_ENVIRONMENT=Development
+ASPNETCORE_URLS=http://+:8080
```
Fosol, line 195: Default to port 8080

jdtoombs (approved):

PR #88, opened 2022-04-14
TNO-283 Updated service config

Fosol:
Fixing service configuration.  I thought I could remove the keycloak config but it's still required for the health API on each service.

Also updated Infrastructure as Code files for the services.


openshift/templates/services/audio/capture/deploy.yaml
```diff
@@ -67,28 +67,6 @@ parameters:
     required: true
     value: tno-services
 
-  - name: VOLUME_SIZE
```
Fosol, line 70: Moved to separate file.

openshift/templates/services/audio/pvc.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: audio-capture-service-deploy
```
Fosol, line 4: Shared volume for audio services.  Need to figure out how to allow multiple pods to connect.

openshift/templates/services/config.yaml
```diff
@@ -78,9 +78,9 @@ objects:
     data:
       API_HOST_URL: ${API_HOST_URL}
 
-      KEYCLOAK_AUTHORITY: ${KEYCLOAK_AUTHORITY}
+      KEYCLOAK_AUTH_SERVER_URL: ${KEYCLOAK_AUTH_SERVER_URL}
```
Fosol, line 81: Fix invalid copy+paste

services/audio/src/main/resources/application.yml
```diff
@@ -14,6 +14,12 @@ service:
   max-failed-attempts: ${MAX_FAILED_ATTEMPTS:3}
   failed-delay: ${FAILED_DELAY:5000}
 
+keycloak:
```
Fosol, line 17: Still required for the health API on each service.

PR #94, opened 2022-04-20
TNO-299 Fix API

Fosol:
Invalid tokens will now return a 401 instead of a 500.


api/net/Middleware/ErrorHandlingMiddleware.cs
```diff
@@ -145,11 +146,12 @@ private async Task HandleExceptionAsync(HttpContext context, Exception ex)
             }
             else if (ex is AuthenticationException)
```
Fosol, line 147: AuthenticationException was referencing invalid exception.  Should be the one from Microsoft.  Deleted the custom one.

PR #96, opened 2022-04-20
PSP-268: Time Input Mask

jdtoombs:
timeinput


app/editor/package.json
```diff
@@ -1,115 +1,117 @@
 {
-  "name": "tno-editor-app",
```
jdtoombs, line 2: linter

Fosol: Why is your linter doing this?  It makes it impossible to know what's going on.  Please find out when you can.

app/editor/src/components/form/timeinput/TimeInput.tsx
```diff
@@ -0,0 +1,17 @@
+import { InputHTMLAttributes } from 'react';
+
+import * as styled from './styled';
+
+/** Component that will enforce the HH:MM:SS time format */
+export const TimeInput: React.FC<InputHTMLAttributes<HTMLInputElement>> = (props) => {
+  const formatTime = (value: string) => {
```
Fosol, line 7: Probably should be a util.  Don't change though.  Not sure what benefit it would have to do make it a util atm.

app/editor/package.json
Fosol: Also you can highlight what changed = react-text-mask.

Fosol (approved):

app/editor/package.json
jdtoombs: Will try to figure out later today. But yes just the types for react-text-mask and the package itself

PR #98, opened 2022-04-20
TNO-214: React tooltip fix

jdtoombs:
When a variable is provided for the tooltip that starts out as undefined/null the tooltip will not be shown. The tooltip needs to be rebound after the variable is provided.

image

### DIFF
--- a/app/editor/src/App.tsx
+++ b/app/editor/src/App.tsx
@@ -23,6 +23,10 @@ function App() {
     });
   }, []);
 
+  React.useEffect(() => {
+    ReactTooltip.rebuild();
+  });
+
   return (
     <BrowserRouter>
       {keycloak ? (


app/editor/src/App.tsx
```diff
@@ -23,6 +23,10 @@ function App() {
     });
   }, []);
 
+  React.useEffect(() => {
+    ReactTooltip.rebuild();
```
Fosol, line 27: I don't understand how this would work?  It'll get called one time at the beginning of the app.  If the tooltip is set as `undefined` shouldn't we just set it to `''`?

jdtoombs: I think it might not be the issue I thought it was but rather that in some cases the tooltip is rendered before the element that uses it which results in nothing being bound. If we are going to leave the tooltip at the app.tsx level I believe we will need this. However; I got it to work moving <ReactTooltip /> within the components themselves that use tooltips

Fosol (approved):

PR #99, opened 2022-04-20
TNO-267: Page size consistency

jdtoombs:
formpage

All that was added to `ContentListView` and `ContentForm` was the addition of wrapping it in the new `<FormPage>`. Indentation is the explanation for all the additions.

### DIFF
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -1,4 +1,5 @@
 import { Area, FieldSize, IOptionItem } from 'components/form';
+import { FormPage } from 'components/form/formpage';
 import {
   FormikCheckbox,
   FormikForm,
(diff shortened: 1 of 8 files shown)


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -80,170 +81,173 @@ export const ContentForm: React.FC = () => {
...
-                        width={FieldSize.Medium}
-                        name="otherSource"
-                        label="Other Source"
-                        onChange={(e) => {
-                          const value = e.currentTarget.value;
-                          props.setFieldValue('otherSource', value);
-                          props.setFieldValue('source', value);
-                          if (!!value) {
-                            props.setFieldValue('dataSourceId', undefined);
+    <FormPage height="fit-content">
```
Fosol, line 84: This isn't ideal.  Generally it's preferred to always have the styled component as the root object returned.  Either the `styled.ContentForm` should inhert from `FormPage` or it should wrap it.

Fosol: Also trying to understand why your auto formatting is messing with the whole file.  It's impossible to see what has changed.

app/editor/src/features/content/list-view/ContentListView.tsx
```diff
@@ -127,265 +128,267 @@ export const ContentListView: React.FC = () => {
...
-                  }
-                />
-              </div>
-            </div>
-          </div>
-        </div>
-        <div className="box">
-          <h2 className="caps">Advanced Search</h2>
-          <div style={{ display: 'flex', flexDirection: 'row', minWidth: '550px' }}>
+    <FormPage>
```
Fosol, line 131: The styled component should always be the root.

Fosol (changes requested):

app/editor/src/features/content/form/ContentForm.tsx
jdtoombs: In this case it was because of the new indentation due to adding a new parent

Fosol (approved):

PR #102, opened 2022-04-21
Rename DataSource.ScheduleType

Fosol:
Forgot to provide a database column name override for DataSource.ScheduleType.

> Requires you run a `make db-refresh`


libs/net/entities/DataSource.cs
```diff
@@ -41,6 +41,7 @@ public class DataSource : AuditColumns
 
     public virtual License? License { get; set; }
 
+    [Column("schedule_type")]
```
Fosol, line 44: Consistent naming for all columns.

PR #106, opened 2022-04-21
Fix service run config error

Fosol:
Mistakenly introduced a run-time bug in the configuration by setting the active profile.

### DIFF
--- a/services/capture/pom.xml
+++ b/services/capture/pom.xml
@@ -126,12 +126,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.opennlp</groupId>
-      <artifactId>opennlp-tools</artifactId>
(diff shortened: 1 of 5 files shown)


services/syndication/pom.xml
```diff
@@ -120,12 +120,6 @@
       <version>2.13.2</version>
     </dependency>
 
-    <dependency>
```
Fosol, line 123: Not needed in this service.

services/capture/pom.xml
```diff
@@ -125,12 +125,6 @@
       <version>2.13.2</version>
     </dependency>
 
-    <dependency>
```
Fosol, line 128: Not needed in this service.

PR #108, opened 2022-04-22
Code changes to support api based data access

sturufous:


services/audio/src/main/java/ca/bc/gov/tno/services/audio/config/AudioConfig.java
```diff
@@ -58,22 +64,28 @@ public AudioConfig() {
...
+        var connection = parent.getConnection();
+        setCaptureService(parent.getCode());
+        setStreamStartTime((long) connection.get("streamStart"));
+        setCaptureDir((String) connection.get("captureDir"));
+        setClipDir((String) connection.get("clipDir"));
+        setClipCmd((String) connection.get("clipCmd"));
+        setTimezone((String) connection.get("timeZone"));
+      }
+    } catch(Exception e) {
+      logger.error("Exception initializing audio configuration.", e);
```
Fosol, line 88: Regrettably this will not register as a failure for the data source, as such it'll keep doing it over and over.

services/audio/src/main/resources/application.yml
```diff
@@ -7,9 +7,6 @@ spring:
   main:
     banner-mode: "off"
 
-  profiles:
```
Fosol, line 10: Put this back in please.  If you've rebased it'll work now.

services/capture/src/main/java/ca/bc/gov/tno/services/capture/config/CaptureConfig.java
```diff
@@ -81,7 +81,7 @@ public CaptureConfig(DataSource dataSource) {
     setMaxFailedAttempts(dataSource.getRetryLimit());
     var connection = dataSource.getConnection();
 
-    setAudioUrl((String) connection.get("audioUrl"));
+    setAudioUrl((String) connection.get("url"));
```
Fosol, line 84: I don't believe this is valid.  It should be `audioUrl` based on the original information you provided.  Additionally, the UI is mapping to `audioUrl` too.  

I actually prefer the more generic `url`.  So if you would like to use it, then you should also update the frontend to use that value too.

image

Fosol (changes requested):

services/audio/src/main/java/ca/bc/gov/tno/services/audio/config/AudioConfig.java
sturufous: The only reason I added that catch block was because it was throwing an exception where it wasn't throwing one before. This was really a stop-gap solution until I could run a permanent solution by you. If I change the constructor of DataSourceConfig, and this constructor, to throw ApiException I think that will address it. Should I go ahead and do that?

services/audio/src/main/resources/application.yml
sturufous: Will do.

services/capture/src/main/java/ca/bc/gov/tno/services/capture/config/CaptureConfig.java
sturufous: The only reason I changed this was because the front end is saving "url" in the connection field. Are you saying the front end should save the url in the field "audioUrl"?

services/audio/src/main/java/ca/bc/gov/tno/services/audio/config/AudioConfig.java
Fosol: If the exception doesn't bubble up to the event handler then it'll never get captured and recorded as such.  I'd remove they try+catch if you can and have the error handled by whatever is calling it.

services/capture/src/main/java/ca/bc/gov/tno/services/capture/config/CaptureConfig.java
Fosol: The frontend label says **Stream URL**.  But the json property is `audioUrl` because that is what you originally where expecting.

image

services/audio/src/main/java/ca/bc/gov/tno/services/audio/config/AudioConfig.java
sturufous: That was the first thing I tried, but the only way to code it was either catch and kill the exception or change the constructor of DataSourceConfig so that it also threw the ApiException. I didn't want to have to change the library if it could be avoided, but will do that so that it bubbles up all the way.

services/capture/src/main/java/ca/bc/gov/tno/services/capture/config/CaptureConfig.java
sturufous: I'll look at this when I get the TNO environment running again. Having used the new UI to create the connection string for a  data source I noticed it stored the audioUrl value in the json attribute "url" which is why I made this change. I can confirm later.

sturufous: It looks like I was looking at an RSS connection string when I saw the "url" attribute name. Now that I can test it I see that the News Radio url is named "audioUrl" so I'll set it back. The other thing that convinced me it should be named "url" is that  you updated the initConfigs() method of the Capture service to include this code:

var approvedDataSources = dataSources.stream()
        .filter((ds) -> ds.getIsEnabled() &&
            ds.getConnection().get("url") != null)
        .toList();

This failed to add any of my data sources to approvedDataSources because the attribute was named "audioUrl". I'll update this code to include "audioUrl" in place of "url".

services/audio/src/main/java/ca/bc/gov/tno/services/audio/config/AudioConfig.java
sturufous: I figured when I first saw this that it might be more complicated than it looks...and the reason is "Java"! In initConfigs() (which throws an api exception) you fill the sourceConfigs collection using the following lambda expression:

    approvedDataSources.forEach(ds -> sourceConfigs.getSources().add(new CaptureConfig(ds)));

The problem is that the constructor for CaptureConfig(ds) now throws an ApiException and because it's called inside the lambda expression it's not in the same scope as the rest of the method. As a result, the compiler requires the call to CaptureConfig(ds) to be wrapped in a try/catch block, which ruins the concise syntax and would consume the exception preventing it from bubbling up.

My suggestion is to use the following which will perform the same function and throw any ApiExceptions properly:

    for (var ds : approvedDataSources) {
      sourceConfigs.getSources().add(new CaptureConfig(ds));
    }

What do you think?

Fosol: I'm relying on your knowledge of java here.  If you think it'll work and is clean.  Go ahead.

services/capture/src/main/java/ca/bc/gov/tno/services/capture/config/CaptureConfig.java
Fosol: You can change to `url`.  I prefer it too.  But you'll need to change all relevant code so that it works throughout.

sturufous: The front-end code will need to be changed. Do you think I have the necessary knowledge to do that?

Fosol: Should be as simple as search and replace `audioUrl` with `url`.  Then you can test locally first.

sturufous: Ok, I'll do that.

Fosol (approved):

PR #113, opened 2022-04-27
TNO-316: Hide time from date range filter

jdtoombs:
image

### DIFF
--- a/app/editor/src/features/content/list-view/ContentListView.tsx
+++ b/app/editor/src/features/content/list-view/ContentListView.tsx
@@ -305,8 +305,6 @@ export const ContentListView: React.FC = () => {
                   selected={
                     !!filterAdvanced.startDate ? new Date(filterAdvanced.startDate) : undefined
                   }
-                  showTimeSelect
-                  dateFormat="Pp"
                   width={FieldSize.Small}
                   onChange={(date) =>
                     storeFilterAdvanced({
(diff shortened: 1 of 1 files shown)


app/editor/src/features/content/list-view/ContentListView.tsx
```diff
@@ -305,8 +305,6 @@ export const ContentListView: React.FC = () => {
                   selected={
                     !!filterAdvanced.startDate ? new Date(filterAdvanced.startDate) : undefined
```
Fosol, line 306: Did you test this to determine what time it does set?  I'm wondering if it will ignore content that is either at the beginning or later in the day depending on the time value.

jdtoombs: Defaults to 12am!

Fosol (changes requested):
You'll need to make sure the `from` date is 12:00:00 AM, while the `to` is 12:00:00 PM.  I think you have to use `11:59:59 PM` though.

app/editor/src/features/content/list-view/ContentListView.tsx
jdtoombs: To confirm: 

- from time starts at 12 AM (00:00:00) 
- to time now is 11:59 PM (23:59:59)

Fosol (approved):

PR #114, opened 2022-04-27
TNO-300, 310, 311: Turn off auto complete on publishedOn, Extend tags field, Reduce summary height

jdtoombs:
Should probably put these one liners in a single PR. Will start doing that now :P

### DIFF
--- a/app/editor/src/features/content/form/styled/ContentForm.tsx
+++ b/app/editor/src/features/content/form/styled/ContentForm.tsx
@@ -2,12 +2,7 @@ import styled from 'styled-components';
 
 export const ContentForm = styled.div`
   #txa-summary {
-    height: 150px;
-    width: 1000px;
-  }
-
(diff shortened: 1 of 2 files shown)


app/editor/src/features/content/form/styled/ContentForm.tsx
```diff
@@ -1,6 +1,15 @@
 import styled from 'styled-components';
 
 export const ContentForm = styled.div`
+  #txa-summary {
```
Fosol, line 4: I always worry this type of styling will cause issues with smaller screens.  It isn't responsive.

jdtoombs: Updated most of it to use FieldSizes, and left height as is as I think we don't have to worry as much about vertical scroll.

Fosol (approved):

PR #116, opened 2022-05-01
Add yarn cache to improve build

Fosol:
The build process in Openshift can take up to 30 minutes to build the frontend application.  I'm hoping by including the yarn cache we can speed this up dramatically.  However, it may cause annoying issues with source control.  We'll give it a go to see.

Added an example access request page for new users who have not yet been authorized/approved to use the site.  It's not functional, but it gives an idea of what we can do.

A number of performance improvements for users after the first visit.  Many of the dropdown lists are now cached in local storage.  While nice, it doesn't solve the first time they need to be downloaded.  Additionally, it adds complexity to deal with stale cache.  Note - Stale cache is not yet handled.

Resolved a number of bugs with the Content List View, like sorting, filtering and state management.

> Requires you run `make refresh n=api` and `make refresh n=editor`

### Summary

- Add yarn cache to improve build
- Fix content sorting
- Fix api config
- Add access request page
- Improve performance
- Add local storage
- tno-core v0.0.120
- tno-core v0.0.121

### Access Request Page
image


app/editor/.gitignore
```diff
@@ -24,7 +24,7 @@ yarn-error.log*
 
 # yarn
 .yarn/*
-.yarn/cache
+!.yarn/cache
```
Fosol, line 27: This is how I'm now including the yarn cache.

app/editor/package.json
```diff
@@ -35,7 +35,7 @@
     "react-tooltip": "4.2.21",
     "redux-logger": "3.0.6",
     "styled-components": "5.3.5",
-    "tno-core": "0.0.119",
+    "tno-core": "0.0.121",
```
Fosol, line 38: Upgraded to latest version

PR #120, opened 2022-05-02
Fix summon hook

Fosol:
There was a logic error in the summon hook which for some reason never showed up until now.  Essentially the first request would fail because the axios instance was not created yet.


libs/npm/core/src/hooks/summon/useSummon.tsx
```diff
@@ -96,7 +96,8 @@ export const useSummon = ({
       },
     );
 
-    (axiosInstances as any)[baseURL ?? 'default'] = instance;
+    (axiosInstances as any)[baseURL ?? 'default'] = summon;
+    instance.current = summon;
```
Fosol, line 100: This is required so that the first attempt to use the summon hook doesn't return `undefined`

PR #122, opened 2022-05-02
TNO-322 Fix content sorting

Fosol:
The Content List View "Section/Page" will now work.  Additionally, fixed DAL to support multiple sort columns.

image


libs/npm/core/src/utils/toQueryString.ts
```diff
@@ -5,8 +5,18 @@
  */
 export const toQueryString = (params: object) => {
   if (params === undefined || params === null) return '';
-  return Object.keys(params)
-    .filter((key) => (params as any)[key] !== undefined)
-    .map((key) => `${encodeURIComponent(key)}=${encodeURIComponent((params as any)[key])}`)
-    .join('&');
+  const values = [];
```
Fosol, line 8: Correctly return array parameters as separate query param values.  Before it was joining them with a comma.  But this would result in a single value instead of multiple.

openshift/kustomize/postgres/crunchy/base/deploy.yaml
```diff
@@ -9,6 +9,11 @@ metadata:
     solution: tno
     app: tno-database
 spec:
+  users:
```
Fosol, line 12: Haven't tried this yet, but hoping we can initialize the database better.

jdtoombs (approved):

PR #124, opened 2022-05-03
TNO-333 Add etag caching

Fosol:
I've added a way to tracking ETag caching for specific API endpoints.  This allows the app to make light AJAX requests without requiring additional load on the database.  Essentially the app sends a request with an ETag asking if it already has the latest version of the data.  If it does, the API will return a 304 response.

This still requires a heavy first load on the initial load of the application regrettably.  Until we offload cache to another server (i.e. Redis) there isn't much we can do about this.

I can also increase the complexity of the app to only make a request for the cache table and then compare against it's local copy before making even the light requests.  But I'm less incline to do this, as it doesn't really solve the initial load issue.

> Requires `make db-refresh`, `make refresh n=api`, and `make refresh n=editor`

image

### DIFF
--- a/api/net/Areas/Editor/Controllers/ActionController.cs
+++ b/api/net/Areas/Editor/Controllers/ActionController.cs
@@ -3,6 +3,7 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Swashbuckle.AspNetCore.Annotations;
 using TNO.API.Areas.Editor.Models.Action;
+using TNO.API.Filters;
 using TNO.API.Models;
 using TNO.DAL.Services;
 
@@ -43,10 +44,13 @@ public class ActionController : ControllerBase
     /// Return an array of Action.
     /// </summary>
     /// <returns></returns>
-    [HttpGet]
+    [HttpGet, HttpHead]
     [Produces("application/json")]
     [ProducesResponseType(typeof(IEnumerable<ActionModel>), (int)HttpStatusCode.OK)]
+    [ProducesResponseType((int)HttpStatusCode.NotModified)]
     [SwaggerOperation(Tags = new[] { "Action" })]
+    [ETagCacheTableFilter("actions")]
+    [ResponseCache(Duration = 5 * 60)]
     public IActionResult FindAll()
     {
         return new JsonResult(_service.FindAll().Select(m => new ActionModel(m)));
--- a/api/net/Program.cs
+++ b/api/net/Program.cs
@@ -187,6 +187,7 @@ builder.Services.AddVersionedApiExplorer(options =>
     options.SubstituteApiVersionInUrl = true;
 
 });
+builder.Services.AddMemoryCache();
 
 var app = builder.Build();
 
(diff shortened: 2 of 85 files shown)


api/net/Areas/Editor/Controllers/ActionController.cs
```diff
@@ -43,10 +44,13 @@ public ActionController(IActionService service)
     /// Return an array of Action.
     /// </summary>
     /// <returns></returns>
-    [HttpGet]
+    [HttpGet, HttpHead]
     [Produces("application/json")]
     [ProducesResponseType(typeof(IEnumerable<ActionModel>), (int)HttpStatusCode.OK)]
+    [ProducesResponseType((int)HttpStatusCode.NotModified)]
     [SwaggerOperation(Tags = new[] { "Action" })]
+    [ETagCacheTableFilter("actions")]
```
Fosol, line 52: Here is the main magic to handle and create ETags.

api/net/Program.cs
```diff
@@ -187,6 +187,7 @@
     options.SubstituteApiVersionInUrl = true;
 
 });
+builder.Services.AddMemoryCache();
```
Fosol, line 190: Not using this yet, but we can add this later to improve performance.  But I'm incline to move things to Redis before doing much with this.

app/editor/src/hooks/api-editor/editor/useApiActions.ts
```diff
@@ -17,8 +17,9 @@ export const useApiActions = (
   const api = useApi(options);
 
   return {
-    getActions: () => {
-      return api.get<IActionModel[]>(`/editor/actions`);
+    getActions: (etag: string | undefined = undefined) => {
+      const config = !!etag ? { headers: { 'If-None-Match': etag } } : undefined;
+      return api.get<IActionModel[]>(`/editor/actions`, config);
```
Fosol, line 22: We can now pass ETag headers.

app/editor/src/store/hooks/lookup/utils/addOrUpdate.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Update the value in the array, or add it to the array if it doesn't exist.
+ * @param value
+ * @param values
+ * @param find
+ * @returns
+ */
+export const addOrUpdate = <T>(value: T, values: T[], find: (value: T) => boolean) => {
```
Fosol, line 8: Probably should move this to the root utils folder.

app/editor/src/store/hooks/lookup/utils/fetchIfNoneMatch.ts
```diff
@@ -0,0 +1,46 @@
+import { AxiosError, AxiosResponse } from 'axios';
+import { ICacheModel } from 'hooks/api-editor';
+
+import { addOrUpdate, getFromLocalStorage, initFromLocalStorage } from '.';
+
+export const fetchIfNoneMatch = async <T>(
```
Fosol, line 6: I need to document this.  This function does a lot of magic for the hooks, redux store, error handling, and ETags.

app/editor/src/store/hooks/lookup/utils/fetchIfNoneMatch.ts
```diff
@@ -0,0 +1,46 @@
...
+        (c) => c.key === key,
+      );
+      localStorage.setItem('cache', JSON.stringify(cache));
+
+      return result;
+    } catch (error) {
+      const ae = error as AxiosError;
+      if (ae.response?.status === 304) {
+        store(items);
+        return items;
```
Fosol, line 41: I think there may be a weird situation where local storage is messed up and results are empty...  May need to handle this oddity.

PR #126, opened 2022-05-03
Expand width of Headline column

jdtoombs:
Splitting this off

`useFlexLayout` allows the `width` attribute in columns to be applied

image


libs/npm/core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tno-core",
   "description": "TNO shared library",
-  "version": "0.0.123",
+  "version": "0.0.124",
```
jdtoombs, line 4: Speaking of linters, to avoid mine going crazy on me there is a quick trick if there are one off times you want to save without your auto linter running. `CTRL + K`, followed by `Ctrl + SHIFT + S`

app/editor/src/features/content/list-view/constants/columns.tsx
```diff
@@ -8,6 +8,7 @@ export const columns: (Column<IContentModel> & UseSortByColumnOptions<IContentMo
     id: 'headline',
     Header: 'Headline',
     accessor: 'headline',
+    width: 300,
```
Fosol, line 11: Will these widths essentially ruin any smaller resolutions?

Fosol (approved):

app/editor/src/features/content/list-view/constants/columns.tsx
jdtoombs: It should be ok, without it it defaults to `150px`. That page does need to be made more responsive though. I was able to do it with the MediaTypes I am working on so I should be apply to apply it over there

responsive

PR #130, opened 2022-05-05
Fix upload/download

Fosol:
The upload/download was broken when the data source dropdown list was changed.  I've modified the design so that this won't happen, and also solved other issues I was aware of.

- Uploaded file path is now stored so that it doesn't change after first created
- Changing the data source after won't break the file

> Note that existing files already uploaded will probably not work.  I don't want to spend time moving them around at this point.

### DIFF
--- a/app/editor/src/components/upload/Upload.tsx
+++ b/app/editor/src/components/upload/Upload.tsx
@@ -103,7 +103,7 @@ export const Upload: React.FC<IUploadProps> = ({
         hide={toggle}
         type="delete"
         headerText="Confirm Removal"
-        body="Are you sure you want to remove this file?"
+        body="Are you sure you want to remove this file?  You will still need to save before it is deleted."
         confirmText="Yes, Remove It"
         onConfirm={() => {
           if (!!fileRef.current) {
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -127,7 +127,6 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
       // If the upload fails, we still need to update the form from the original update.
       if (!!contentResult) {
         setContent(toForm(contentResult));
-        toast.error(`${contentResult.headline} file was not uploaded successfully.`);
         if (!originalId) navigate(`/contents/${contentResult.id}`);
       }
     }
@@ -197,9 +196,9 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
                           ''
                         }
                         onChange={(newValue: any) => {
-                          props.setFieldValue('dataSourceId', newValue.value);
-                          props.setFieldValue('source', newValue.label);
                           const dataSource = dataSources.find((ds) => ds.id === newValue.value);
+                          props.setFieldValue('dataSourceId', newValue.value);
+                          props.setFieldValue('source', dataSource?.code ?? '');
                           if (!!dataSource) {
                             props.setFieldValue('mediaTypeId', dataSource.mediaTypeId);
                             props.setFieldValue('licenseId', dataSource.licenseId);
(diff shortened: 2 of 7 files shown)


app/editor/src/components/upload/Upload.tsx
```diff
@@ -103,7 +103,7 @@ export const Upload: React.FC<IUploadProps> = ({
         hide={toggle}
         type="delete"
         headerText="Confirm Removal"
-        body="Are you sure you want to remove this file?"
+        body="Are you sure you want to remove this file?  You will still need to save before it is deleted."
```
Fosol, line 106: Clarity

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -127,7 +127,6 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
       // If the upload fails, we still need to update the form from the original update.
       if (!!contentResult) {
         setContent(toForm(contentResult));
-        toast.error(`${contentResult.headline} file was not uploaded successfully.`);
```
Fosol, line 130: No need for another error

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -197,9 +196,9 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
                           ''
                         }
                         onChange={(newValue: any) => {
-                          props.setFieldValue('dataSourceId', newValue.value);
-                          props.setFieldValue('source', newValue.label);
                           const dataSource = dataSources.find((ds) => ds.id === newValue.value);
+                          props.setFieldValue('dataSourceId', newValue.value);
+                          props.setFieldValue('source', dataSource?.code ?? '');
```
Fosol, line 201: Use the code, not the value from the dropdown label

libs/net/dal/Extensions/FileReferenceExtensions.cs
```diff
@@ -39,7 +53,7 @@ public static string GetFilePath(this IReadonlyFileReference fileReference, TNOC
         // TODO: Handle when the original file uploaded has different path than the new one.
         var dataConnection = JsonSerializer.Deserialize<Dictionary<string, object>>(location.Connection) ?? new Dictionary<string, object>();
         var connectionPath = dataConnection.ContainsKey("path") ? $"{((string?)dataConnection["path"])?.RemoveStartAndEnd("/")}" : "";
-        var path = Path.Combine(storageConfig.GetPath(), source ?? "", connectionPath);
```
Fosol, line 42: Don't include the source as part of the storage location

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -121,7 +121,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.HasIndex("Name")
                         .IsUnique();
 
-                    b.ToTable("action");
+                    b.ToTable("action", (string)null);
```
Fosol: Hmm this doesn't look right.

jdtoombs (approved):

libs/net/dal/Models/ContentFileReference.cs
```diff
@@ -124,7 +159,7 @@ public ContentFileReference(FileReference fileReference, IFormFile file)
         this.File = file ?? throw new ArgumentNullException(nameof(file));
         this.Id = fileReference.Id;
         this.FileName = file.FileName;
-        this.Path = GenerateFileName(this.Content, file);
+        this.Path = GenerateFilePath(this.Content, file);
```
Fosol, line 162: Keep the whole path to the file (not including the data location config path).

libs/npm/core/src/hooks/summon/useSummon.tsx
```diff
@@ -20,14 +20,10 @@ let axiosInstances = {
 /**
  * useSummon hook properties.
  */
-export interface ISummonProps {
+export interface ISummonProps extends AxiosRequestConfig<any> {
```
Fosol, line 23: Allow us to configure summon/axios going forward.

PR #131, opened 2022-05-06
TNO-303: Admin Media Types

jdtoombs:
Still need to do a scan over with fresh eyes in the morning. 

Things to fix:

- Updating a media type is still failing (may need some help, might be obvious from PR)

image

Also tried to match styling on `DataSourceList`:
image


api/net/Areas/Admin/Models/MediaType/MediaTypeModel.cs
```diff
@@ -0,0 +1,76 @@
...
+    #endregion
+
+    #region Methods
+    /// <summary>
+    /// Explicit conversion to entity.
+    /// </summary>
+    /// <param name="model"></param>
+    public static explicit operator Entities.MediaType(MediaTypeModel model)
+    {
+        var entity = new Entities.MediaType(model.Name)
```
Fosol, line 65: Looks like you're missing setting the `Version` property.  Which will result in failures for updates because of concurrency.

app/editor/src/features/admin/media-types/constants/strings.ts
```diff
@@ -0,0 +1,2 @@
+export const NEW_MEDIA_TYPE_ROUTE = '/admin/media/types/0';
```
Fosol, line 1: I'm not a big fan of these types of constants.  URL paths should be found in one place, not all over the place.  I realize you're essentially trying to use a constant to make it in one place.  But now there will be file here and there, and everywhere with paths.  I'd move this to the routes component folder and keep them all in one place.  

Related thought - if we're making components self contained, we wouldn't want these types of constants used anyway.  We'd have a prop with a default value in the component.  That way we can control it for each instances.

app/editor/src/features/admin/media-types/styled/MediaTypeList.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from 'styled-components';
+
+export const MediaTypeList = styled.div`
+  width: 100%;
```
Fosol, line 4: Don't think we should use `100%` ever.  I think `auto` is safer.  But I'm not an expert in this area.  Just finding issues when we do it this way.

app/editor/src/features/admin/styled/AdminPage.tsx
```diff
@@ -1,8 +1,5 @@
 import styled from 'styled-components';
 
 export const AdminPage = styled.div`
-  padding: 0.5em;
-  margin-left: 5%;
-  margin-right: 5%;
   width: 100%;
```
Fosol, line 4: Another `100%`

app/editor/src/features/navbar/NavBar.tsx
```diff
@@ -45,6 +45,11 @@ export const NavBar: React.FC = () => {
               label="CBRA Report"
               claim={Claim.administrator}
             />
+            <NavBarItem
+              navigateTo="/admin/media/types"
```
Fosol, line 49: Just an example of where you've used constant sometimes and not others.  I think we probably shouldn't use constant values for paths.  Instead just have props with default values that we can override if required.

app/editor/src/hooks/api-editor/admin/useApiAdminDataSources.ts
```diff
@@ -3,8 +3,8 @@ import { defaultEnvelope, ILifecycleToasts } from 'tno-core';
 import { IDataSourceModel, IPaged, useApi } from '..';
 
 /**
- * Common hook to make requests to the API.
- * @returns CustomAxios object setup for the API.
+ * Common hook to make requests to the PIMS API.
```
Fosol, line 6: You added back in the old copy+paste PIMS

app/editor/src/hooks/api-editor/admin/useApiAdminMediaTypes.ts
```diff
@@ -0,0 +1,39 @@
...
+    baseURL?: string;
+  } = {},
+) => {
+  const api = useApi(options);
+
+  return {
+    findAllMediaTypes: () => {
+      return api.get<IMediaTypeModel[]>(`/admin/media/types`);
+    },
+    findMediaTypes: () => {
```
Fosol, line 23: You don't have any filter passed to this

api/net/Areas/Admin/Controllers/MediaTypeController.cs
```diff
@@ -0,0 +1,146 @@
...
+        _service = service;
+    }
+    #endregion
+
+    #region Endpoints
+    /// <summary>
+    /// Find a page of media type for the specified query filter.
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet]
```
Fosol, line 52: I think for admin we should probably use a filter `/all` for all items.

api/net/Areas/Admin/Controllers/MediaTypeController.cs
```diff
@@ -0,0 +1,146 @@
...
+        var uri = new Uri(this.Request.GetDisplayUrl());
+        var query = Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery(uri.Query);
+        return new JsonResult(_service.FindAll().Select(ds => new MediaTypeModel(ds)));
+    }
+
+    /// <summary>
+    /// Find a page of media type for the specified query filter.
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet("find")]
```
Fosol, line 67: I think for admin we should have the page/search results as the root path `/`

app/editor/src/features/admin/media-types/MediaType.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const [, api] = useMediaTypes();
+  const { state } = useLocation();
+  const { id } = useParams();
+  const navigate = useNavigate();
+  const mediaTypeId = Number(id);
+  const [mediaType, setMediaType] = React.useState<IMediaTypeModel>(
+    (state as any)?.mediaType ?? defaultMediaType,
+  );
+
+  const { toggle, isShowing } = useModal();
```
Fosol, line 27: For organization and standards.  Keep all your `use*()` statements at the top of the component/hook.   Then a space.
Then all your `useState()` statements grouped together.
Then a space.
Then your custom variables needs from the above.
Then a space.
Then all your `useEffects()` statements grouped together.
Then a space.
Then all your functions and handler statements grouped together.

app/editor/src/features/admin/media-types/MediaType.tsx
```diff
@@ -0,0 +1,125 @@
...
+        iconType="back"
+        label="Back to Media Types"
+        className="back-button"
+        onClick={() => navigate('/admin/media/types')}
+      />
+      <FormikForm
+        initialValues={mediaType}
+        validate={(values) => {
+          const errors = {};
+          return errors;
```
Fosol, line 64: Were you planning to do anything with this?

app/editor/src/features/admin/media-types/MediaTypeFilter.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const MediaTypeFilter: React.FC<IMediaTypeFilterProps> = ({ setGlobalFilter }) => {
+  const [filter, setFilter] = React.useState<string>('');
+  const navigate = useNavigate();
+  return (
+    <>
+      <Row className="add-media" justify="flex-end">
+        <IconButton
+          iconType="plus"
+          label="Add New Media Type"
+          onClick={() => navigate(NEW_MEDIA_TYPE_ROUTE)}
```
Fosol, line 21: Pass the route as a prop

app/editor/src/features/admin/media-types/MediaTypeList.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const navigate = useNavigate();
+  const [{ mediaTypes }, api] = useMediaTypes();
+  const [{ requests }] = useApp();
+
+  const [items, setItems] = React.useState<IMediaTypeModel[]>([]);
+
+  console.log(mediaTypes.length);
+
+  React.useEffect(() => {
+    // if (!!mediaTypes.length) {
```
Fosol, line 23: Remove if not needed.

app/editor/src/features/admin/media-types/MediaTypeList.tsx
```diff
@@ -0,0 +1,47 @@
...
+import * as styled from './styled';
+
+export const MediaTypeList: React.FC = () => {
+  const navigate = useNavigate();
+  const [{ mediaTypes }, api] = useMediaTypes();
+  const [{ requests }] = useApp();
+
+  const [items, setItems] = React.useState<IMediaTypeModel[]>([]);
+
+  console.log(mediaTypes.length);
```
Fosol, line 20: Remove log

app/editor/src/store/hooks/admin/mediatypes/useMediaTypes.ts
```diff
@@ -0,0 +1,81 @@
...
+      },
+      deleteMediaType: async (model: IMediaTypeModel) => {
+        const result = await dispatch<IMediaTypeModel>('delete-data-source', () =>
+          api.deleteMediaType(model),
+        );
+        store.storeMediaTypes(getMediaTypes().filter((ds) => ds.id !== result.id));
+        return result;
+      },
+    }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Fosol, line 76: This is dangerous.  Were you getting infinite loops?  I know I've had to do this too, but I hate it.  It worries me.  Should we be doing this a different way?  Perhaps a singleton and not use hooks?

libs/net/dal/Services/MediaTypeService.cs
```diff
@@ -20,5 +24,53 @@ public IEnumerable<MediaType> FindAll()
...
+
+        if (!String.IsNullOrWhiteSpace(filter.Name))
+            query = query.Where(c => EF.Functions.Like(c.Name, $"%{filter.Name}%"));
+        if (!String.IsNullOrWhiteSpace(filter.Description))
+            query = query.Where(c => c.Description == filter.Description);
+
+        var total = query.Count();
+
+        if (filter.Sort?.Any() == true)
+            query = query.OrderByProperty(filter.Sort);
```
Fosol, line 44: Need to fix this with the current code we're using within the `ContentService.Find` method.  I just fixed this bug.

libs/net/dal/Services/MediaTypeService.cs
```diff
@@ -20,5 +24,53 @@ public IEnumerable<MediaType> FindAll()
...
+        else
+            query = query.OrderBy(c => c.Name).ThenBy(c => c.IsEnabled);
```
Fosol, line 46: Default sort should be `SortOrder`, then `Name`

libs/net/dal/Services/MediaTypeService.cs
```diff
@@ -20,5 +24,53 @@ public IEnumerable<MediaType> FindAll()
...
+    public IPaged<MediaType> Find(MediaTypeFilter filter)
+    {
+        var query = this.Context.MediaTypes
+            .Include(c => c.Name)
+            .Include(c => c.Description)
+            .Include(c => c.IsEnabled)
+            .AsQueryable();
+
+        if (!String.IsNullOrWhiteSpace(filter.Name))
+            query = query.Where(c => EF.Functions.Like(c.Name, $"%{filter.Name}%"));
```
Fosol, line 37: I think case-sensitive will cause us issues.

Fosol (changes requested):

api/net/Areas/Admin/Controllers/MediaTypeController.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet]
+    [Produces("application/json")]
+    [ProducesResponseType(typeof(IPaged<MediaTypeModel>), (int)HttpStatusCode.OK)]
+    [SwaggerOperation(Tags = new[] { "MediaType" })]
+    public IActionResult FindAll()
+    {
+        var uri = new Uri(this.Request.GetDisplayUrl());
+        var query = Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery(uri.Query);
```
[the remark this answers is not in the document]
jdtoombs: Will do!!

app/editor/src/features/admin/media-types/constants/strings.ts
jdtoombs: Makes sense, I am getting flashbacks to the REDUX constants from before 😛

app/editor/src/store/hooks/admin/mediatypes/useMediaTypes.ts
jdtoombs: Oops removed and seems to work fine!

api/net/Areas/Admin/Models/MediaType/MediaTypeModel.cs
jdtoombs: @Fosol, added Version. Still getting a failure (my whole container dies). Will try to gather some further logs

api/net/Areas/Admin/Models/MediaType/MediaTypeModel.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            Id = model.Id,
+            Name = model.Name,
+            Description = model.Description,
+            IsEnabled = model.IsEnabled,
+            SortOrder = model.SortOrder,
+            Version = model.Version ?? 0,
```
Fosol, line 72: I wonder if we should allow the audit column version to be null?

app/editor/src/store/hooks/admin/mediatypes/useMediaTypes.ts
```diff
@@ -0,0 +1,80 @@
...
+  updateMediaType: (model: IMediaTypeModel) => Promise<IMediaTypeModel>;
+  deleteMediaType: (model: IMediaTypeModel) => Promise<IMediaTypeModel>;
+}
+
+export const useMediaTypes = (): [IAdminState, IMediaTypeController] => {
+  const api = useApiAdminMediaTypes();
+  const dispatch = useApiDispatcher();
+  const [state, store] = useAdminStore();
+
+  const getMediaTypes = () => state.mediaTypes;
```
Fosol, line 20: I'm curious what you're trying here?  Based on this warning I suspect it isn't working though.

Fosol (commented):
If you finish the few remaining comments I'll merge.
Hoping you have a great vacation!

app/editor/src/store/hooks/admin/mediatypes/useMediaTypes.ts
jdtoombs: Was just keeping consistent with useDataSources.ts but got rid of getMediaTypes now!

app/editor/src/features/admin/media-types/MediaType.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const navigate = useNavigate();
+  const mediaTypeId = Number(id);
+  const [mediaType, setMediaType] = React.useState<IMediaTypeModel>(
+    (state as any)?.mediaType ?? defaultMediaType,
+  );
+
+  const { toggle, isShowing } = useModal();
+
+  React.useEffect(() => {
+    if (mediaType?.id !== mediaTypeId) {
```
Fosol, line 30: This will infinite loop if no data comes back.  I've fixed this in my PR.

libs/net/dal/Services/MediaTypeService.cs
```diff
@@ -20,5 +24,59 @@ public IEnumerable<MediaType> FindAll()
...
+    {
+        return this.Context.MediaTypes.FirstOrDefault(c => c.Id == id);
+    }
+
+    public MediaType? FindByName(string name)
+    {
+        return this.Context.MediaTypes.FirstOrDefault(c => c.Name.ToLower() == name.ToLower());
+    }
+
+    public override MediaType Add(MediaType entity)
```
Fosol, line 71: Not needed as you're note overriding anything.  I've fixed this in my PR.

libs/net/dal/Services/MediaTypeService.cs
```diff
@@ -20,5 +24,59 @@ public IEnumerable<MediaType> FindAll()
...
+    {
+        base.Add(entity);
+        return entity;
+    }
+
+    public override MediaType Update(MediaType entity)
+    {
+        return this.Update(entity);
```
Fosol, line 79: Here was your infinite loop issue.  I've fixed this in my PR.

libs/net/dal/Services/MediaTypeService.cs
```diff
@@ -20,5 +24,59 @@ public IEnumerable<MediaType> FindAll()
...
+        return this.Context.MediaTypes.FirstOrDefault(c => c.Name.ToLower() == name.ToLower());
+    }
+
+    public override MediaType Add(MediaType entity)
+    {
+        base.Add(entity);
+        return entity;
+    }
+
+    public override MediaType Update(MediaType entity)
```
Fosol, line 77: Not needed as you're not overriding anything.  I've fixed this in my PR.

libs/net/dal/Services/MediaTypeService.cs
```diff
@@ -20,5 +24,59 @@ public IEnumerable<MediaType> FindAll()
     {
         return this.Context.MediaTypes.OrderBy(a => a.SortOrder).ThenBy(a => a.Name).ToArray();
     }
+
+    public IPaged<MediaType> Find(MediaTypeFilter filter)
+    {
+        var query = this.Context.MediaTypes
+            .Include(c => c.Name)
```
Fosol, line 31: Include statements are for child entities (foreign keys).  They are not for property names like this.  I've fixed this in my PR.

Fosol (commented):
I've submitted a PR to replace this one.  Wish I could just update your PR, but it's not possible.
I've fixed a number of bugs, and also added data locations to data sources.

PR #134, opened 2022-05-11
TNO-336 Add lookup endpoint

Fosol:
Improving performance by adding a lookup endpoint.  One request is sometimes more performant than a bunch of small requests.

We do still have a fundamental performance issue with the API in that it should handle hundreds of concurrent requests, but presently on the first load it has issues and at some point in the communication network a gateway timeout occurs.   The API shouldn't be so slow the first time, but perhaps it simply needs to be initialized before users hit it.

> Not required, but I have updated the Initial migration scripts `make db-refresh` if you want the latest cache keys.


app/editor/src/store/hooks/lookup/useLookup.ts
```diff
@@ -79,153 +91,269 @@ export const useLookup = (): [ILookupState, ILookupController] => {
         store.storeCache(result);
         return result;
       },
+      getLookups: async () => {
+        return await fetchIfNoneMatch<ILookupModel>(
+          'lookups',
+          dispatch,
+          (etag) => lookups.getLookups(etag),
+          (results) => {
+            if (!!results) {
```
Fosol, line 100: We don't store the lookup response in local storage.  Instead we store each lookup value in local storage.  This eliminates us having to keep two copies in sync.

libs/net/dal/Extensions/DbContextExtensions.cs
```diff
@@ -122,6 +129,10 @@ public static TNOContext OnUpdate<T>(this TNOContext context, T entity, ClaimsPr
             cache.Value = Guid.NewGuid().ToString();
             context.Update(cache);
         }
+        else
+        {
+            context.Add(new Cache(key, Guid.NewGuid()));
```
Fosol, line 134: Adding it to cache is helpful to handle situations where we haven't created an entry for the cache table.

libs/net/entities/Action.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace TNO.Entities;
 
-[Cache("actions")]
+[Cache("actions", "lookups")]
```
Fosol, line 6: This ensures all these cache keys will get updated when this type of entity is updated/added/removed.

PR #138, opened 2022-05-16
TNO-336 Update services

Fosol:
Remove kafka logger from Java services.  It was causing the PVC storage to fill up very quickly.

Moved 'Connection Settings' to the 'Service Settings' tab in Data Source management.

### DIFF
--- a/libs/net/services/Controllers/StatusController.cs
+++ b/libs/net/services/Controllers/StatusController.cs
@@ -60,19 +60,5 @@ public class StatusController : ControllerBase
         _service.State.Resume();
         return new JsonResult(_service.State);
     }
-
-    /// <summary>
-    /// Request service to stop.
-    /// </summary>
-    /// <returns></returns>
-    [HttpPut("stop")]
-    [Produces("application/json")]
-    [ProducesResponseType(typeof(ServiceStateModel), (int)HttpStatusCode.OK)]
-    [ProducesResponseType((int)HttpStatusCode.BadRequest)]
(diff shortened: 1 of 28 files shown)


libs/net/services/Controllers/StatusController.cs
```diff
@@ -60,19 +60,5 @@ public IActionResult ResumeService()
...
-
-    /// <summary>
-    /// Request service to stop.
-    /// </summary>
-    /// <returns></returns>
-    [HttpPut("stop")]
-    [Produces("application/json")]
-    [ProducesResponseType(typeof(ServiceStateModel), (int)HttpStatusCode.OK)]
-    [ProducesResponseType((int)HttpStatusCode.BadRequest)]
-    public IActionResult StopService()
```
Fosol, line 72: No longer relevant

libs/net/services/ServiceManager`.cs
```diff
@@ -74,7 +74,8 @@ public async Task RunAsync()
         // Run at the shortest interval of all schedules.
         var delay = dataSources.Min(ds => ds.DataSourceSchedules.Where(s => s.Schedule?.DelayMS != 0).Min(s => s.Schedule?.DelayMS)) ?? _options.DefaultDelayMS;
 
-        while (this.State.Status != ServiceStatus.Stopped)
+        // Always keep looping until an unexpected failure occurs.
+        while (true)
```
Fosol, line 78: I want to look for a better way to keep a thread running in a loop.

PR #141, opened 2022-05-17
TNO-336 Fix service api

Fosol:
Fix service API controller.

### DIFF
--- a/libs/net/services/Controllers/HealthController.cs
+++ b/libs/net/services/Controllers/HealthController.cs
@@ -12,7 +12,7 @@ namespace TNO.Services.Controllers;
 public class HealthController : ControllerBase
 {
     #region Variables
-    private readonly IDataSourceManager _service;
+    private readonly IServiceManager _service;
     #endregion
 
     #region Constructors
@@ -20,7 +20,7 @@ public class HealthController : ControllerBase
     /// Creates a new instance of a HealthController object, initializes with specified parameters.
     /// </summary>
     /// <param name="service"></param>
-    public HealthController(IDataSourceManager service)
+    public HealthController(IServiceManager service)
     {
         _service = service;
     }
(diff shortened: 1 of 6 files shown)


libs/net/services/Controllers/HealthController.cs
```diff
@@ -12,15 +12,15 @@ namespace TNO.Services.Controllers;
 public class HealthController : ControllerBase
 {
     #region Variables
-    private readonly IDataSourceManager _service;
+    private readonly IServiceManager _service;
```
Fosol, line 15: Base service is required for dependency injection

openshift/kustomize/services/content/base/deploy.yaml
```diff
@@ -109,7 +109,7 @@ spec:
               path: "/health"
               port: 8080
               scheme: HTTP
-            initialDelaySeconds: 60
+            initialDelaySeconds: 30
```
Fosol, line 112: Service is up in less than 30 seconds

services/net/content/ContentService.cs
```diff
@@ -38,7 +38,7 @@ protected override IServiceCollection ConfigureServices(IServiceCollection servi
         services
             .Configure<ContentOptions>(this.Configuration.GetSection("Service"))
             // .AddTransient<IServiceAction<ContentOptions>, ContentAction>()
-            .AddTransient<IServiceManager, ContentManager>();
+            .AddSingleton<IServiceManager, ContentManager>();
```
Fosol, line 41: Singleton so that the api gets the same object as the service.

services/net/syndication/SyndicationService.cs
```diff
@@ -38,7 +38,7 @@ protected override IServiceCollection ConfigureServices(IServiceCollection servi
             .Configure<SyndicationOptions>(this.Configuration.GetSection("Service"))
             .AddTransient<IIngestAction<SyndicationOptions>, SyndicationAction>()
             .AddTransient<DataSourceIngestManagerFactory<SyndicationDataSourceManager, SyndicationOptions>>()
-            .AddTransient<IServiceManager, SyndicationManager>();
+            .AddSingleton<IServiceManager, SyndicationManager>();
```
Fosol, line 41: Singleton so that the api gets the same object as the service.

PR #147, opened 2022-05-20
Content service duplicates

Fosol:
The Content Service was built fast to pull content to Editors.  It was creating a lot of duplicates.  This update should reduce the duplication as it asks the API if the content identified by the `source` and `uid` already exists.

I've also increased the memory limit for the Content Service because 150Mi wasn't enough.

### DIFF
--- a/libs/net/services/Runners/BaseService.cs
+++ b/libs/net/services/Runners/BaseService.cs
@@ -1,4 +1,5 @@
 ﻿using System.IdentityModel.Tokens.Jwt;
+using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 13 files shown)


libs/net/services/Runners/BaseService.cs
```diff
@@ -49,6 +50,7 @@ public BaseService(string[] args)
         this.Configuration = Configure(builder).Build();
         ConfigureServices(builder.Services);
         this.App = builder.Build();
+        Console.OutputEncoding = Encoding.UTF8;
```
Fosol, line 53: Trying to fix the log encoding issue in Openshift.  Not sure how to do it however.

PR #148, opened 2022-05-24
TNO-44 Fix token refresh

Fosol:
The services were not updating the access token reliably.  I have updated the code to handle a scenario where the refresh token was not provided.

### DIFF
--- a/libs/net/core/Http/OpenIdConnectRequestClient.cs
+++ b/libs/net/core/Http/OpenIdConnectRequestClient.cs
@@ -91,25 +91,27 @@ namespace TNO.Core.Http
         public async Task<string> RequestAccessToken()
         {
             HttpResponseMessage response;
-            if (_accessToken == null || String.IsNullOrWhiteSpace(_accessToken.AccessToken) || (!String.IsNullOrWhiteSpace(_accessToken.RefreshToken) && _tokenHandler.ReadJwtToken(_accessToken.RefreshToken).ValidTo <= DateTime.UtcNow))
+            var now = DateTime.UtcNow;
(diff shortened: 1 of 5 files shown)


libs/net/core/Http/OpenIdConnectRequestClient.cs
```diff
@@ -91,25 +91,27 @@ public class OpenIdConnectRequestClient : HttpRequestClient, IOpenIdConnectReque
         public async Task<string> RequestAccessToken()
         {
             HttpResponseMessage response;
-            if (_accessToken == null || String.IsNullOrWhiteSpace(_accessToken.AccessToken) || (!String.IsNullOrWhiteSpace(_accessToken.RefreshToken) && _tokenHandler.ReadJwtToken(_accessToken.RefreshToken).ValidTo <= DateTime.UtcNow))
+            var now = DateTime.UtcNow;
```
Fosol, line 94: Asking for the current datetime multiple times is not performant, and will result in different values.

libs/net/core/Http/OpenIdConnectRequestClient.cs
```diff
@@ -91,25 +91,27 @@ public class OpenIdConnectRequestClient : HttpRequestClient, IOpenIdConnectReque
...
+            var accessToken = _accessToken?.AccessToken != null ? _tokenHandler.ReadJwtToken(_accessToken.AccessToken) : null;
+            var refreshToken = _accessToken?.RefreshToken != null ? _tokenHandler.ReadJwtToken(_accessToken.RefreshToken) : null;
+            if (accessToken == null ||
```
Fosol, line 97: This is where the bug I believe originated.  It wasn't handling the scenario where the refresh token was not provided.

PR #149, opened 2022-05-25
TNO-351 Add Capture service

Fosol:
Refactoring the Java services into .NET for performance, maintainability, tight integration with other .NET libraries, and for development efficiency.  Updated Initial DB migration to handle data source optimistic concurrency issues.

> You will need to update most of your services.  I have created a new `make renew` command to do it all for you.

There is now a generic Command Service that we can use to schedule arbitrary commands.  I've overloaded this to make the Capture Service.

The Capture Service now handles shutdown requests and failures.  The only downside to failures is that it will start up another capture file instead of appending to the prior one.  I don't think we would want to append, but it makes it more challenging for the Clip Service.  As I will need to handle these types of failures somehow.

### Summary

- Add generic Command Service
- Add Capture Service

### Data Source Management
image


Makefile
```diff
@@ -131,6 +131,15 @@ remove: ## Remove all containers
 	$(info Remove all containers)
 	@docker-compose rm -sv
 
+renew: ## Refresh all relevant services that were impacted by prior Pull Request.
```
Fosol, line 134: A new command we can update when we need to update specific services based on the prior PR.  It's not a perfect solution because if you fall behind in running it, you'll have to manually perform the necessary refreshes.

api/net/Areas/Admin/Models/DataSource/DataSourceModel.cs
```diff
@@ -174,9 +174,9 @@ public DataSourceModel(Entities.DataSource entity, JsonSerializerOptions options
         this.Topic = entity.Topic;
         this.ParentId = entity.ParentId;
         this.Connection = JsonSerializer.Deserialize<Dictionary<string, object>>(entity.Connection, options) ?? new Dictionary<string, object>();
-        this.LastRanOn = entity.LastRanOn;
         this.RetryLimit = entity.RetryLimit;
-        this.FailedAttempts = entity.FailedAttempts;
+        this.LastRanOn = entity.DataService?.LastRanOn;
```
Fosol, line 178: Needed to separate the properties that are updated by the ingestion services so that they don't cause optimistic concurrency isssues.

api/net/Areas/Services/Controllers/DataSourceController.cs
```diff
@@ -79,34 +78,22 @@ public IActionResult FindByMediaType(string mediaTypeName)
...
-            {
-                var schedule = result.SchedulesManyToMany.FirstOrDefault(s => s.ScheduleId == dss.ScheduleId)?.Schedule;
-                if (schedule != null && schedule.RunOn != dss.Schedule.RunOn)
-                {
-                    schedule.RunOn = dss.Schedule.RunOn;
-                    _serviceSchedule.Update(schedule);
-                }
-            }
-        });
+        _serviceDataService.AddOrUpdate(model.ToEntity(_serializerOptions).DataService!);
```
Fosol, line 93: Only need to update the DataService, not the DataSource.

app/editor/src/components/form/text/Text.tsx
```diff
@@ -30,6 +30,12 @@ export interface ITextProps extends InputHTMLAttributes<HTMLInputElement> {
    * Form field size.
    */
   width?: FieldSize;
+  /**
+   * Function to format the value.
+   */
+  formatter?: (
```
Fosol, line 36: Now we can format the value.  I don't think this will handle masks though.  It's just a quick and dirty way to format values after a render.

app/editor/src/components/form/text/utils/formatDate.ts
```diff
@@ -0,0 +1,19 @@
+import moment from 'moment';
+
+/**
+ * Attempts to format the value in the text input.
+ * @param value The text input value.
+ * @param format The format you want to apply.  Defaults to local.
+ * @returns A formatted date value, or the original value provided.
+ */
+export const formatDate = (
```
Fosol, line 9: I don't love the location of this util.  It's really specifically related to input values, as such I want it to stay with the form values.

libs/net/services/Interfaces/IServiceActionManager.cs
```diff
@@ -26,6 +26,12 @@ public interface IServiceActionManager
     /// <returns></returns>
     public Task RunAsync();
 
+    /// <summary>
+    /// Ask the process action(s) to stop.
+    /// </summary>
+    /// <returns></returns>
+    public Task StopAsync();
```
Fosol, line 33: Need a way to stop service actions after they start.

libs/net/services/Interfaces/IServiceAction`.cs
```diff
@@ -16,6 +16,7 @@ public interface IServiceAction<TOptions>
     /// Perform the action for the specified manager.
     /// </summary>
     /// <param name="manager"></param>
-    public Task PerformActionAsync(IServiceActionManager manager);
+    /// <param name="name"></param>
+    public Task PerformActionAsync(IServiceActionManager manager, string? name = null);
```
Fosol, line 20: Adding the name provides us a way to execute different actions.

libs/net/services/ServiceState.cs
```diff
@@ -45,7 +45,7 @@ public int RecordFailure()
     {
         this.Failures++;
 
-        if (this.Failures >= this.MaxRetryLimit) this.Status = ServiceStatus.Sleeping;
+        if (this.Failures >= this.MaxRetryLimit) this.Status = ServiceStatus.RequestSleep;
```
Fosol, line 48: Need transition status so that the service manager can ask service actions to stop.

PR #152, opened 2022-05-27
TNO-351 Refactor Clip Service

Fosol:
The Clip Service has now been refactored into .NET.  It will now validate the capture stream before attempting to clip a file.  It does this by checking when the capture stream was created and the duration.  It will also generate clips for any schedule that is missing a file.

> Use the `make renew` to get the latest changes.

### Summary

- Refactore Clip Service
- Update Capture Service
- Update Command Service
- Update Data Source Admin page

image


app/editor/package.json
```diff
@@ -6,9 +6,9 @@
   "homepage": "/",
   "packageManager": "yarn@3.2.0",
   "dependencies": {
-    "@headlessui/react": "1.6.1",
+    "@headlessui/react": "1.6.3",
```
Fosol, line 9: Updated deps to latest versions

PR #153, opened 2022-05-30
TNO-302 Fix Clip Service

Fosol:
The Clip Service is erroring out quickly if the Capture Service doesn't contain the stream.  I've modified the Clip Service to hopefully keep it running even when this occurs.

Also updated the Data Source Admin to be more consistent in naming conventions.

image


app/editor/src/features/admin/sources/ContentReferenceList.tsx
```diff
@@ -86,6 +86,7 @@ export const ContentReferenceList: React.FC<IContentReferenceListProps> = (props
 
   return (
     <styled.ContentReferenceList>
+      <h2>{values.name}</h2>
```
Fosol, line 89: Name of Data Source which makes managing tabs more clear.

PR #159, opened 2022-06-01
TNO-340: Advanced Schedule Validation

jdtoombs:
Fields must be touched in order for validation to occur. I tried to mimic what Formik does on submit. Seems to be the only work around.

See: https://github.com/jaredpalmer/formik/issues/2734#issuecomment-690810715

validationOnDone

### DIFF
--- a/app/editor/src/hooks/api-editor/validation/ScheduleSchema.ts
+++ b/app/editor/src/hooks/api-editor/validation/ScheduleSchema.ts
@@ -1,7 +1,6 @@
 import { IScheduleModel, ScheduleTypeName } from 'hooks';
 import * as Yup from 'yup';
 
-import { ScheduleType } from '../constants';
 import { AuditColumnsSchema } from './AuditColumnsSchema';
 
 /**
@@ -28,7 +27,6 @@ export const ScheduleSchema: Yup.SchemaOf<IScheduleModel> = AuditColumnsSchema.s
     }
   }),
   startAt: Yup.string().when('scheduleType', (scheduleType) => {
-    console.log(scheduleType === ScheduleType.Daily);
     if (scheduleType === ScheduleTypeName.Daily || scheduleType === ScheduleTypeName.Advanced) {
       return Yup.string().required(`Required for ${scheduleType} schedules.`);
     } else {


app/editor/src/features/admin/sources/services/ScheduleAdvanced.tsx
```diff
@@ -88,7 +92,7 @@ export const ScheduleAdvanced: React.FC<IScheduleAdvancedProps> = (props) => {
             </Button>
           )}
           {index !== undefined && (
-            <Button onClick={handleDone} variant={ButtonVariant.secondary}>
+            <Button type="submit" onClick={handleDone} variant={ButtonVariant.secondary}>
```
Fosol, line 94: Does `onClick` get fired if user presses enter in the form somewhere else?
It's very odd to have the button as the `submit` but it's using the `secondary` variant.  Why is this?

app/editor/src/features/admin/sources/services/ScheduleAdvanced.tsx
```diff
@@ -39,9 +39,13 @@ export const ScheduleAdvanced: React.FC<IScheduleAdvancedProps> = (props) => {
   const handleDone = async (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
     if (index !== undefined) {
       try {
-        // TODO: Validation doesn't work.
-        await AdvancedScheduleSchema.validate(values.schedules[index], { abortEarly: false });
-        setIndex(undefined);
+        // Mock what formik does on submit, values must be "touched" for validation to fire.
+        const errors = await validateForm();
```
Fosol, line 42: I find the logic challenging here.  What will cause an error to be thrown here?  I would expect if validation to fail it would throw an error, which then would never result in `setTouched`.

app/editor/src/features/admin/sources/services/ScheduleAdvanced.tsx
```diff
@@ -39,9 +39,13 @@ export const ScheduleAdvanced: React.FC<IScheduleAdvancedProps> = (props) => {
   const handleDone = async (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
     if (index !== undefined) {
       try {
-        // TODO: Validation doesn't work.
-        await AdvancedScheduleSchema.validate(values.schedules[index], { abortEarly: false });
-        setIndex(undefined);
+        // Mock what formik does on submit, values must be "touched" for validation to fire.
```
Fosol, line 41: This also seems like something we may need in a common utils function.

app/editor/src/features/admin/sources/services/ScheduleAdvanced.tsx
jdtoombs: oops this was meant to be removed!

app/editor/src/features/admin/sources/services/ScheduleAdvanced.tsx
jdtoombs: Doing a few other tweaks right now with the YupSchema

app/editor/src/hooks/api-editor/validation/ScheduleSchema.ts
```diff
@@ -14,13 +15,27 @@ export const ScheduleSchema: Yup.SchemaOf<IScheduleModel> = AuditColumnsSchema.s
...
-  delayMS: Yup.number().defined(),
+  delayMS: Yup.number().defined('Must be defined'),
   runOn: Yup.date()
     .optional()
     .default(undefined)
     .transform((curr, orig) => (!orig ? undefined : curr)),
-  startAt: Yup.string().optional().default(undefined),
-  stopAt: Yup.string().optional().default(undefined),
+  stopAt: Yup.string().when('scheduleType', (scheduleType) => {
+    console.log(scheduleType === ScheduleType.Daily);
```
Fosol, line 24: remove

Fosol (changes requested):
Just the one `console.debug` needs to be removed.

app/editor/src/hooks/api-editor/validation/ScheduleSchema.ts
jdtoombs: Woops, gone!

Fosol (changes requested):

Fosol (approved):

PR #161, opened 2022-06-02
TNO-327: Pager styling

jdtoombs:
Disabled button was working prior, background color was the same when the page wasn't properly expanding (which has been fixed in a separate PR already merged), top margin removed.

We could make the disabled colour more obvious if needed. 

image


Fosol (changes requested):
Can you make the active button blue or something.  It's impossible to see on many screens.

jdtoombs:
@Fosol, updated. Guessing the colour will need a revise but that is easy enough.

Fosol (approved):

Fosol:
Not the prettiest blue, but it'll do for now.

PR #164, opened 2022-06-02
TNO-378 Fix syndication feed dates

Fosol:
There are two date time bugs.

- Many feeds have an invalid time-zone abbreviation.  Which results in the syndication feed failing.  I've created a simple string parser to determine a possible time-zone offset for the abbreviation.
- The services compare the local system datetime to the `DataSource.LastRanOn` datetime, but it would result in an incorrect comparison.  I don't know if I've fixed this, but it now defaults to UTC.  I'll need to test the Capture and Clip services at some point.

### DIFF
--- a/libs/net/services/Config/ServiceOptions.cs
+++ b/libs/net/services/Config/ServiceOptions.cs
@@ -27,6 +27,6 @@ public class ServiceOptions
     /// <summary>
     /// get/set - The service timezone.
     /// </summary>
-    public string TimeZone { get; set; } = "Pacific Standard Time";
+    public string TimeZone { get; set; } = "UTC";
     #endregion
 }
--- a/services/net/syndication/SyndicationAction.cs
+++ b/services/net/syndication/SyndicationAction.cs
@@ -203,13 +203,16 @@ public class SyndicationAction : IngestAction<SyndicationOptions>
 
         try
         {
-            var xmlr = XmlReader.Create(new StringReader(data));
-            return SyndicationFeed.Load(xmlr);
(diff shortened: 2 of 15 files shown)


libs/net/services/Config/ServiceOptions.cs
```diff
@@ -27,6 +27,6 @@ public class ServiceOptions
     /// <summary>
     /// get/set - The service timezone.
     /// </summary>
-    public string TimeZone { get; set; } = "Pacific Standard Time";
+    public string TimeZone { get; set; } = "UTC";
```
Fosol, line 30: The system time is normally UTC.  Hoping this fix for Syndication Service doesn't break the Capture and Clip Services.

services/net/syndication/SyndicationAction.cs
```diff
@@ -203,13 +203,16 @@ private async Task<SyndicationFeed> GetFeedAsync(Uri url, IDataSourceIngestManag
...
         {
+            // Reformat dates because timezone abbreviations don't work...
+
             var xmlr = XmlReader.Create(new StringReader(data));
-            return SyndicationFeed.Load(xmlr);
+            var feed = SyndicationFeed.Load(xmlr);
+
+            // Need to manually test if `pubDate` is valid.  Microsoft didn't bother providing a way to work with valid dates...
+            // Essentially timezone values are context sensitive.
+            var pubDate = feed.Items.FirstOrDefault()?.PublishDate;
```
Fosol, line 213: Forcing an error if it fails.

PR #165, opened 2022-06-08
TNO-342 Add Storage Controller

Fosol:
I've added a Storage Controller which gives us access to files uploaded by the Capture and Clip services (anything that uploads to a mapped volume).  The API provides a way to navigate through folders on the mapped volume and perform basic functions (upload, download, stream, move, delete).

The UI is very basic and ugly, but generally functional.

> The API streaming endpoint currently has no security.  We need to find out how to use AJAX to provide the stream to the video player.

image


api/net/Areas/Editor/Controllers/StorageController.cs
```diff
@@ -0,0 +1,200 @@
...
+        return new JsonResult(new ItemModel(safePath));
+    }
+
+    /// <summary>
+    /// Stream the file for the specified path.
+    /// </summary>
+    /// <param name="path"></param>
+    /// <param name="location"></param>
+    /// <returns></returns>
+    [AllowAnonymous] // TODO: Temporary to test HTML 5 video
```
Fosol, line 106: We need to figure out how to use this endpoint with the HTML 5 video player.

api/net/appsettings.json
```diff
@@ -42,5 +42,9 @@
     "ValueLengthLimit": 524288000,
     "MultipartBodyLengthLimit": 524288000,
     "MultipartHeadersLengthLimit": 524288000
+  },
+  "Storage": {
```
Fosol, line 46: Default storage mapped locations.

docker-compose.yml
```diff
@@ -179,7 +179,8 @@ services:
       - database
       - keycloak
     volumes:
-      - tno-upload:/data
+      - tno-api-data:/data
```
Fosol, line 182: API has access to both mapped volumes.

app/editor/src/hooks/api-editor/editor/useApiStorage.ts
```diff
@@ -0,0 +1,118 @@
...
+      var response = await api.get<any, AxiosResponse<any, never>, any>(
+        `/editor/storage${location ? `/${location}` : ''}/download?${toQueryString(params)}`,
+        {
+          responseType: 'blob',
+          headers: { accept: '*.*' },
+        },
+      );
+
+      const uri = window.URL.createObjectURL(new Blob([response.data]));
+      const link = document.createElement('a');
```
Fosol, line 89: I'd like to make an option that allows the user more control when downloading files.  Presently this auto starts the download to their default/last location.

jdtoombs (approved):

PR #167, opened 2022-06-09
TNO-342 Fix stream bug

Fosol:
Streaming was hard-coded to localhost and needed to be updated to be relative.

Also updated Data Source admin UI by moving **Capture Source** and **Data Location** to the **Ingest Settings** tab, as it makes more sense there.  However, the validation doesn't currently inform the user to go to that tab, so this will need to be updated at some point.

@jdtoombs Note the changes to the Data Source admin page.  Not sure if it'll impact you when demoing.

image

### DIFF
--- a/app/editor/src/features/admin/sources/DataSourceDetails.tsx
+++ b/app/editor/src/features/admin/sources/DataSourceDetails.tsx
@@ -1,4 +1,4 @@
-import { FieldSize, IOptionItem, OptionItem } from 'components/form';
+import { IOptionItem } from 'components/form';
 import { FormikCheckbox, FormikSelect, FormikText, FormikTextArea } from 'components/formik';
 import { useFormikContext } from 'formik';
 import { IDataSourceModel } from 'hooks/api-editor';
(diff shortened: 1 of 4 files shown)


app/editor/src/features/admin/sources/DataSourceDetails.tsx
```diff
@@ -101,22 +83,6 @@ export const DataSourceDetails: React.FC<IDataSourceDetailsProps> = () => {
           options={licenses}
           required
         />
-        <FormikSelect
```
Fosol, line 104: Moved to Ingest Settings tab

app/editor/src/features/storage/StorageListView.tsx
```diff
@@ -35,9 +35,7 @@ export const StorageListView: React.FC = (props) => {
   const selectItem = (item?: IItemModel) => {
     setItem(item);
     setStreamUrl(
-      !!item
-        ? `http://localhost:40080/api/editor/storage/stream?path=${folder.path}/${item.name}`
-        : undefined,
+      !!item ? `/api/editor/storage/stream?path=${folder.path}/${item.name}` : undefined,
```
Fosol, line 38: Fixed bug

jdtoombs (approved):

PR #168, opened 2022-06-13
Install and configure the Prometheus client library

andrelashley:


api/net/Program.cs
```diff
@@ -238,6 +239,9 @@
 app.UseRouting();
 app.UseCors();
 
+app.UseMetricServer();
```
Fosol, line 242: Appears to be missing configuration required to connect to Prometheus.  Also according to documentation it should be configured before error handling middleware.

Fosol: Best practices require these endpoints to require authentication/authorization.  Adding it this way will be open to all.

andrelashley:
Closing this PR to recreate it.

PR #169, opened 2022-06-16
TNO 360, 330, 339

jdtoombs:
This PR Includes:

**1.  Combined snippet and list view (iteration 1)**
combinedView

**2. Dynamic navigation bar (only appear on hover of navigation)** 
dynamicNavbar

3. Ability to delete time log entries
deleteTimeLogEntry


app/editor/src/features/content/combined-view/CombinedView.tsx
```diff
@@ -0,0 +1,59 @@
+import 'react-reflex/styles.css';
```
jdtoombs, line 1: Initially had logic on regular content form for a condensed version for the half screen size but got too messy. Made this instead. I think spending some time on the scalability of our components would be beneficial as they "kinda" are right now. But they do not look great when the collapse.

app/editor/src/features/content/form/CondensedContentForm.tsx
```diff
@@ -0,0 +1,359 @@
+import { faUpRightFromSquare } from '@fortawesome/free-solid-svg-icons';
```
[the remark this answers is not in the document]
jdtoombs: Will discuss some of the issues I had with mockup vs actual application with you on Monday!

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -220,6 +231,22 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
                         required={!!props.values.otherSource}
                       />
                     </Row>
+                    <Row>
+                      <FormikText
```
Fosol, line 235: Is this a new field or is it just moved?

app/editor/src/hooks/api-editor/interfaces/ITimeTrackingModel.ts
```diff
@@ -5,4 +5,5 @@ export interface ITimeTrackingModel {
   activity: string;
   userName?: string;
   createdOn?: any;
+  id?: number;
```
Fosol, line 8: Is `id` truly undefinable?  Or is it mostly likely `0` for a new row?

libs/npm/core/src/components/grid-table/GridTable.tsx
```diff
@@ -191,6 +192,8 @@ export const GridTable = <T extends object = Record<string, unknown>>({
     onChangeSort(sortBy);
   }, [onChangeSort, sortBy]);
 
+  const { id } = useParams();
```
Fosol, line 195: What does the `id` do for the GridTable?  This looks like contamination of purpose.

Fosol: Is every single GridTable going to use this functionality?  I doubt it.  Abstract it so that it makes sense for general use.

app/editor/src/features/content/form/ContentForm.tsx
jdtoombs: Just moved, also regarding the two content forms. I think I am going to take another shot at controlling it with a `condensed` prop and some conditional rendering - working on that right now

jdtoombs: Nevermind I think it is more confusing with that...

app/editor/src/hooks/api-editor/interfaces/ITimeTrackingModel.ts
jdtoombs: In some areas of the app (when passing the parsed data to the pop up) the id is undefined at that time, could adjust this if needed

Fosol (changes requested):

libs/npm/core/src/components/grid-table/GridTable.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useParams } from 'react-router-dom';
```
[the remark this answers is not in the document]
jdtoombs: Good catch

app/editor/src/hooks/api-editor/interfaces/ITimeTrackingModel.ts
Fosol: The model should represent what the API will return.  Does the API ever return null?

libs/npm/core/src/components/center/Center.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+import * as styled from './styled';
+
+export interface ICenterProps extends React.HtmlHTMLAttributes<HTMLDivElement> {}
+
+/** Simple component used for centering text/items within a cell of a table */
+export const Center: React.FC<ICenterProps> = ({ children }) => {
+  return <styled.Center className="center">{children}</styled.Center>;
```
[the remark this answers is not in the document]
Fosol: From a maintenance perspective it's overkill and unless we can solve testing the shared library may frustrate us with future changes.  As far as I can tell you're just centering, which can be achieved with a className.  I was expecting the component to do more than just center text.  I don't see a lot of value in this one.  I suppose it saves you writing `<div className="center"></div>`.  I suppose if it did some additional stuff it would make sense.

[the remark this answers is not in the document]
jdtoombs: Fair enough, I can take it out!

app/editor/src/features/content/list-view/constants/columns.tsx
```diff
@@ -46,10 +46,10 @@ export const columns: (Column<IContentModel> & UseSortByColumnOptions<IContentMo
     accessor: (row) => row.status,
   },
   {
-    id: 'publishedOn',
+    id: 'createdOn',
```
Fosol, line 49: @jdtoombs I just changed this because the `createdOn` doesn't seem to be useful unless the `publishedOn` is null.  Is there a reason you overwrote?

app/editor/src/features/content/list-view/constants/condensedColumns.tsx
```diff
@@ -0,0 +1,87 @@
...
+  },
+  {
+    id: 'status',
+    Header: () => <div className="center">Status</div>,
+    width: 100,
+    accessor: (row) => row.status,
+    Cell: ({ value }: { value: string }) => <div className="center">{value}</div>,
+  },
+  {
+    id: 'createdOn',
```
Fosol, line 62: See my comments for the `columns.tsx` file

Fosol (changes requested):

app/editor/src/features/content/list-view/constants/columns.tsx
jdtoombs: Ugh no my rebase was all messed one second

Fosol (approved):

Fosol (approved):

PR #170, opened 2022-06-20
Added contents/stream endpoint

sturufous:
Add streaming endpoint so that audio/video can be previewed in the Snippets page of the editor.

### DIFF
--- a/api/net/Areas/Editor/Controllers/ContentController.cs
+++ b/api/net/Areas/Editor/Controllers/ContentController.cs
@@ -2,13 +2,17 @@ using System.Net;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
 using Swashbuckle.AspNetCore.Annotations;
+using TNO.API.Areas.Editor.Models.Storage;
 using TNO.API.Areas.Editor.Models.Content;
 using TNO.API.Models;
 using TNO.DAL.Models;
 using TNO.DAL.Services;
+using TNO.DAL.Config;
 using TNO.Entities;
 using TNO.Entities.Models;
+using TNO.Core.Extensions;
 
 namespace TNO.API.Areas.Editor.Controllers;
 
@@ -30,6 +34,7 @@ public class ContentController : ControllerBase
     #region Variables
     private readonly IContentService _contentService;
     private readonly IFileReferenceService _fileReferenceService;
+    private readonly StorageConfig _config;
     #endregion
 
     #region Constructors
@@ -38,10 +43,11 @@ public class ContentController : ControllerBase
     /// </summary>
     /// <param name="contentService"></param>
     /// <param name="fileReferenceService"></param>
-    public ContentController(IContentService contentService, IFileReferenceService fileReferenceService)
+    public ContentController(IContentService contentService, IFileReferenceService fileReferenceService, IOptions<StorageConfig> options)
     {
         _contentService = contentService;
         _fileReferenceService = fileReferenceService;
+        _config = options.Value;
     }
     #endregion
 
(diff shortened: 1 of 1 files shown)


libs/net/dal/Config/ContentConfig.cs
```diff
@@ -0,0 +1,86 @@
+using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Hosting;
+
+namespace TNO.DAL.Config;
+
+/// <summary>
+/// ContentConfig class, provides a way to configure storage of files.
+/// </summary>
+public class ContentConfig
```
Fosol, line 9: Don't duplicate code and functionality that doesn't need to be duplicated.  We already have a `StorageConfig` class that does all of this.

api/net/Areas/Editor/Models/Content/FolderModel.cs
```diff
@@ -0,0 +1,51 @@
+using TNO.Core.Extensions;
+
+namespace TNO.API.Areas.Editor.Models.Content;
+
+/// <summary>
+/// FolderModel class, provides a model that represents an file or directory.
+/// </summary>
+public class FolderModel
```
Fosol, line 8: What is this model used for within this PR?

api/net/Areas/Editor/Models/Content/ItemModel.cs
```diff
@@ -0,0 +1,61 @@
+using MimeTypes;
+
+namespace TNO.API.Areas.Editor.Models.Content;
+
+/// <summary>
+/// ItemModel class, provides a model that represents an file or directory.
+/// </summary>
+public class ItemModel
```
Fosol, line 8: This model is a duplicate of the `ItemModel` for the `StorageController`, but you are not using it as a model.  You are only using for the functionality to extract some of the properties for the stream.  Don't duplicate code if you don't need to.  There are a few exceptions of this rule, which a true model would be one of them, but you are not using this model for the purpose of a model.

Fosol (changes requested):

libs/net/dal/Config/ContentConfig.cs
sturufous: I assumed that as you had asked me to reproduce the streaming endpoint in the ContentController that I should also reproduce supporting classes. I believe the streaming endpoint in the StorageController will stream from both storage and upload locations. I'm happy to do whatever fits with your original intentions.

Fosol: The StorageController endpoints were an example of how to get a stream to work.  Just remove duplicate where possible and reasonable.

api/net/Areas/Editor/Models/Content/FolderModel.cs
sturufous: I thought I had included this because I was getting compilation errors without it, but this seems not to be the case. I've removed the file.

api/net/Areas/Editor/Models/Content/ItemModel.cs
sturufous: This was duplicated for the same reason as ContentConfig.cs. I can use the functionality in ItemModel under storage if that's the best approach.

sturufous: Waiting for your response before making changes and resubmitting.

Fosol: Just use the `ItemModel` that already exists if you need the same functionality and you're not actually using the model for what a model is used for.  Models are for serialization.  We duplicate models because they are just properties with no logic.  That way endpoints can mutate without breaking other endpoints.

libs/net/dal/Config/ContentConfig.cs
sturufous: Removed this class.

api/net/Areas/Editor/Models/Content/ItemModel.cs
sturufous: Altered the code to use the ItemModel.cs in Storage.

Fosol (approved):

PR #172, opened 2022-06-20
TNO-356 Fix Clip Service

Fosol:
The Clip Service would shutdown as a result of scheduled run failures.  The failure was expected (the Capture file was missing), however based on configuration the service should continue to run other schedules without erroring out.

This update wraps the schedule runs with `try+catch` and correctly handles the failures based on config.

### DIFF
--- a/libs/net/services/Actions/IngestAction.cs
+++ b/libs/net/services/Actions/IngestAction.cs
@@ -34,7 +34,7 @@ public abstract class IngestAction<TOptions> : ServiceAction<TOptions>, IIngestA
     /// <param name="manager"></param>
     /// <param name="name"></param>
     /// <param name="cancellationToken"></param>
-    public abstract Task PerformActionAsync(IDataSourceIngestManager manager, string? name = null, CancellationToken cancellationToken = default);
+    public abstract Task PerformActionAsync<T>(IDataSourceIngestManager manager, string? name = null, T? data = null, CancellationToken cancellationToken = default) where T : class;
 
     /// <summary>
     /// Perform the action for the specified data source.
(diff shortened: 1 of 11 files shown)


libs/net/services/Actions/IngestAction.cs
```diff
@@ -34,7 +34,7 @@ public IngestAction(IApiService api, IOptions<TOptions> options) : base(api, opt
     /// <param name="manager"></param>
     /// <param name="name"></param>
     /// <param name="cancellationToken"></param>
-    public abstract Task PerformActionAsync(IDataSourceIngestManager manager, string? name = null, CancellationToken cancellationToken = default);
+    public abstract Task PerformActionAsync<T>(IDataSourceIngestManager manager, string? name = null, T? data = null, CancellationToken cancellationToken = default) where T : class;
```
Fosol, line 37: This isn't required presently, but allows us to pass generic data when making calling the action.

libs/net/services/Managers/DataSourceManager`.cs
```diff
@@ -116,6 +116,14 @@ public override async Task RunAsync()
                         await manager.RecordFailureAsync();
                         this.State.RecordFailure();
                     }
+                    catch (AggregateException ex)
```
Fosol, line 119: Not required, but helps to debug issues in async code.

services/net/clip/ClipAction.cs
```diff
@@ -56,55 +56,66 @@ public ClipAction(IKafkaMessenger kafka, IApiService api, IOptions<ClipOptions>
...
+                    await StopProcessAsync(process, cancellationToken);
+                    RemoveProcess(manager, schedule);
                 }
             }
-            else if (name == "stop")
+            catch (Exception ex)
             {
-                await StopProcessAsync(process, cancellationToken);
-                RemoveProcess(manager, schedule);
+                if ((ex is MissingFileException || (ex is AggregateException && ex.InnerException is MissingFileException)) &&
```
Fosol, line 115: This is the new code to fix the bug.  The rest is the result of wrapping in the `try+catch`

PR #175, opened 2022-06-22
TNO-358 Update Content Service

Fosol:
The Content Service will now listen to all configured topics and import content.  It will also copy clip files to the API if they exist.

> The Initial DB migration has been refreshed, you will need to `make renew`


app/editor/src/features/admin/sources/connections/AudioClip.tsx
```diff
@@ -70,6 +70,15 @@ export const AudioClip: React.FC = (props) => {
           setFieldValue('connection.throwOnMissingFile', e.currentTarget.checked);
         }}
       />
+      <p>Do not turn on "Import Content" until you have successfully ingested content.</p>
+      <FormikCheckbox
```
Fosol, line 74: I don't love this solution, but it functions for now.  Essentially, the Content Service will error out if the Kafka topic does not exist.  If an Ingest Service hasn't pushed any messages to the Kafka topic, then the topic will probably not exist (unless it was manually created).

libs/net/models/Areas/Services/Models/DataSource/DataLocationModel.cs
```diff
@@ -8,6 +10,15 @@ namespace TNO.API.Areas.Services.Models.DataSource;
...
     #region Properties
+    /// <summary>
+    /// get/set - The location type.
+    /// </summary>
+    public DataLocationType LocationType { get; set; }
+
+    /// <summary>
+    /// get/set - The connection settings.
+    /// </summary>
+    public Dictionary<string, object> Connection { get; set; } = new Dictionary<string, object>();
```
Fosol, line 21: Need these properties so that Ingest Services have access to it.  I don't love returning credentials this way in the future so we'll need to tighten up security for this endpoint.

libs/net/services/Config/IngestServiceOptions.cs
```diff
@@ -7,8 +7,19 @@ public class IngestServiceOptions : ServiceOptions
 {
     #region Properties
     /// <summary>
-    /// get/set - An array of media type this service will ingest.
+    /// get/set - An array of media type this service will ingest (comma separated).
     /// </summary>
-    public string[] MediaTypes { get; set; } = Array.Empty<string>();
+    public string MediaTypes { get; set; } = "";
```
Fosol, line 12: Arrays are additive, not subtractive, which results in unexpected configuration override behaviour.  Using a string with a comma-separator is easier.

services/net/.devcontainer/devcontainer.json
```diff
@@ -66,8 +66,8 @@
 	"remoteUser": "vscode",
 	"features": {
 		"git": "latest"
-	},
-  "mounts": [
-    "source=${localWorkspaceFolder}/capture/data,target=/data,type=bind,consistency=cached"
-  ]
+	}
+  // "mounts": [
```
Fosol, line 70: Mount is not required

services/net/content/ContentManager.cs
```diff
@@ -106,17 +129,28 @@ private async Task HandleMessageAsync(ConsumeResult<string, SourceContent> resul
...
+            if (content != null && !String.IsNullOrWhiteSpace(result.Message.Value.FilePath))
+            {
+                // TODO: Handle different storage locations.
+                // Remote storage locations may not be easily accessible by this service.
+                var sourcePath = Path.Join(_options.ClipPath, result.Message.Value.FilePath);
+                if (File.Exists(sourcePath))
+                {
+                    var file = File.OpenRead(sourcePath);
+                    var fileName = Path.GetFileName(sourcePath);
+                    await _api.UploadFileAsync(content.Id, content.Version ?? 0, file, fileName);
```
Fosol, line 150: I don't love this solution as a large file will block the service for a while.  However, it is the simplest way to support remote architectural complexity issues.

tools/scripts/gen-env-files.sh
```diff
@@ -63,6 +63,7 @@ NLP_PORT=$portNlp
 INDEXING_PORT=$portIndexing
 CAPTURE_PORT=$portCapture
 CLIP_PORT=$portClip
+CONTENT_PORT=$portContent
```
Fosol, line 66: If you're running into `.env` issues, just rerun the `make setup`

PR #176, opened 2022-06-22
Fix data source admin

Fosol:
Fixing issue where import settings were not visible on the Data Source Admin page.

image


app/editor/src/features/admin/sources/services/ServiceIngestSettings.tsx
```diff
@@ -18,8 +18,8 @@ export const ServiceIngestSettings: React.FC<IServiceIngestSettingsProps> = () =
   const { values, setFieldValue } = useFormikContext<IDataSourceModel>();
   const [lookups] = useLookup();
 
-  const showKafkaTopic =
-    !!values.connection.serviceType && values.connection.serviceType !== 'stream';
+  // TODO: Need a more selective condition to determine if this section should be visible.
+  const showImportContent = values.connection.serviceType !== 'stream';
```
Fosol, line 22: The downside to the current logic is that it will be displayed when the `serviceType` is not yet defined.

PR #180, opened 2022-06-28
TNO-344: Manual Page Size

jdtoombs:
Splitting off a ticket.

Add ability for TNO admin's to adjust the page size of certain lists in our application.

tno-344


app/editor/src/features/admin/media-types/MediaTypeList.tsx
```diff
@@ -33,6 +34,7 @@ export const MediaTypeList: React.FC = () => {
         <GridTable
           columns={columns}
           header={MediaTypeFilter}
+          manualPageSize={keycloak.hasClaim(Claim.administrator)}
```
jdtoombs, line 37: May not be necessary because you need to be an admin to access this section?

app/editor/src/features/admin/sources/DataSourceLists.tsx
```diff
@@ -35,6 +37,7 @@ export const DataSourceList: React.FC<IDataSourceListProps> = (props) => {
         columns={dataSourceColumns}
         isLoading={!!requests.length}
         sorting={{ sortBy: [{ id: 'id', desc: false }] }}
+        manualPageSize={keycloak.hasClaim(Claim.administrator)}
```
jdtoombs, line 40: Same comment as above, may not be necessary?

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -355,15 +355,16 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
                         </>
                       }
                     >
-                      {active === 'properties' ? (
+                      <Show visible={active === 'properties'}>
```
jdtoombs, line 358: Little adjustment, just find it easier to read

libs/npm/core/src/components/text/Text.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { InputHTMLAttributes } from 'react';
```
jdtoombs, line 1: Added Text to tno-core so we could use it in our package.

sturufous (approved):

app/editor/src/features/admin/sources/DataSourceLists.tsx
Fosol: Not needed

libs/npm/core/src/components/grid-table/Pager.tsx
```diff
@@ -92,43 +100,66 @@ export const Pager: React.FC<IPagerProps> = ({
...
-        variant={ButtonVariant.info}
-        onClick={() => gotoPage(pageCount - 1)}
-        disabled={!canNextPage}
-      >
-        <BiLastPage />
-      </Button>
+        <Show visible={!!manualPageSize}>
+          <Text
+            className="page-size"
+            width="5em"
```
Fosol, line 144: Always inline styling is horrible.  It's hard to override.  Put it in the styled component.

libs/npm/core/src/components/text/Text.tsx
Fosol: Follow same rules for location as our main library `/components/form/text`.
Also remove from our main library or we won't maintain them both.
Eventually all our form components will be moved over, but we really need to figure out how to test without publishing.

Fosol (changes requested):

libs/npm/core/src/components/text/Text.tsx
jdtoombs: Currently doing that, also realized I left in the old component. So updating it all to use the newly moved one!

Fosol (approved):

PR #181, opened 2022-06-28
TNO-386 Add Purge Service

Fosol:
We now have a simple Purge Service that uses a CronJob to delete folders (based on convention) that are older than two days.

The CronJob is configured to run nightly and executes a shell script file.

image

### DIFF
--- a/openshift/kustomize/cron/base/config-map.yaml
+++ b/openshift/kustomize/cron/base/config-map.yaml
@@ -1,0 +1,62 @@
+---
+# Configuration settings
+kind: ConfigMap
+apiVersion: v1
+metadata:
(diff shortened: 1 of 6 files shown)


openshift/kustomize/cron/base/config-map.yaml
```diff
@@ -0,0 +1,62 @@
...
+          echo "Keep - $d"
+        fi
+
+      done
+
+      echo "************************************"
+    }
+
+    purge_folders "/data/capture" 2
+    purge_folders "/data/clip" 2
```
Fosol, line 62: Purge two different folders.

PR #182, opened 2022-06-29
Fix Capture Service

Fosol:
Ingest services using the Command Service were not cleaning up when a premature exit occurred on the process.  For example when `ffmpeg` received a 404 it would exit, however the Capture and Clip Service would keep a reference to the process that ran.  This process then would continue to be reused even when it should no longer be used.

Now an exit will remove the referenced process so that a new process can be created.

### DIFF
--- a/services/net/command/CommandAction`.cs
+++ b/services/net/command/CommandAction`.cs
@@ -96,14 +96,14 @@ public abstract class CommandAction<TOptions> : IngestAction<TOptions>
     /// </summary>
     /// <param name="process"></param>
     /// <exception cref="Exception"></exception>
-    protected void RunProcess(System.Diagnostics.Process process)
+    protected void RunProcess(ICommandProcess process)
     {
-        var cmd = process.StartInfo.Arguments;
+        var cmd = process.Process.StartInfo.Arguments;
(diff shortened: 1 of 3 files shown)


services/net/command/CommandProcess.cs
```diff
@@ -0,0 +1,40 @@
+namespace TNO.Services.Command;
+
+/// <summary>
+/// CommandProcess class, provides a way to store a process and related information.
+/// </summary>
+public class CommandProcess : ICommandProcess
```
Fosol, line 6: A wrapper object to keep additional information about the process.

services/net/command/CommandAction`.cs
```diff
@@ -209,6 +212,12 @@ protected async Task OnExitedAsync(object? sender, IDataSourceIngestManager mana
             {
                 this.Logger.LogDebug("Service command '{cmd}' exited", cmd);
             }
+
+            // The process has exited, remove it from the manager so that it can get recreated again.
+            var key = process.StartInfo.Verb;
```
Fosol, line 217: Clean up so that a new process will be created the next time a run is executed.

PR #183, opened 2022-06-30
TNO-350 Add DB Triggers

Fosol:
Adding triggers to the `Content` and `ContentReference` tables so that it can be tracked when their status changes.  Presently it doesn't track any other types of edits, but in theory could be updated to handle more complex scenarios.

> You don't need this, but if you want the triggers applied `make renew`


app/editor/package.json
```diff
@@ -1,129 +1,129 @@
 {
-    "name": "tno-editor-app",
-    "description": "Today's News Online Editor application",
-    "version": "0.1.0",
-    "license": "Apache-2.0",
-    "homepage": "/",
-    "packageManager": "yarn@3.2.0",
-    "dependencies": {
-        "@fortawesome/fontawesome-svg-core": "^6.1.1",
```
Fosol, line 9: @jdtoombs The app should have fixed versions otherwise when it gets installed in other environments it could be different than our local.

app/editor/src/features/admin/sources/index.ts
```diff
@@ -3,7 +3,7 @@ export * from './DataSourceActions';
 export * from './DataSourceDetails';
 export * from './DataSourceFilter';
 export * from './DataSourceForm';
-export * from './DataSourceLists';
+export * from './DataSourceList';
```
Fosol, line 6: Renamed to be correct

libs/npm/core/src/components/grid-table/GridTable.tsx
```diff
@@ -196,8 +196,9 @@ export const GridTable = <T extends object>({
   }, [onChangePage, pageIndex, pageSize, currentPage]);
 
   React.useEffect(() => {
-    onChangeSort(sortBy);
-  }, [onChangeSort, sortBy]);
+    if (sortBy.length !== initialSortBy.length || !sortBy.every((v, i) => v === initialSortBy[i]))
```
Fosol, line 199: Not sure if this will work, still need to test.  For some reason react-table is spamming changes to `sortBy` even though nothing changes.  I'm not sure why it does this.

jdtoombs (approved):

PR #184, opened 2022-07-04
TNO-239: User filter / sorting / styling

jdtoombs:
Putting up as draft while I polish up here

- server side sorting on user table
- server side filtering on user table
- add ellipsis on columns
- minor styling

### DIFF
--- a/app/editor/src/features/admin/users/UserFilter.tsx
+++ b/app/editor/src/features/admin/users/UserFilter.tsx
@@ -1,23 +1,25 @@
 import { IconButton, OptionItem, Select } from 'components/form';
-import { IUserFilter, UserStatus, UserStatusName } from 'hooks';
+import { UserStatus, UserStatusName } from 'hooks';
 import React, { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { useUsers } from 'store/hooks/admin';
 import { FieldSize, Text } from 'tno-core';
 import { Row } from 'tno-core/dist/components/flex';
 
+import { IUserListFilter } from './interfaces/IUserListFilter';
+
 interface IUserFilterProps {}
 
 export const UserFilter: React.FC<IUserFilterProps> = () => {
   const [{ userFilter }, { storeFilter }] = useUsers();
-  const [filter, setFilter] = useState<IUserFilter>(userFilter);
+  const [filter, setFilter] = useState<IUserListFilter>(userFilter);
   const statusOptions = [
-    new OptionItem<number>(UserStatusName.Preapproved, UserStatus.Preapproved),
-    new OptionItem<number>(UserStatusName.Approved, UserStatus.Approved),
-    new OptionItem<number>(UserStatusName.Activated, UserStatus.Activated),
-    new OptionItem<number>(UserStatusName.Authenticated, UserStatus.Authenticated),
-    new OptionItem<number>(UserStatusName.Denied, UserStatus.Denied),
-    new OptionItem<number>(UserStatusName.Requested, UserStatus.Requested),
+    new OptionItem(UserStatusName.Preapproved, UserStatus.Preapproved),
+    new OptionItem(UserStatusName.Approved, UserStatus.Approved),
+    new OptionItem(UserStatusName.Activated, UserStatus.Activated),
+    new OptionItem(UserStatusName.Authenticated, UserStatus.Authenticated),
+    new OptionItem(UserStatusName.Denied, UserStatus.Denied),
+    new OptionItem(UserStatusName.Requested, UserStatus.Requested),
   ];
 
   const navigate = useNavigate();
@@ -37,6 +39,7 @@ export const UserFilter: React.FC<IUserFilterProps> = () => {
           }}
           placeholder="Search by keyword"
           name="keyword"
+          value={filter.keyword}
         />
         <Select
           onChange={(e: any) => {
@@ -46,13 +49,15 @@ export const UserFilter: React.FC<IUserFilterProps> = () => {
           options={statusOptions}
           name="status"
           placeholder="Search by status"
+          value={statusOptions.find((s) => s.value === Number(filter.status))}
         />
         <Text
           onChange={(e) => {
-            setFilter({ ...filter, role: e.target.value });
+            setFilter({ ...filter, roles: [e.target.value] });
           }}
           name="role"
           placeholder="Search by role"
+          value={filter.roles}
         />
         <IconButton
           iconType="search"
@@ -63,7 +68,7 @@ export const UserFilter: React.FC<IUserFilterProps> = () => {
         <IconButton
           iconType="reset"
           onClick={() => {
-            setFilter({ sort: [] });
+            setFilter({ sort: [], roles: [], keyword: '', status: undefined });
             storeFilter({ sort: [] });
           }}
         />
--- a/app/editor/src/features/admin/users/UserFilter.tsx
+++ b/app/editor/src/features/admin/users/UserFilter.tsx
@@ -49,15 +49,15 @@ export const UserFilter: React.FC<IUserFilterProps> = () => {
           options={statusOptions}
           name="status"
           placeholder="Search by status"
-          value={statusOptions.find((s) => s.value === Number(filter.status))}
+          value={statusOptions.find((s) => s.value === Number(filter.status)) || ''}
         />
         <Text
           onChange={(e) => {
-            setFilter({ ...filter, roles: [e.target.value] });
+            setFilter({ ...filter, roleName: e.target.value });
           }}
           name="role"
           placeholder="Search by role"
-          value={filter.roles}
+          value={filter.roleName}
         />
         <IconButton
           iconType="search"
@@ -68,7 +68,7 @@ export const UserFilter: React.FC<IUserFilterProps> = () => {
         <IconButton
           iconType="reset"
           onClick={() => {
-            setFilter({ sort: [], roles: [], keyword: '', status: undefined });
+            setFilter({ sort: [], roleName: '', keyword: '', status: undefined });
             storeFilter({ sort: [] });
           }}
         />
--- a/app/editor/src/features/admin/users/UserFilter.tsx
+++ b/app/editor/src/features/admin/users/UserFilter.tsx
@@ -1,5 +1,5 @@
 import { IconButton, OptionItem, Select } from 'components/form';
-import { UserStatus, UserStatusName } from 'hooks';
+import { UserStatusName } from 'hooks';
 import React, { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { useUsers } from 'store/hooks/admin';
@@ -13,13 +13,14 @@ interface IUserFilterProps {}
 export const UserFilter: React.FC<IUserFilterProps> = () => {
   const [{ userFilter }, { storeFilter }] = useUsers();
   const [filter, setFilter] = useState<IUserListFilter>(userFilter);
+
   const statusOptions = [
-    new OptionItem(UserStatusName.Preapproved, UserStatus.Preapproved),
-    new OptionItem(UserStatusName.Approved, UserStatus.Approved),
-    new OptionItem(UserStatusName.Activated, UserStatus.Activated),
-    new OptionItem(UserStatusName.Authenticated, UserStatus.Authenticated),
-    new OptionItem(UserStatusName.Denied, UserStatus.Denied),
-    new OptionItem(UserStatusName.Requested, UserStatus.Requested),
+    new OptionItem('Preapproved', UserStatusName.Preapproved),
+    new OptionItem('Approved', UserStatusName.Approved),
+    new OptionItem('Activated', UserStatusName.Activated),
+    new OptionItem('Authenticated', UserStatusName.Authenticated),
+    new OptionItem('Denied', UserStatusName.Denied),
+    new OptionItem('Requested', UserStatusName.Requested),
   ];
 
   const navigate = useNavigate();
@@ -49,7 +50,7 @@ export const UserFilter: React.FC<IUserFilterProps> = () => {
           options={statusOptions}
           name="status"
           placeholder="Search by status"
-          value={statusOptions.find((s) => s.value === Number(filter.status)) || ''}
+          value={statusOptions.find((s) => s.value === filter.status) || ''}
         />
         <Text
           onChange={(e) => {
--- a/app/editor/src/features/admin/users/UserFilter.tsx
+++ b/app/editor/src/features/admin/users/UserFilter.tsx
@@ -1,14 +1,25 @@
-import { IconButton } from 'components/form';
-import React from 'react';
+import { IconButton, OptionItem, Select } from 'components/form';
+import { IUserFilter, UserStatus, UserStatusName } from 'hooks';
+import React, { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
-import { Row, Text } from 'tno-core';
+import { useUsers } from 'store/hooks/admin';
+import { FieldSize, Text } from 'tno-core';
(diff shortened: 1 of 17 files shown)


app/editor/src/features/admin/users/UserFilter.tsx
```diff
@@ -1,14 +1,25 @@
...
 
-interface IUserFilterProps {
-  setGlobalFilter: (filterValue: any) => void;
-}
+interface IUserFilterProps {}
+
+export const UserFilter: React.FC<IUserFilterProps> = () => {
+  const [{ userFilter }, { storeFilter }] = useUsers();
+  const [filter, setFilter] = useState<IUserFilter>(userFilter);
+  const statusOptions = [
```
Fosol, line 14: I'm curious what's going on here?  Why the mix of number and string?  The API is currently configured to convert enum values to string when serializing.

app/editor/src/features/admin/users/UserList.tsx
```diff
@@ -11,30 +14,62 @@ import { UserFilter } from './UserFilter';
...
+    [findUsers],
+  );
+
   React.useEffect(() => {
-    if (!users.items.length) {
-      api.findUsers({ isSystemAccount: false }).then((data) => {
-        setPage(new Page(data.page - 1, data.quantity, data.items, data.total));
-      });
+    // Only make a request if the user has been set.
+    if (isUserReady()) {
```
Fosol, line 54: Why does it matter if the user is ready at this point?  Is it because the filter may not container their id at this point?

app/editor/src/features/content/list-view/ContentListView.tsx
```diff
@@ -37,6 +37,7 @@ export const ContentListView: React.FC = () => {
   const fetch = React.useCallback(
     async (filter: IContentListFilter) => {
       try {
+        alert('try');
```
Fosol, line 40: remove

libs/net/dal/Models/UserFilter.cs
```diff
@@ -32,6 +34,8 @@ public UserFilter(Dictionary<string, StringValues> queryParams) : base(queryPara
         this.IsEnabled = filter.GetBoolNullValue(nameof(this.IsEnabled));
         this.Status = filter.GetEnumNullValue<Entities.UserStatus>(nameof(this.Status));
         this.IsSystemAccount = filter.GetBoolNullValue(nameof(this.IsSystemAccount));
+        this.Keyword = filter.GetStringValue(nameof(this.Keyword));
+        // this.Roles = filter.GetStringArrayValue(nameof(this.Sort));
```
Fosol, line 38: You'll need to extract roles?

libs/net/dal/Models/UserFilter.cs
```diff
@@ -15,6 +14,9 @@ public class UserFilter : PageFilter
     public Entities.UserStatus? Status { get; set; }
     public bool? IsSystemAccount { get; set; }
     public string[] Sort { get; set; } = Array.Empty<string>();
+    public string? Keyword { get; set; }
+    public IEnumerable<Entities.Role> Roles { get; set; } = Array.Empty<Entities.Role>();
```
Fosol, line 18: I don't think you should be filtering on role objects.  You should just be filtering on role names.

libs/net/dal/Models/UserFilter.cs
```diff
@@ -32,6 +34,8 @@ public UserFilter(Dictionary<string, StringValues> queryParams) : base(queryPara
         this.IsEnabled = filter.GetBoolNullValue(nameof(this.IsEnabled));
         this.Status = filter.GetEnumNullValue<Entities.UserStatus>(nameof(this.Status));
         this.IsSystemAccount = filter.GetBoolNullValue(nameof(this.IsSystemAccount));
+        this.Keyword = filter.GetStringValue(nameof(this.Keyword));
```
Fosol, line 37: I suspect there should be a null value handled here.

libs/net/dal/Services/UserService.cs
```diff
@@ -40,14 +40,19 @@ public IPaged<User> Find(UserFilter filter)
             query = query.Where(c => EF.Functions.Like(c.FirstName.ToLower(), $"{filter.FirstName.ToLower()}%"));
         if (!String.IsNullOrWhiteSpace(filter.LastName))
             query = query.Where(c => EF.Functions.Like(c.LastName.ToLower(), $"{filter.LastName.ToLower()}%"));
-
+        if (!String.IsNullOrWhiteSpace(filter.Keyword))
```
Fosol, line 43: You're converting keyword to lower multiple times.  Do it once.

app/editor/src/features/admin/users/UserList.tsx
```diff
@@ -11,30 +14,62 @@ import { UserFilter } from './UserFilter';
...
+        storeFilter({ ...userFilter, sort: sorts });
+      }
+    },
+    [storeFilter, userFilter],
+  );
+
+  const fetch = React.useCallback(
+    async (filter: IUserFilter) => {
+      try {
+        const data = await findUsers(makeUserFilter(filter));
```
Fosol, line 39: `IUserFilter` is already the correct shape for the API.  Why do you need to now convert it?

app/editor/src/features/content/list-view/utils/makeUserFilter.ts
```diff
@@ -0,0 +1,28 @@
+import { IUserFilter } from 'hooks';
+
+export const makeUserFilter = (filter: IUserFilter) => {
```
Fosol, line 3: It doesn't really make sense to pass an `IUserFilter` and return ` IUserFilter`.

app/editor/src/features/content/list-view/utils/makeUserFilter.ts
```diff
@@ -0,0 +1,28 @@
...
+  return {
+    username: filter.username,
+    status: filter.status,
+    role: filter.role,
```
Fosol, line 7: The API isn't expected a `role` it's expecting an array of `roles`

app/editor/src/hooks/api-editor/interfaces/IUserFilter.ts
```diff
@@ -8,5 +6,8 @@ export interface IUserFilter {
   lastName?: string;
   isEnabled?: boolean;
   isSystemAccount?: boolean;
-  status?: UserStatusName;
+  status?: string;
+  sort: any;
+  role?: string;
```
Fosol, line 11: This interface is representing what the API is expected.  The API is expected the enum `status` values.  `sort` should follow the same structure you can see in the `IContentFilter`.  `role` doesn't exist in the API, it should be `roles` and an array of string.

app/editor/src/features/admin/users/UserList.tsx
```diff
@@ -11,30 +14,62 @@ import { UserFilter } from './UserFilter';
...
     <styled.UserList>
       <FormPage>
         <PagedTable
           columns={columns}
           header={UserFilter}
+          sorting={{ sortBy: userFilter.sort }}
           isLoading={!!requests.length}
           page={page}
           onRowClick={(row) => navigate(`${row.original.id}`)}
+          onChangeSort={handleChangeSort}
```
Fosol, line 72: Did you find that the `onChangeSort` gets called a lot when testing?  I was surprised how often react.table detects a change.

Fosol (changes requested):

app/editor/src/features/admin/users/UserList.tsx
jdtoombs: This was to make sure sort was working correctly (getting the appropriate column etc)

app/editor/src/features/admin/users/UserList.tsx
jdtoombs: Copy and paste error

app/editor/src/features/admin/users/UserList.tsx
jdtoombs: Yes, couldn't figure out all the triggers...

libs/net/dal/Services/UserService.cs
```diff
@@ -40,6 +40,11 @@ public IPaged<User> Find(UserFilter filter)
             query = query.Where(c => EF.Functions.Like(c.FirstName.ToLower(), $"{filter.FirstName.ToLower()}%"));
         if (!String.IsNullOrWhiteSpace(filter.LastName))
             query = query.Where(c => EF.Functions.Like(c.LastName.ToLower(), $"{filter.LastName.ToLower()}%"));
+        if (!String.IsNullOrWhiteSpace(filter.Keyword))
+            query = query.Where(c => EF.Functions.Like(c.Username.ToLower(), $"{filter.Keyword.ToLower()}%") || EF.Functions.Like(c.Email.ToLower(), $"{filter.Keyword.ToLower()}%") || EF.Functions.Like(c.FirstName.ToLower(),
+            $"{filter.Keyword.ToLower()}%") || EF.Functions.Like(c.LastName.ToLower(), $"{filter.Keyword.ToLower()}%"));
+        if (!String.IsNullOrWhiteSpace(filter.RoleName))
+            query = query.Where(c => c.Roles.Any(r => r.Name.ToLower() == filter.RoleName.ToLower()));
```
jdtoombs, line 47: This doesn't seem to be working, is there something obvious I am doing wrong here? @Fosol

app/editor/src/features/admin/users/UserFilter.tsx
```diff
@@ -22,18 +35,41 @@ export const UserFilter: React.FC<IUserFilterProps> = ({ setGlobalFilter }) => {
...
+          onClick={() => {
+            storeFilter(filter);
+          }}
         />
         <IconButton
           iconType="reset"
           onClick={() => {
-            setFilter('');
-            setGlobalFilter('');
+            setFilter({ sort: [], roles: [''], keyword: '', status: '' });
```
Fosol, line 71: Why `roles: ['']` instead of `roles: []`?

sturufous (approved):

app/editor/src/features/admin/users/UserFilter.tsx
```diff
@@ -1,14 +1,27 @@
...
-  setGlobalFilter: (filterValue: any) => void;
-}
+import { IUserListFilter } from './interfaces/IUserListFilter';
+
+interface IUserFilterProps {}
+
+export const UserFilter: React.FC<IUserFilterProps> = () => {
+  const [{ userFilter }, { storeFilter }] = useUsers();
+  const [filter, setFilter] = useState<IUserListFilter>(userFilter);
+  const statusOptions = [
```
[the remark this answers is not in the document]
Fosol: @jdtoombs You're mixing enums up here I believe.  The API will convert enum values into string values.  Can you please confirm what the API is doing?

libs/net/dal/Services/UserService.cs
Fosol: Search by role isn't working?  You'd have to debug the API and see the SQL it's generating.

Fosol (changes requested):

libs/net/dal/Services/UserService.cs
jdtoombs: Fixed it :)

app/editor/src/features/admin/users/UserFilter.tsx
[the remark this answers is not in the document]
jdtoombs: Looks like it is sending the numeric value to the API, it seems to be filtering correctly too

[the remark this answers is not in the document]
Fosol: While the API may understand how to deserialize the number to an enum, it is configured to serialize enum values as strings.  Which means you should be able to just use the string values.  Can you test the string values and see if it works?  It'll make you code a lot cleaner and the filter will be more human readable.

app/editor/src/features/content/list-view/utils/makeUserFilter.ts
```diff
@@ -0,0 +1,31 @@
+import { IUserListFilter } from 'features/admin/users/interfaces/IUserListFilter';
+import { IUserFilter } from 'hooks';
+
+import { ISortBy } from '../interfaces';
+
+export const makeUserFilter = (filter: IUserListFilter): IUserFilter => {
```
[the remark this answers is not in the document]
Fosol: You can do this to clean up your code.

Fosol (changes requested):

app/editor/src/features/admin/users/UserFilter.tsx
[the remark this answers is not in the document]
jdtoombs: I tried passing it just the string values; however, the `Select` component will throw an error as it is expecting `OptionItems` which require more than just a string value. Is this something I should try to tweak? I noticed the other areas in which we use the Select component they follow the same format as I am doing above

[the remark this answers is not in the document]
Fosol: Do we do this exact type of filter with content?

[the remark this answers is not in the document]
jdtoombs: Oh wait maybe I am misunderstanding, I took this as you wanting to pass ["item 1", "item 2"] but are you just wanting the value sent to the api to be the string? So it could be OptionItem("label", stringValue)?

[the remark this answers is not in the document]
Fosol: The filter sent to the API should be the string value by itself "value".  Which the API will understand and deserialize into the appropriate enum.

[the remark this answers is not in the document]
jdtoombs: Think I may have it now, I was overcomplicating it in my head thinking you just wanted to provide the label, with no value.. which obviously didn't click for me at first lol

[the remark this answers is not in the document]
Fosol: Won't `UserStatusName.Preapproved === 'Preapproved'`?  Why do you need to manually add the string values?  Don't we have an existing method that converts enums to options?

Fosol (changes requested):

libs/net/dal/Models/UserFilter.cs
Fosol, line 37: Shouldn't this be nullable?  `filter.GetStringNullValue(...)`?

libs/net/dal/Models/UserFilter.cs
```diff
@@ -32,6 +34,8 @@ public UserFilter(Dictionary<string, StringValues> queryParams) : base(queryPara
...
+        this.RoleName = filter.GetStringValue(nameof(this.RoleName));
```
Fosol, line 38: Shouldn't this be nullable?  `filter.GetStringNullValue(...)`?

Fosol (changes requested):
Still seeing a few things that need addressing

libs/net/dal/Models/UserFilter.cs
jdtoombs: We don't have an extension method for `GetStringNullValue`, should I create one or do we not need one for strings?

Fosol: Hmmm looks like the `GetStringValue` returns a nullable string.  Not ideal, but normal before we had the `?` nullable syntax option.
We should probably update this, but it would require changes unrelated to this PR.  Let's leave it now as tech debt.

libs/net/dal/Services/UserService.cs
Fosol: If you have `filter.Keywork.ToLower()` multiple times you are reducing performance.  You should only do this once.

Fosol (changes requested):

jdtoombs:
@Fosol updated again... might be the one 🤔

Fosol (approved):

PR #185, opened 2022-07-05
TNO-45 Update error handling

Fosol:
Improved error messages returned from API.  

Update Editor App so that errors on other tabs are identified when validation occurs.

Provided a way to test `tno-core` npm package updates locally before publishing to npm!

Within the `/libs/npm/core` devcontainer there is now a `make pack` command.  This creates a `tgz` package and installs it locally.  Then you need to `make refresh n=editor` to test locally.

image


app/editor/src/features/admin/sources/DataSourceForm.tsx
```diff
@@ -58,6 +59,10 @@ export const DataSourceForm: React.FC<IDataSourceProps> = (props) => {
     } catch {}
   };
 
+  const hasErrors = (errors: any, props: string[]) => {
```
Fosol, line 62: I don't love this solution, but it's all I could think of today.  Essentially just checks if an error property is positive.

app/editor/src/features/admin/sources/DataSourceForm.tsx
```diff
@@ -82,16 +87,34 @@ export const DataSourceForm: React.FC<IDataSourceProps> = (props) => {
           handleSubmit(values);
           setSubmitting(false);
         }}
+        validateOnBlur={true}
```
Fosol, line 90: On blur so that validation occurs and is cleared after it has been run.  I don't love this solution either because it runs validation early and the tabs go red if they have required fields.  Once again I don't have a better way to do this presently.

libs/npm/core/scripts/package.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
jdtoombs, line 1: 🎉

jdtoombs (approved):

PR #187, opened 2022-07-06
TNO-385: Navbar hover

jdtoombs:
Hovering the navbar item should change the menu appropriately, and the sub menu should not shift down the whole page.

navbar


sturufous (approved):
Love it!

app/editor/src/features/navbar/NavBar.tsx
```diff
@@ -1,38 +1,43 @@
...
 import { Row } from 'tno-core/dist/components/flex/row';
-import { NavBarGroup, NavBarItem } from 'tno-core/dist/components/navbar';
 
 /**
  * The navigation bar that is used throughout the TNO editor application. Add or remove navigation bar items here.
  */
 export const NavBar: React.FC = () => {
-  const location = useLocation();
+  const [activeHover, setActiveHover] = useState<'admin' | 'editor' | ''>('');
+  console.log(activeHover);
```
Fosol, line 10: Remove

app/editor/src/features/navbar/NavBar.tsx
```diff
@@ -1,38 +1,43 @@
...
+import { Claim, NavBarGroup, NavBarItem, Show } from 'tno-core';
 import { Row } from 'tno-core/dist/components/flex/row';
-import { NavBarGroup, NavBarItem } from 'tno-core/dist/components/navbar';
 
 /**
  * The navigation bar that is used throughout the TNO editor application. Add or remove navigation bar items here.
  */
 export const NavBar: React.FC = () => {
-  const location = useLocation();
+  const [activeHover, setActiveHover] = useState<'admin' | 'editor' | ''>('');
```
Fosol, line 9: What's the value of hardcoded values here?  Why not just use string?

app/editor/src/features/navbar/NavBar.tsx
```diff
@@ -1,38 +1,43 @@
...
+              claim={Claim.administrator}
+            />
+          </div>
         </Row>
       </NavBarGroup>
-      <NavBarGroup className="navbar">
-        {showEditor && activeHover && (
-          <Row>
+      <NavBarGroup hover className="navbar">
+        <Row hidden={activeHover === ''}>
```
Fosol, line 35: Might be cleaner to `!activeHover`

libs/npm/core/src/components/navbar/NavBarGroup.tsx
```diff
@@ -2,14 +2,20 @@ import React from 'react';
...
  * The element that groups the various navigation bar items together.
  * @param children the navigation bar items
  * @returns navigation bar group
  */
-export const NavBarGroup: React.FC<React.HTMLAttributes<HTMLDivElement>> = ({
-  children,
-  ...rest
-}) => {
-  return <styled.NavBarGroup {...rest}>{children}</styled.NavBarGroup>;
+export const NavBarGroup: React.FC<INavBarGroupProps> = ({ children, hover, ...rest }) => {
```
Fosol, line 15: You don't need to extract `hover` if you're just going to place it on the same line as `{...rest}`

app/editor/src/features/navbar/NavBar.tsx
```diff
@@ -1,38 +1,43 @@
...
   return (
-    <div onMouseOver={() => setActiveHover(true)} onMouseLeave={() => setActiveHover(false)}>
+    <div onMouseLeave={() => setActiveHover('')}>
       <NavBarGroup className="navbar">
         <Row>
-          <NavBarItem navigateTo="/contents" label="Editor" claim={Claim.editor} />
-          <NavBarItem navigateTo="/admin" label="Admin" claim={Claim.administrator} />
+          <div className="editor" onMouseOver={() => setActiveHover('editor')}>
+            <NavBarItem
+              activeHoverTab={activeHover}
```
Fosol, line 18: Is the `activeHoverTab` prop to keep track of which parent is active?

Fosol (changes requested):

app/editor/src/features/navbar/NavBar.tsx
jdtoombs: Correct

app/editor/src/features/navbar/NavBar.tsx
Fosol: I'm wondering how we can convert this to a generic component, but doesn't need to be done now.

Fosol (approved):

PR #191, opened 2022-07-08
Functioning CP News syndication service

sturufous:
I've been having issues with api requests over the past week which has introduced delays. The symptoms are: when reading data sources the API connection times out with:

TNO.Services.Syndication.SyndicationManager[0]
Failed to fetch data sources for media type
System.Threading.Tasks.TaskCanceledException: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.

The error message in question is reported in DataSourceManager`.GetDataSourcesAsync()

The only way I could solve this for the first few days was by rebooting my system. I've since discovered I can fix it by restarting Docker Desktop. This happens once every 8 (or so) debug sessions.

The code in this PR has one known flaw - because the CP News articles don't have a year in the published date I default to the current year. This will return inaccurate published dates on January 1st. I could add logic to check for this condition but left it as-is for now.


app/editor/src/features/admin/sources/connections/Syndication.tsx
```diff
@@ -34,6 +43,14 @@ export const Syndication: React.FC = (props) => {
           setFieldValue('connection.import', e.currentTarget.checked);
         }}
       />
+      <FormikCheckbox
+        label="Fetch Content Body Separately"
+        name="connection.fetch-content"
+        tooltip="Whether content body is located remotely"
+        onChange={(e) => {
+          setFieldValue('connection.fetch-content', e.currentTarget.checked);
```
Fosol, line 51: Change name of property to `connection.fetchContent` so that it's a valid property name.

services/net/syndication/SyndicationAction.cs
```diff
@@ -80,6 +84,19 @@ private async Task ImportFeedAsync(IDataSourceIngestManager dataSource, Syndicat
         {
             try
             {
+                // Fetch content body separately if required
+                if (bool.Parse(dataSource.DataSource.GetConnectionValue("fetch-content")))
+                {
+                    var link = item.Links.FirstOrDefault(l => l.RelationshipType == "alternate")?.Uri.ToString() ?? "";
+                    var content =  await this.GetContentAsync(link);
```
Fosol, line 91: Should only attempt this if the `Uri` is valid.

services/net/syndication/SyndicationAction.cs
```diff
@@ -124,6 +141,39 @@ private async Task ImportFeedAsync(IDataSourceIngestManager dataSource, Syndicat
         }
     }
 
+    /// <summary>
+    /// Make an AJAX request to fetch the CP News article identified by url.
+    /// </summary>
+    /// <param name="url">The web location of a CP News article</param>
+    /// <returns>An HTML formatted news article</returns>
+    private async Task<string> GetContentAsync(string url)
```
Fosol, line 149: Change param from `string` to `Uri` to ensure a valid value.

Fosol (changes requested):

services/net/syndication/SyndicationAction.cs
```diff
@@ -80,6 +84,26 @@ private async Task ImportFeedAsync(IDataSourceIngestManager dataSource, Syndicat
         {
             try
             {
+                // Fetch content body separately if required
+                if (bool.Parse(dataSource.DataSource.GetConnectionValue("fetchContent")))
```
Fosol, line 88: Will this fail if the `fetchContent` property does not exist?

services/net/syndication/SyndicationAction.cs
```diff
@@ -80,6 +84,26 @@ private async Task ImportFeedAsync(IDataSourceIngestManager dataSource, Syndicat
...
+                {
+                    var link = item.Links.FirstOrDefault(l => l.RelationshipType == "alternate")?.Uri;
+                    if (Uri.IsWellFormedUriString(link.ToString(), UriKind.Absolute))
+                    {
+                        var content =  await this.GetContentAsync(link);
+                        var date = await GetPubDateTimeAsync(content, dataSource.DataSource);
+
+                        item.Id = link.ToString();
+                        item.PublishDate = date;
+                        content = StringExtensions.SanitizeContent(content);
```
Fosol, line 98: FYI - Extensions can also be used the following way `content.SantizeContent()`, but isn't required.

Fosol (approved):

PR #199, opened 2022-07-28
TNO-315: React 18 Warning

jdtoombs:
`StrictMode` creates an infinite loop with `Keycloak`, to get the benefits of React 18 we can temporarily remove `StrictMode` until a fix is pushed for this. Not sure which one is more beneficial for us.

I also tried moving StrictMode to a different location, which seemed to work; however, an error toast was constantly appearing saying "Data may have changed, please refresh to confirm".

Also after the change `UserInfo` was causing an infinite loop with the app dependency.

### DIFF
--- a/app/editor/src/features/login/UserInfo.tsx
+++ b/app/editor/src/features/login/UserInfo.tsx
@@ -26,7 +26,8 @@ export const UserInfo: React.FC<IUserInfoProps> = ({ children }) => {
       appStore.storeUserInfo();
       setInit(true);
     }
-  }, [app, appStore, keycloak.authenticated, init]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [appStore, keycloak.authenticated, init]);
 
   return <>{children}</>;
 };
--- a/app/editor/src/index.tsx
+++ b/app/editor/src/index.tsx
@@ -1,7 +1,7 @@
 import './index.scss';
 
 import React from 'react';
-import ReactDOM from 'react-dom';
+import ReactDOM from 'react-dom/client';
 import { Provider } from 'react-redux';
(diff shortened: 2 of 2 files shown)


app/editor/src/index.tsx
```diff
@@ -11,21 +11,21 @@ import App from './App';
 import css from './css/_variables.module.scss';
 import * as serviceWorker from './serviceWorker';
 
+/** TODO: Enable React.StrictMode when keycloak fix is released. */
```
Fosol, line 14: I suspect no one is maintaining keycloak-js

app/editor/src/features/login/UserInfo.tsx
```diff
@@ -26,7 +26,8 @@ export const UserInfo: React.FC<IUserInfoProps> = ({ children }) => {
       appStore.storeUserInfo();
       setInit(true);
     }
-  }, [app, appStore, keycloak.authenticated, init]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [appStore, keycloak.authenticated, init]);
```
Fosol, line 30: This worries me as we keep getting forced to remove dependencies.  The `app` is responsible for determining if the user info is loaded.  I wonder if removing this will result in some odd bugs with new users.

Fosol: Have you tested after clearing cache and local storage?

jdtoombs: Seems to work after clearing local storage and cache, just tried a few tests leaving app in as well and unfortunately couldn't escape the infinite loop

Fosol:
Going to postpone this, as I believe turning of strict mode will lead to worse issues down the road.

PR #202, opened 2022-08-03
TNO-406

jdtoombs:
Give the page context of where the mouse is located so hitting enter does not trigger the search when the user clicks enter while writing a summary.


Fosol (approved):
Pretty simple solution.

PR #203, opened 2022-08-03
TNO-409 Add Image Ingest Service

Fosol:
This is not a fully functional Image Ingest Service.  It is only a base for getting started.

image


app/editor/src/features/admin/sources/services/ServiceIngestSettings.tsx
```diff
@@ -62,57 +62,61 @@ export const ServiceIngestSettings: React.FC<IServiceIngestSettingsProps> = () =
       </p>
       <Row colGap="1em" nowrap>
         <Col flex="1 1">
```
Fosol, line 64: Flipped the columns so that source settings are before destination settings.

app/editor/src/features/login/LoginPanel.tsx
```diff
@@ -3,7 +3,9 @@ import { Button, Col, useKeycloakWrapper } from 'tno-core';
 export const LoginPanel: React.FC = () => {
   const keycloak = useKeycloakWrapper();
 
-  const isLocal = window.location.host.startsWith('localhost');
+  const isLocal =
+    window.location.host.startsWith('localhost') ||
+    window.location.host.startsWith('host.docker.internal');
```
Fosol, line 8: Firefox isn't liking the cookie if it comes from a different domain anymore.  Which requires using `host.docker.internal` instead of `localhost`.

libs/net/dal/Migrations/Initial/Up/PostUp/00-ContentType.sql
```diff
@@ -34,6 +34,14 @@ INSERT INTO public.content_type (
   , ''
   , DEFAULT_USER_ID
   , ''
+), (
+  'Frontpage'
```
Fosol, line 38: New Content Type which will allow us to create a different form for this.

jdtoombs (approved):

PR #205, opened 2022-08-03
Active hover adjustment

jdtoombs:
Noticed a few places where the rows were still all blue, this fixes that.


libs/npm/core/src/components/grid-table/GridTable.tsx
```diff
@@ -225,7 +225,7 @@ export const GridTable = <T extends object>({
             prepareRow(row);
             return (
               <div
-                className={activeId && activeId === (row.original as any).id ? 'active' : ''}
+                className={!!activeId && activeId === (row.original as any).id ? 'active' : ''}
```
Fosol, line 228: Why was this needed?

Fosol (approved):

libs/npm/core/src/components/grid-table/GridTable.tsx
jdtoombs: For some reason on the time log table when no prop was provided it was still triggering.. but it had different behaviour on other pages (like ingestion)

PR #207, opened 2022-08-04
TNO-407 Fixing Syndication Ingest Service

Fosol:
When fetching content the prior process expected to find a date in the body, but it doesn't always exist.  The new process defaults to `DateTime.UtcNow` instead.

### DIFF
--- a/services/net/syndication/SyndicationAction.cs
+++ b/services/net/syndication/SyndicationAction.cs
@@ -180,6 +180,7 @@ public class SyndicationAction : IngestAction<SyndicationOptions>
 
     /// <summary>
     /// Takes a CP News article and extracts the published date/time.
+    /// If no date is found it will default to now.  This isn't ideal, but it's better than a min date.
     /// </summary>
     /// <param name="content"></param>
     /// <param name="dataSource"></param>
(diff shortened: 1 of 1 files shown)


services/net/syndication/SyndicationAction.cs
```diff
@@ -180,23 +180,26 @@ private async Task<string> GetContentAsync(Uri url)
...
-        var dateStr = $"{comps[1]} {comps[0]} {DateTime.Now.Year} {comps[2]} {offset}";
-        var date = DateTime.ParseExact(dateStr, "dd MMM yyyy HH:mm z", CultureInfo.InvariantCulture);
+        if (matches.Any())
+        {
+            var pubDate = matches[0].Groups[1].Value;
+            var comps = pubDate.Split(' ');
+            var timeZoneStr = dataSource.GetConnectionValue("timeZone");
+            var timeZone = TimeZoneInfo.FindSystemTimeZoneById(timeZoneStr);
+            var offset = timeZone.GetUtcOffset(DateTime.Now).Hours; // Handles daylight saving time
+            var dateStr = $"{comps[1]} {comps[0]} {DateTime.Now.Year} {comps[2]} {offset}";
```
Fosol, line 198: This is still not ideal, I need to confirm the comps has 3 parts.

sturufous (approved):

PR #208, opened 2022-08-04
TNO-416: Tone values saving on create not update

jdtoombs:
Previously tone values were only being saved correctly on create

### DIFF
--- a/app/editor/src/features/content/form/CondensedContentForm.tsx
+++ b/app/editor/src/features/content/form/CondensedContentForm.tsx
@@ -112,9 +112,11 @@ export const CondensedContentForm: React.FC<ICondensedContentFormProps> = ({
         // Only new content is initialized.
         values.contentTypeId = contentType;
         values.ownerId = userId;
-        const defaultTonePool = tonePools.find((t) => t.name === 'Default');
-        values.tonePools = !!defaultTonePool ? [{ ...defaultTonePool, value: +values.tone }] : [];
       }
+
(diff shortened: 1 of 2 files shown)


app/editor/src/features/content/form/CondensedContentForm.tsx
```diff
@@ -112,9 +112,11 @@ export const CondensedContentForm: React.FC<ICondensedContentFormProps> = ({
         // Only new content is initialized.
         values.contentTypeId = contentType;
         values.ownerId = userId;
-        const defaultTonePool = tonePools.find((t) => t.name === 'Default');
-        values.tonePools = !!defaultTonePool ? [{ ...defaultTonePool, value: +values.tone }] : [];
       }
+
+      const defaultTonePool = tonePools.find((t) => t.name === 'Default');
+      values.tonePools = !!defaultTonePool ? [{ ...defaultTonePool, value: +values.tone }] : [];
```
Fosol, line 118: I'm curious how this works as changing the `values.tonePools` here won't be recognized by state management because it's not part of a `useEffect`?

jdtoombs: Get's triggered whenever the update/create button is clicked, and the correct value will be managed in `values.tone` with `setFieldValue` from the radio buttons. So whatever value is in `values.tone` at the time of pressing the button will be added!

jdtoombs: and the `toModel` function is ran on the values above

Fosol (approved):

PR #210, opened 2022-08-04
TNO-405: Time log table displaying wrong month

jdtoombs:
`date.getMonth()` is 0 indexed so we have to add +1 for desired output

### DIFF
--- a/app/editor/src/features/content/form/constants/timeLogColumns.tsx
+++ b/app/editor/src/features/content/form/constants/timeLogColumns.tsx
@@ -36,7 +36,9 @@ export const timeLogColumns = (
     Cell: ({ value }: any) => {
       const date = new Date(value);
       return (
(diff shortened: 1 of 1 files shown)


app/editor/src/features/content/form/constants/timeLogColumns.tsx
```diff
@@ -36,7 +36,9 @@ export const timeLogColumns = (
     Cell: ({ value }: any) => {
       const date = new Date(value);
       return (
-        <div className="center">{`${date.getMonth()}/${date.getDate()}/${date.getFullYear()}`}</div>
+        <div className="center">{`${
+          date.getMonth() + 1
```
Fosol, line 40: 0 indexes will get you every time.

Fosol (approved):

PR #213, opened 2022-08-09
TNO-434 Fix Syndication Service

Fosol:
Syndication service was broken when a bug was introduced with the CP Feed enhancements.


services/net/syndication/SyndicationAction.cs
```diff
@@ -148,7 +148,7 @@ private async Task ImportFeedAsync(IDataSourceIngestManager dataSource, Syndicat
     private async Task FetchContent(DataSourceModel dataSource, SyndicationItem item, Uri? link)
     {
         // Fetch content body separately if required
-        if (bool.Parse(dataSource.GetConnectionValue("fetchContent")))
```
Fosol, line 151: This will fail because `DataSource.Connection.fetchContent = undefined` in most cases.

PR #215, opened 2022-08-15
Tone pools bug fix

jdtoombs:
State was out of line with tone pools

### DIFF
--- a/app/editor/src/features/content/form/CondensedContentForm.tsx
+++ b/app/editor/src/features/content/form/CondensedContentForm.tsx
@@ -105,7 +105,7 @@ export const CondensedContentForm: React.FC<ICondensedContentFormProps> = ({
   });
 
   const handleSubmit = async (values: IContentForm) => {
-    var contentResult: IContentModel | null = null;
+    let contentResult: IContentModel | null = null;
     const originalId = values.id;
     try {
       if (!values.id) {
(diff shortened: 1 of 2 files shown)


app/editor/src/features/content/form/CondensedContentForm.tsx
```diff
@@ -105,7 +105,7 @@ export const CondensedContentForm: React.FC<ICondensedContentFormProps> = ({
   });
 
   const handleSubmit = async (values: IContentForm) => {
-    var contentResult: IContentModel | null = null;
+    let contentResult: IContentModel | null = null;
```
jdtoombs, line 108: linting warning about using var

Fosol: If the value never changes use `const`

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -126,9 +126,9 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
       // Now upload the file if it exists.
       if (!!values.file) {
         const uploadResult = await upload(contentResult, values.file);
-        setContent(toForm(uploadResult));
+        setContent(toForm({ ...uploadResult, tonePools: values.tonePools }));
```
Fosol, line 129: I assume there is an issue with arrays or something here.  Can you explain?

app/editor/src/features/content/form/CondensedContentForm.tsx
jdtoombs: Changes on line 124

app/editor/src/features/content/form/ContentForm.tsx
jdtoombs: For some reason when no tone values were assigned to an existing piece of content the array was having issues being assigned the new value into state. Didn't spend too much time on alternative methods to solve this; could look into more if need be

Fosol (approved):

PR #221, opened 2022-08-17
Update Image Service

Fosol:
Updating Image Service logic to only work with data sources with the configured media type (i.e. "Image").

Also adding logic to handle fetching files with private keys.

### DIFF
--- a/services/net/.gitignore
+++ b/services/net/.gitignore
@@ -12,6 +12,7 @@ nupkg/
 
 capture/data
 data
+image/keys
 
 # IDE - VSCode
 .vscode/*
--- a/services/net/image/Config/ImageOptions.cs
+++ b/services/net/image/Config/ImageOptions.cs
@@ -9,6 +9,11 @@ public class ImageOptions : IngestServiceOptions
 {
     #region Properties
     /// <summary>
+    /// get/set - The path to private key files
+    /// </summary>
+    public string PrivateKeysPath { get; set; } = "";
+
+    /// <summary>
     /// get/set - The path to store files.
     /// </summary>
     public string OutputPath { get; set; } = "";
--- a/services/net/image/ImageAction.cs
+++ b/services/net/image/ImageAction.cs
@@ -71,7 +71,7 @@ public class ImageAction : IngestAction<ImageOptions>
         // If the data source is configured to use a schedule then
         var content = CreateContentReference(manager.DataSource);
 
-        // TODO: Fetch image from source data location.  Only continue if the image exists.
+        await FetchImage(manager.DataSource);
 
         var reference = await this.Api.FindContentReferenceAsync(content.Source, content.Uid);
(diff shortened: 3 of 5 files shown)


services/net/.gitignore
```diff
@@ -12,6 +12,7 @@ nupkg/
 
 capture/data
 data
+image/keys
```
Fosol, line 15: Don't add private keys to source

services/net/image/Config/ImageOptions.cs
```diff
@@ -8,6 +8,11 @@ namespace TNO.Services.Image.Config;
 public class ImageOptions : IngestServiceOptions
 {
     #region Properties
+    /// <summary>
+    /// get/set - The path to private key files
+    /// </summary>
+    public string PrivateKeysPath { get; set; } = "";
```
Fosol, line 14: This allows us to change the location if need per environment.

services/net/image/ImageAction.cs
```diff
@@ -105,6 +105,45 @@ public ImageAction(IKafkaMessenger kafka, IApiService api, IOptions<ImageOptions
...
+        throw new NotImplementedException();
+    }
+
+    /// <summary>
+    /// Perform image processing based on configuration.
+    /// </summary>
+    /// <param name="dataSource"></param>
+    /// <returns></returns>
+    /// <exception cref="NotImplementedException"></exception>
+    private async Task ProcessImage(DataSourceModel dataSource)
```
Fosol, line 128: We don't need to do this presently, but this is where the logic would exist.

services/net/image/ImageAction.cs
```diff
@@ -105,6 +105,45 @@ public ImageAction(IKafkaMessenger kafka, IApiService api, IOptions<ImageOptions
...
+        throw new NotImplementedException();
+    }
+
+    /// <summary>
+    ///
+    /// </summary>
+    /// <param name="dataSource"></param>
+    /// <returns></returns>
+    /// <exception cref="NotImplementedException"></exception>
+    private async Task CopyImage(DataSourceModel dataSource)
```
Fosol, line 140: Not sure if file copy will be async.  If it's not, just change the function accordingly.

services/net/image/ImageManager.cs
```diff
@@ -27,4 +28,28 @@ public class ImageManager : DataSourceManager<ImageDataSourceManager, ImageOptio
...
+    /// <returns></returns>
+    public override async Task<IEnumerable<DataSourceModel>> GetDataSourcesAsync()
+    {
+        var dataSources = await base.GetDataSourcesAsync();
+
+        return dataSources.Where(ds => IsImage(ds));
+    }
+
+    /// <summary>
+    /// Determine if the data source is a stream service type.
```
Fosol, line 46: Only image media type is required.

PR #222, opened 2022-08-17
Configure Prometheus client library to use authorization

andrelashley:


Fosol (approved):

Fosol:
I wonder how we can control who has access to the metrics?  It isn't critical atm.

PR #223, opened 2022-08-18
TNO-430: Hidden expand button on combined view

jdtoombs:
Add way to bypass `min-width` on the `FormPage` component, this also returns the combined view to the proper styling


app/editor/src/components/form/formpage/FormPage.tsx
```diff
@@ -3,6 +3,9 @@ import { InputHTMLAttributes } from 'react';
 import * as styled from './styled';
 
 export interface IFormPageProps extends InputHTMLAttributes<HTMLDivElement> {
+  /** bypass the default min width */
+  bypassMinWidth?: boolean;
```
Fosol, line 7: Seems like a odd way to do this.  But presently I can't think of a better way.

Fosol (approved):

PR #228, opened 2022-08-24
TNO-425: Fix the add a tag error

jdtoombs:
As tags behave slightly differently than admin areas with a string id, there was some incorrect behaviour when creating a new one. 

i.e) `useEffect` loop firing when it is not supposed to because `undefined !== 0`

Can now create a new tag as expected, form will submit the name value as the id. Also fixed some copy paste errors I ran into.

### DIFF
--- a/app/editor/src/features/admin/tags/TagListFilter.tsx
+++ b/app/editor/src/features/admin/tags/TagListFilter.tsx
@@ -16,7 +16,7 @@ export const TagListFilter: React.FC<IAdminFilterProps> = ({ setGlobalFilter }) 
         <IconButton
           iconType="plus"
           label={`Add new tag`}
-          onClick={() => navigate(`/admin/tags/0`)}
+          onClick={() => navigate(`/admin/tags/NEW`)}
         />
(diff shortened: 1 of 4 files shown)


app/editor/src/features/admin/tags/TagListFilter.tsx
```diff
@@ -16,7 +16,7 @@ export const TagListFilter: React.FC<IAdminFilterProps> = ({ setGlobalFilter })
         <IconButton
           iconType="plus"
           label={`Add new tag`}
-          onClick={() => navigate(`/admin/tags/0`)}
+          onClick={() => navigate(`/admin/tags/NEW`)}
```
Fosol, line 19: Let's hope they never want to create a NEW tag lol

Fosol (approved):

app/editor/src/features/admin/tags/TagListFilter.tsx
jdtoombs: Lol was thinking about that...

PR #233, opened 2022-08-25
TNO-377: CHES Integration

jdtoombs:
Putting this up as draft to get some eyes on existing code. Working through some issues now with the notification queue.

### DIFF
--- a/libs/net/ches/ChesService.cs
+++ b/libs/net/ches/ChesService.cs
@@ -66,6 +66,82 @@ namespace TNO.Ches
         }
 
         /// <summary>
+        /// Send a request to the specified endpoint.
+        /// </summary>
+        /// <typeparam name="TR"></typeparam>
+        /// <param name="endpoint"></param>
+        /// <param name="method"></param>
+        /// <returns></returns>
+        private async Task<string> SendAsync(string endpoint, HttpMethod method)
+        {
+            var url = GenerateUrl(endpoint);
+
+            var headers = new HttpRequestMessage().Headers;
+            try
+            {
+                var response = await this.Client.SendAsync(url, method, headers);
+                return await response.Content.ReadAsStringAsync();
+            }
+            catch (HttpClientRequestException ex)
+            {
+                _logger.LogError(ex, $"Failed to send/receive request: {ex.StatusCode} {url}");
+                var response = await this.Client?.DeserializeAsync<Ches.Models.ErrorResponseModel>(ex.Response);
+                throw new ChesException(ex, this.Client, response);
+            }
+        }
+
+        /// <summary>
+        /// Send a request to the specified endpoint.
+        /// </summary>
+        /// <typeparam name="TR"></typeparam>
+        /// <param name="endpoint"></param>
+        /// <param name="method"></param>
+        /// <returns></returns>
+        private async Task<TR> SendAsync<TR>(string endpoint, HttpMethod method)
+        {
+            var url = GenerateUrl(endpoint);
+
+            var headers = new HttpRequestMessage().Headers;
+            try
+            {
+                return await this.Client.SendAsync<TR>(url, method, headers);
+            }
+            catch (HttpClientRequestException ex)
+            {
+                _logger.LogError(ex, $"Failed to send/receive request: {ex.StatusCode} {url}");
+                var response = await this.Client?.DeserializeAsync<Ches.Models.ErrorResponseModel>(ex.Response);
+                throw new ChesException(ex, this.Client, response);
+            }
+        }
+
+        /// <summary>
+        /// Send a request to the specified endpoint.
+        /// Make a request to get an access token if required.
+        /// </summary>
+        /// <typeparam name="TR"></typeparam>
+        /// <typeparam name="TD"></typeparam>
+        /// <param name="endpoint"></param>
+        /// <param name="method"></param>
+        /// <param name="data"></param>
+        /// <returns></returns>
+        private async Task<TR> SendAsync<TR, TD>(string endpoint, HttpMethod method, TD data)
+            where TD : class
+        {
+            var url = GenerateUrl(endpoint);
+
+            var headers = new HttpRequestMessage().Headers;
+            try
+            {
+                return await this.Client.SendJsonAsync<TR, TD>(url, method, headers, data);
+            }
+            catch (HttpClientRequestException ex)
+            {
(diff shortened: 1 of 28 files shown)


libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+    {
+        #region Variables
+        private readonly ClaimsPrincipal _user;
+        private TokenModel _token = null;
+        private readonly JwtSecurityTokenHandler _tokenHandler;
+        private readonly ILogger<IChesService> _logger;
+        #endregion
+
+        #region Properties
+        protected IHttpRequestClient Client { get; }
```
Fosol, line 34: You should use the `OpenIdConnectRequestClient` instead of this `HttpRequestClient`.  It has a bunch of the functionality you've duplicated in this (i.e. RequestAccessToken).

Fosol (changes requested):
Looks familiar :D  Just update the service to use the OpenIdConnectRequestClient so that we don't duplicate functionality in different libraries.

libs/net/ches/ChesService.cs
jdtoombs: Let me know if I did that right, not 100% sure!

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+        {
+            return $"{this.Options.HostUri}{endpoint}";
+        }
+
+        /// <summary>
+        /// Ensure we have an active access token.
+        /// Make an HTTP request if one is needed.
+        /// </summary>
+        /// <returns></returns>
+        private async Task RefreshAccessTokenAsync()
```
Fosol, line 74: You shouldn't need this as the `IOpenIdConnectRequestClient` has this ability

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+    {
+        #region Variables
+        private readonly ClaimsPrincipal _user;
+        private TokenModel _token = null;
+        private readonly JwtSecurityTokenHandler _tokenHandler;
+        private readonly ILogger<IChesService> _logger;
+        #endregion
+
+        #region Properties
+        protected OpenIdConnectRequestClient Client { get; }
```
Fosol, line 34: Probably should use the interface not the implementation here.

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+        /// <summary>
+        /// Send a request to the specified endpoint.
+        /// </summary>
+        /// <typeparam name="TR"></typeparam>
+        /// <param name="endpoint"></param>
+        /// <param name="method"></param>
+        /// <returns></returns>
+        private async Task<string> SendAsync(string endpoint, HttpMethod method)
+        {
+            await RefreshAccessTokenAsync();
```
Fosol, line 92: Shouldn't be required because `IOpenIdConnectRequestClient` should do this automatically when making requests.

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+namespace TNO.Ches
+{
+    /// <summary>
+    /// ChesService class, provides a service for integration with Ches API services.
+    /// </summary>
+    public class ChesService : IChesService
+    {
+        #region Variables
+        private readonly ClaimsPrincipal _user;
+        private TokenModel _token = null;
```
Fosol, line 28: I don't think you need to manage the token here.  I think it's covered in the `IOpenIdConnectRequestClient' class

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+        /// <summary>
+        /// Send a request to the specified endpoint.
+        /// </summary>
+        /// <typeparam name="TR"></typeparam>
+        /// <param name="endpoint"></param>
+        /// <param name="method"></param>
+        /// <returns></returns>
+        private async Task<TR> SendAsync<TR>(string endpoint, HttpMethod method)
+        {
+            await RefreshAccessTokenAsync();
```
Fosol, line 121: Shouldn't be required because IOpenIdConnectRequestClient should do this automatically when making requests.

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+        /// <typeparam name="TR"></typeparam>
+        /// <typeparam name="TD"></typeparam>
+        /// <param name="endpoint"></param>
+        /// <param name="method"></param>
+        /// <param name="data"></param>
+        /// <returns></returns>
+        private async Task<TR> SendAsync<TR, TD>(string endpoint, HttpMethod method, TD data)
+            where TD : class
+        {
+            await RefreshAccessTokenAsync();
```
Fosol, line 153: Shouldn't be required because IOpenIdConnectRequestClient should do this automatically when making requests.

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,369 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Make an HTTP request to CHES to get an access token for the specified 'username' and 'password'.
+        /// </summary>
+        /// <param name="username"></param>
+        /// <param name="password"></param>
+        /// <returns></returns>
+        public async Task<TokenModel> GetTokenAsync(string username = null, string password = null)
```
Fosol, line 178: Shouldn't be required because IOpenIdConnectRequestClient should do this automatically when making requests.

Fosol (changes requested):

libs/net/ches/ChesService.cs
```diff
@@ -0,0 +1,310 @@
...
+namespace TNO.Ches
+{
+    /// <summary>
+    /// ChesService class, provides a service for integration with Ches API services.
+    /// </summary>
+    public class ChesService : IChesService
+    {
+        #region Variables
+        private readonly ClaimsPrincipal _user;
+        private readonly JwtSecurityTokenHandler _tokenHandler;
```
Fosol, line 28: Does not appear to be required.  Linting should have picked this up and said unused variable.

Fosol (changes requested):
One more minor change.

libs/net/ches/ChesService.cs
jdtoombs: Ooops thought I got rid of that

Fosol (approved):

PR #237, opened 2022-08-29
TNO-529 Fix add data source

Fosol:
When adding a new data source if a schedule was included it would fail.  This update resolves this issue.

### DIFF
--- a/libs/net/dal/Extensions/DataSourceExtensions.cs
+++ b/libs/net/dal/Extensions/DataSourceExtensions.cs
@@ -20,7 +20,7 @@ public static class DataSourceExtensions
     {
         updated.ActionsManyToMany.SetEntityState(context);
         updated.MetricsManyToMany.SetEntityState(context);
-        updated.SchedulesManyToMany.SetEntityState(context);
+        updated.Schedules.AddRange(updated.SchedulesManyToMany.Select(s => s.Schedule!).Where(s => s != null).ToArray());
         return updated;
     }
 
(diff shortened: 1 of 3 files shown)


libs/net/dal/Extensions/DataSourceExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static DataSource AddToContext(this DataSource updated, TNOContext contex
     {
         updated.ActionsManyToMany.SetEntityState(context);
         updated.MetricsManyToMany.SetEntityState(context);
-        updated.SchedulesManyToMany.SetEntityState(context);
+        updated.Schedules.AddRange(updated.SchedulesManyToMany.Select(s => s.Schedule!).Where(s => s != null).ToArray());
```
Fosol, line 23: This resolves the issue.

libs/net/dal/Services/DataSourceService.cs
```diff
@@ -8,32 +8,62 @@
...
             .Include(ds => ds.MediaType)
             .Include(ds => ds.License)
             .Where(ds => ds.ContentTypeId != 0)
-            .OrderBy(ds => ds.Code).ThenBy(ds => ds.Name).ToArray();
+            .OrderBy(ds => ds.Code)
+            .ThenBy(ds => ds.Name)
+            .ToArray()
+            .Select(ds =>
+            {
+                ds.Connection = "{}";
```
Fosol, line 52: Remove sensitive info from common endpoints.

PR #241, opened 2022-08-31
TNO-525 Fix AV editing attach

Fosol:
Fixed the issue when attaching an existing clip to a new snippet.  Also resolve an issue when opening the Clips tab on content from a data source that doesn't have an associated folder (i.e. `/clip/DATA`).

The solution isn't elegant, but it follows the same pattern for uploading files.  Essentially, after content has been successfully added, it will make a subsequent request to attach the file.  This isn't ideal for a number of reasons, but is the simplest implementation.  Two requests is always slower than one.  If the first or second fails it results in bad data.

Added new API endpoint to check if a folder exists `/api/storage/exists?path`.

image


app/editor/src/features/content/form/ContentClipForm.tsx
```diff
@@ -59,14 +74,27 @@ export const ContentClipForm: React.FC<IContentClipFormProps> = ({
...
   };
 
   const onAttach = async (item: IItemModel) => {
-    await storage.attach(content.id, `${folder.path}/${item.name}`).then((data) => {
-      setContent(toForm(data));
-      toast.success('Attachment added to this snippet.');
-    });
+    if (values.id === 0) {
+      // Add a reference to the file so that it can be copied to the API when the content is saved.
+      setFieldValue('fileReferences', [
```
Fosol, line 83: Here's the magic to attach the file to the content.  However, this isn't the complete picture because this file will not actually be accessible to the API endpoints until the file is copied over.

app/editor/src/features/content/form/ContentClipForm.tsx
```diff
@@ -121,10 +149,9 @@ export const ContentClipForm: React.FC<IContentClipFormProps> = ({
   };
 
   const navigateUp = () => {
-    var pathComps = folder.path.split('/');
-    if (pathComps.length === 4) {
-      setPath('clip/' + pathComps[2]);
-    }
+    var path = folder.path.split('/').filter((v) => !!v);
```
Fosol, line 152: Fixes the basic navigation to work.

app/editor/src/features/content/form/ContentClipForm.tsx
```diff
@@ -231,15 +258,24 @@ export const ContentClipForm: React.FC<IContentClipFormProps> = ({
...
             confirmText="Yes, Remove It"
             onConfirm={async () => {
-              await storage.delete(`${folder.path}/${item?.name}`);
-              onSelect();
-              setFolder({ ...folder, items: folder.items.filter((i) => i.name !== item?.name) });
-              toast.success(`${item?.name} has been deleted`);
-              toggle();
+              try {
+                // TODO: Only certain users should be allowed to delete certain files/folders.
+                await storageApi.delete(
```
Fosol, line 267: We can now delete folders.

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -96,10 +94,21 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType = Content
...
+        const result = await upload(contentResult, values.file);
+        setContent(toForm({ ...result, tonePools: values.tonePools }));
+      } else if (
+        !originalId &&
+        !!values.fileReferences.length &&
+        !values.fileReferences[0].isUploaded
+      ) {
+        // TODO: Make it possible to upload on the initial save instead of a separate request.
+        // A file was attached but hasn't been uploaded to the API.
+        const fileReference = values.fileReferences[0];
```
Fosol, line 109: This resolves the issue where the attached clip will be copied over to the API properly.  It's not a great solution.

app/editor/src/features/content/form/constants/clipDirectoryColumns.tsx
```diff
@@ -90,7 +94,7 @@ export const clipDirectoryColumns = (
             className="delete fa-lg"
             title="Delete"
             onClick={() => {
-              onDelete(row.values);
+              onDelete(row.original);
```
Fosol, line 97: Must pass `original` if we want properties that are not in the table.

app/editor/src/features/content/form/constants/clipDirectoryColumns.tsx
```diff
@@ -62,26 +62,30 @@ export const clipDirectoryColumns = (
     Header: () => <div>Actions</div>,
     accessor: 'isDirectory',
     maxWidth: 30,
-    Cell: ({ row, data }: any) => (
-      <Row className={`file-actions ${row.values.isDirectory ? 'hidden' : 'center'}`} wrap="nowrap">
+    Cell: ({ row }) => (
+      <Row className={`file-actions ${row.original.isDirectory && 'directory'}`} wrap="nowrap">
```
Fosol, line 66: Displays delete for folders now.

app/editor/src/features/content/form/ContentSummaryForm.tsx
```diff
@@ -242,7 +242,7 @@ export const ContentSummaryForm: React.FC<IContentSummaryFormProps> = ({
             variant={ButtonVariant.danger}
             className="top-spacer"
             onClick={() => {
-              const regex = /\[.*\]/; // TODO: This is far too eager and could remove value content.
+              const regex = /\[.*\]/; // TODO: This is far too eager and could remove valuable content.
```
jdtoombs, line 245: Wonder if we should add a story to capture this, I like the way GitHub does their tags for PRs

jdtoombs (approved):
Don't think it'll mess with my work too much :)

app/editor/src/features/content/form/ContentSummaryForm.tsx
Fosol: @jdtoombs Yes create a story for this.

PR #247, opened 2022-09-02
TNO-451 Fix kafka consumer

Fosol:
Making it easier to do .NET development without using a devcontainer.  You will need to install the SDK locally, but it may work better for you in some scenarios.

Fixing a bug with Kafka consumers that attempt to listen to a topic that doesn't exist.  Essentially the consumer would go into basic hibernation mode until a new list of topics was provided.  I think it's better that the service stops running so it's not perceived as okay.


.code-workspace
```diff
@@ -0,0 +1,16 @@
+{
```
Fosol, line 1: This is used if you want to open the project as a workspace.

.vscode/launch.json
```diff
@@ -34,6 +34,23 @@
       "stopAtEntry": false,
       "envFile": "${workspaceFolder}/services/net/indexing/.env"
     },
+    {
+      // Use IntelliSense to find out which attributes exist for C# debugging
+      // Use hover for the description of the existing attributes
+      // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
+      "name": "Run Transcription Service",
```
Fosol, line 41: Run Transcription service from the root

TNO.sln
```diff
@@ -0,0 +1,194 @@
+﻿
```
Fosol, line 1: Build all dotnet projects from root

libs/net/TNO.sln
```diff
@@ -27,6 +27,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TNO.Test.Services", "tests\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TNO.Kafka", "kafka\TNO.Kafka.csproj", "{7CAAB3D8-6135-4398-B5DC-8F2AEA28843D}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TNO.Ches", "ches\TNO.Ches.csproj", "{08A82D74-0854-498F-9C74-E0A7242FE430}"
```
Fosol, line 30: Adding missing project so that it can be built from devcontainer root

openshift/kustomize/app/editor/build/base/build.yaml
```diff
@@ -53,7 +53,7 @@ spec:
   resources:
     requests:
       cpu: 20m
-      memory: 250Mi
+      memory: 500Mi
     limits:
       cpu: 500m
-      memory: 2Gi
+      memory: 3Gi
```
Fosol, line 59: app was running out of memory when building.

services/net/content/ContentManager.cs
```diff
@@ -167,6 +167,9 @@ private bool ConsumerErrorHandler(object sender, ErrorEventArgs e)
         this.State.RecordFailure();
         if (e.GetException() is ConsumeException ex)
         {
+            // Need to tell Kafka that this means it can't continue.
+            if (ex.Message == "Broker: Unknown topic or partition")
```
Fosol, line 171: Force a failure that will stop the consuming thread.  I'm not 100% this is the behaviour we want, but I also don't want the service pretending it's running.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -166,6 +166,9 @@ private async Task HandleMessageAsync(ConsumeResult<string, TranscriptRequest> r
             // Identify requests for transcription for content that does not exist.
             this.Logger.LogWarning("Content does not exist for this message. Key: {Key}, Content ID: {ContentId}", result.Message.Key, result.Message.Value.ContentId);
         }
+
+        // Successful run clears any errors.
+        this.State.ResetFailures();
```
Fosol, line 171: Only reset after successfully consuming a message.

PR #248, opened 2022-09-02
TNO-520: Breadcrumb trail for clips

jdtoombs:
Having some bugs with the query params after latest changes, removed for now as breadcrumb functionality works. Query params will update but do not control state currently. Will create ticket to capture this.

Also ended up just making our own component, seemed to fit better for our scenario. 

breadcrumb

### DIFF
--- a/app/editor/src/features/content/form/Breadcrumb.tsx
+++ b/app/editor/src/features/content/form/Breadcrumb.tsx
@@ -13,7 +13,7 @@ export interface IBreadcrumbProps {
  *  search params to the current directory
  */
 export const Breadcrumb: React.FC<IBreadcrumbProps> = ({ path, setPath }) => {
-  const [searchParams, setSearchParams] = useSearchParams();
+  const [, setSearchParams] = useSearchParams();
   /** split path into array in order to display breadcrumb as individual items */
   let splitPath = path.split('/').filter((s) => !!s);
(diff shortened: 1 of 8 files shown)


app/editor/src/features/content/form/Breadcrumb.tsx
```diff
@@ -0,0 +1,39 @@
...
+  /** the current path being viewed in the clip explorer */
+  path: string;
+  /** set path of the parent component */
+  setPath: (path: string) => void;
+}
+
+/** component used to navigate state controlled path directory, also updates the
+ *  search params to the current directory
+ */
+export const Breadcrumb: React.FC<IBreadcrumbProps> = ({ path, setPath }) => {
```
Fosol, line 15: Should this be a component not a feature?  `./components/breadcrumb`?  
Remember `features` are for business functionality while `components` are reusable tools that we could put in multiple different projects.

Fosol (changes requested):

app/editor/src/features/content/form/Breadcrumb.tsx
jdtoombs: I was actually going to ask if it is general enough to move to tno-core? Guessing maybe it would need some work before that though

Fosol (approved):

PR #258, opened 2022-09-09
TNO-531. 530

jdtoombs:
Quick little fixes

### DIFF
--- a/libs/net/dal/Migrations/Initial/Up/PostUp/01-DataSource.sql
+++ b/libs/net/dal/Migrations/Initial/Up/PostUp/01-DataSource.sql
@@ -2932,7 +2932,7 @@ INSERT INTO public.data_source (
   , CBC_CAPTURE_CODE
   , ''
   , true -- is_enabled
-  , 3 -- content_type_id
+  , 0 -- content_type_id
   , 3 -- media_type_id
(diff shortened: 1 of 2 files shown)


libs/net/dal/Migrations/Initial/Up/PostUp/01-DataSource.sql
```diff
@@ -2932,7 +2932,7 @@ INSERT INTO public.data_source (
   , CBC_CAPTURE_CODE
   , ''
   , true -- is_enabled
-  , NULL -- content_type_id
+  , 3 -- content_type_id
```
Fosol, line 2935: This is incorrect.  CBC capture does not generate any content, which is `0`.  It does highlight that I should probably update the DB entity so that it's not allowed to be `null` though.  As all data sources that don't generate content should be set to `0`.

Fosol (changes requested):

libs/net/dal/Migrations/Initial/Up/PostUp/01-DataSource.sql
jdtoombs: fixed!

Fosol (approved):

PR #259, opened 2022-09-09
TNO-428: Print content

jdtoombs:
Splitting off the frontend changes for PrintContent as I am battling some bugs with the backend currently.


app/editor/src/features/content/form/interfaces/IContentForm.tsx
```diff
@@ -22,7 +22,7 @@ export interface IContentForm {
   ownerId: number | '';
   contentTypeId: number;
   mediaTypeId: number;
-  licenseId: number;
+  licenseId?: number;
```
Fosol, line 25: The DB requires a license, so it's odd to allow it to be nullable on the form.  Is this to force the user to select a value?

app/editor/src/features/content/form/utils/toModel.ts
```diff
@@ -20,6 +20,7 @@ export function toModel(values: IContentForm): IContentModel {
     contentTypeId: values.contentTypeId,
     mediaTypeId: values.mediaTypeId,
     licenseId: values.licenseId,
+    // licenseId: values.contentTypeId === ContentType.Snippet ? values.licenseId : undefined,
```
Fosol, line 23: I don't think this is correct.

app/editor/src/features/content/list-view/ContentListView.tsx
```diff
@@ -98,6 +99,7 @@ export const ContentListView: React.FC = () => {
 
   const handleRowClick = (content: IContentModel) => {
     setActiveId(content.id);
+    setContentType(content.contentTypeId);
```
Fosol, line 102: How does this work when creating new content?

app/editor/src/hooks/api-editor/interfaces/IContentModel.ts
```diff
@@ -29,7 +29,7 @@ export interface IContentModel extends IAuditColumnsModel {
   contentType?: IContentTypeModel;
   mediaTypeId: number;
   mediaType?: IMediaTypeModel;
-  licenseId: number;
+  licenseId?: number;
```
Fosol, line 32: This is definitely incorrect, this interface represents data from the API.  It will never be `null`, this is a required value.

Fosol (changes requested):

app/editor/src/features/content/list-view/ContentListView.tsx
jdtoombs: Sets to snippet when Create Snippet is clicked, and to Print when the corresponding button clicked!

app/editor/src/features/content/form/utils/toModel.ts
jdtoombs: removing

Fosol (approved):

PR #260, opened 2022-09-09
Fixing transcription service

Fosol:
Fixing issue in Transcription Service that would ignore errors that occurred while performing transcriptions.  They will now be logged and recorded.

### DIFF
--- a/services/net/transcription/TranscriptionManager.cs
+++ b/services/net/transcription/TranscriptionManager.cs
@@ -152,6 +152,7 @@ public class TranscriptionManager : ServiceManager<TranscriptionOptions>
         var content = await _api.FindContentByIdAsync(result.Message.Value.ContentId);
         if (content != null)
         {
+            // TODO: Handle multi-threading so that more than one transcription can be performed at a time.
             await UpdateTranscriptionAsync(content);
         }
         else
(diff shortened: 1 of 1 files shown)


services/net/transcription/TranscriptionManager.cs
```diff
@@ -152,6 +152,7 @@ private async Task HandleMessageAsync(ConsumeResult<string, TranscriptRequest> r
         var content = await _api.FindContentByIdAsync(result.Message.Value.ContentId);
         if (content != null)
         {
+            // TODO: Handle multi-threading so that more than one transcription can be performed at a time.
```
Fosol, line 155: Currently the design of this service results in only a single transcription being processed.  Generally speaking it's a 1:1 process, for every minute of speech it takes a minute to process.  This means the Transcription Service locks up for a long time doing nothing.  However, it's not so easily solved either.  If we multi-thread we need to be able to limit the number, handle errors, and safely shutdown.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -239,11 +241,19 @@ private async Task<string> RequestTranscriptionAsync(byte[] data, string languag
             if (result.Reason == ResultReason.RecognizedSpeech)
             {
                 sb.Append(result.Text);
+                this.Logger.LogDebug("Speech transcription process \"{text}...\"", result.Text?[0..Math.Min(result.Text.Length, 25)]);
             }
+            // TODO: Handle other reasons.
```
Fosol, line 246: There are some other interesting reasons like `Intent` that could be valuable to capture.

PR #261, opened 2022-09-09
Fix data source bug

Fosol:
I introduced a bug when I removed data source connection information from the service endpoints.  Our ingest services require the connection information to determine if they should run.  It's not ideal including the connection information in the generic endpoints as it can contain secrets.  To secure these endpoints we'll need to add claim based authorization in another PR.

I've also added the Capture and Clip Services to the root launch file so we can run them outside of a devcontainer.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -38,6 +38,40 @@
       // Use IntelliSense to find out which attributes exist for C# debugging
       // Use hover for the description of the existing attributes
       // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
+      "name": "Run Capture Service",
+      "type": "coreclr",
+      "request": "launch",
+      "preLaunchTask": "build-capture",
+      // If you have changed target frameworks, make sure to update the program path.
+      "program": "${workspaceFolder}/services/net/capture/bin/Debug/net6.0/TNO.Services.Capture.dll",
+      "args": [],
+      "cwd": "${workspaceFolder}/services/net/capture",
+      // For more information about the 'console' field, see https://aka.ms/VSCode-CS-LaunchJson-Console
+      "console": "internalConsole",
+      "stopAtEntry": false,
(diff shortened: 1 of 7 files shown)


.vscode/launch.json
```diff
@@ -34,6 +34,40 @@
       "console": "internalConsole",
       "stopAtEntry": false
     },
+    {
```
Fosol, line 37: We can now run Capture and Clip Services from root outside of devcontainer.

libs/net/dal/Services/DataSourceService.cs
```diff
@@ -112,13 +113,14 @@ public IPaged<DataSource> Find(DataSourceFilter filter)
...
     /// Find and return all data sources that match the specified 'mediaTypeName'.
-    /// Removes connection information with possible secrets.
+    /// Removes connection information with possible secrets by default.
     /// </summary>
-    /// <param name="mediaTypeName"></param>
+    /// <param name="mediaTypeName">Name of media type</param>
+    /// <param name="includeConnection">Whether or not to include connection information</param>
     /// <returns></returns>
-    public IEnumerable<DataSource> FindByMediaType(string mediaTypeName)
+    public IEnumerable<DataSource> FindByMediaType(string mediaTypeName, bool includeConnection)
```
Fosol, line 121: Include connection information if required.

libs/net/kafka/KafkaListener`.cs
```diff
@@ -141,6 +141,8 @@ public async Task ConsumeAsync(Func<ConsumeResult<TKey, TValue>, Task> action, C
             if (!this.IsReady) throw new InvalidOperationException("Consumer is not ready for consuming, it must be opened first.");
 
             this.IsConsuming = true;
+            _logger.LogDebug("Waiting to receive message from Kafka topics:'{topic}'", String.Join(", ", this.Consumer.Subscription));
```
Fosol, line 144: Helps to debug to know if the service is waiting for a message, otherwise it's not clear.

PR #264, opened 2022-09-13
TNO-597 Update DB migration

Fosol:
To resolve a few naming inconsistencies the Initial DB migration has been updated.

> This PR will require a `make db-refresh`.  Don't forget to backup and restore any data you need to keep.  I will refresh the DEV database.

- `DataSource.ContentTypeId` is now not nullable.  It is a foreign key to ContentType.
- `Content.Transcription` has been renamed to `Content.Body` to reflect that it is the actual story text.
- `ContentType` now has a record for `1:None`.  This provides a way to set the data source as one that does not generate content.
- `DataSource.Connection` has been updated to not include `import`.  By default it will be true if the `ContentTypeId > 1`
- `DataSource.Topic` will not default to the `DataSource.Code` value.  This may be duplicate and might be removed eventually, but the idea is to provide an override, or a way to send messages to Kafka in a reliable way but also not confuse the administrator.

Kafka Consumer services had a bug where a failed HTTP request would result in a dead thread.  I've added a `OnStop` event so hopefully get it restarted.  At some point I will look into a way to possible better handle failed HTTP requests as it will result in lost content.


libs/net/dal/Migrations/Initial/Up/PostUp/00-ContentType.sql
```diff
@@ -5,14 +5,25 @@ BEGIN
...
   "name"
   , "description"
+  , "sort_order"
   , "is_enabled"
   , "created_by_id"
   , "created_by"
   , "updated_by_id"
   , "updated_by"
 ) VALUES (
+  'None'
```
Fosol, line 15: New row

libs/net/dal/Migrations/Initial/Up/PostUp/01-DataSource.sql
```diff
@@ -27,11 +27,11 @@ INSERT INTO public.data_source (
   , 'DAILYHIVE'
   , ''
   , true -- is_enabled
-  , 3 -- content_type_id
+  , 4 -- content_type_id
```
Fosol, line 30: Bump all values up one as we have `1:None` now.

libs/net/entities/Content.cs
```diff
@@ -126,19 +127,19 @@ public class Content : AuditColumns
...
     [Column("summary")]
     public string Summary { get; set; } = "";
 
     /// <summary>
-    /// get/set -
+    /// get/set - Story body text.
     /// </summary>
-    [Column("transcription")]
-    public string Transcription { get; set; } = "";
+    [Column("body")]
```
Fosol, line 138: Renamed

libs/net/entities/DataSource.cs
```diff
@@ -79,7 +79,7 @@ public class DataSource : AuditColumns
     /// This controls the editor form UI.
     /// </summary>
     [Column("content_type_id")]
-    public int? ContentTypeId { get; set; }
+    public int ContentTypeId { get; set; }
```
Fosol, line 82: not nullable

libs/net/kafka/KafkaListener`.cs
```diff
@@ -200,6 +205,7 @@ public void Stop()
             this.IsConsuming = false;
             this.IsReady = false;
             this.Consumer.Close();
+            OnStop?.Invoke(this, new EventArgs());
```
Fosol, line 208: Inform the owner of this object it has stopped.  This way the owner can cancel the thread.

openshift/kustomize/api/build/base/build.yaml
```diff
@@ -31,7 +31,7 @@ metadata:
     created-by: jeremy.foster
     branch: dev
 spec:
-  completionDeadlineSeconds: 1800
+  completionDeadlineSeconds: 2700
```
Fosol, line 34: Increase defaults to reduce build failures

openshift/kustomize/app/editor/build/base/build.yaml
```diff
@@ -31,7 +31,7 @@ metadata:
     created-by: jeremy.foster
     branch: dev
 spec:
-  completionDeadlineSeconds: 1800
+  completionDeadlineSeconds: 2700
```
Fosol, line 34: Increase defaults to reduce build failures

openshift/templates/tekton/pipelines/complete.yaml
```diff
@@ -118,18 +118,6 @@ objects:
             mounts /zap/wrk to store generated configs and results.
 
       tasks:
```
Fosol, line 120: Reorder pipeline so that maintenance only gets turned on if all the builds are successful.  This will reduce downtime.

services/net/content/ContentManager.cs
```diff
@@ -97,11 +107,13 @@ public override async Task RunAsync()
...
                         .Where(ds => ds.IsEnabled &&
                             ds.ContentTypeId > 0 &&
-                            !String.IsNullOrWhiteSpace(ds.Topic) &&
-                            ds.Connection.ContainsKey("import") &&
-                            ((JsonElement)ds.Connection["import"]).GetBoolean())
+                            !String.IsNullOrWhiteSpace(ds.Topic))
                         .Select(ds => ds.Topic).ToArray());
 
+                    // Only include topics that exist.
+                    var kafkaTopics = this.KafkaAdmin.ListTopics();
```
Fosol, line 114: Now that we assume any data source with a `DataSource.ContentTypeId > 1` is importing, we need to remove all the non-existing topics.

jdtoombs (approved):

PR #266, opened 2022-09-13
TNO-581, 602

jdtoombs:
- Fix previous and next content navigation buttons
- Default form to Snippet while the other content forms are undeveloped 
- These fixes also address the popout button not expanding to the correct page


jdtoombs:
@Fosol these were easy so just checked off the list quickly, could you drag them into the sprint for me :) I don't have permission on JIRA

Fosol (approved):

PR #267, opened 2022-09-13
Fix Kafka Consumer Services

Fosol:
Fixing issue with Kafka consumer services that stop consuming messages after an unexpected failure.  Normally that failure occurs within an HTTP request to the API.  Also added additional logging information for HTTP errors.

Updated the pipeline to build one thing at a time.  When attempting to do them all at once we run out of resources and the pipeline gets stuck.  Sequential however results in build times of over 30 minutes...

- Content Service
- Transcription Service
- Indexing Service
- NLP Service

### DIFF
--- a/libs/net/core/Exceptions/HttpClientRequestException.cs
+++ b/libs/net/core/Exceptions/HttpClientRequestException.cs
@@ -46,7 +46,7 @@ namespace TNO.Core.Exceptions
         /// </summary>
         /// <param name="response"></param>
         /// <returns></returns>
-        public HttpClientRequestException(HttpResponseMessage response) : base($"HTTP Request '{response?.RequestMessage?.RequestUri}' failed", null, response?.StatusCode)
+        public HttpClientRequestException(HttpResponseMessage response) : base($"HTTP Request {response?.StatusCode}:'{response?.RequestMessage?.RequestUri}' failed", null, response?.StatusCode)
         {
             this.Response = response ?? throw new ArgumentNullException(nameof(response)); // NOSONAR
             var body = response.Content.ReadAsStringAsync().Result;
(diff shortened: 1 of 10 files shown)


libs/net/core/Exceptions/HttpClientRequestException.cs
```diff
@@ -46,7 +46,7 @@ public HttpClientRequestException(string message, Exception innerException, Http
         /// </summary>
         /// <param name="response"></param>
         /// <returns></returns>
-        public HttpClientRequestException(HttpResponseMessage response) : base($"HTTP Request '{response?.RequestMessage?.RequestUri}' failed", null, response?.StatusCode)
+        public HttpClientRequestException(HttpResponseMessage response) : base($"HTTP Request {response?.StatusCode}:'{response?.RequestMessage?.RequestUri}' failed", null, response?.StatusCode)
```
Fosol, line 49: Include the status code in the error.

libs/net/kafka/KafkaListener`.cs
```diff
@@ -162,13 +163,23 @@ public async Task ConsumeAsync(Func<ConsumeResult<TKey, TValue>, Task> action, C
         catch (ConsumeException ex)
         {
             // https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#fatal-consumer-errors
-            _logger.LogError(ex, "Consumer error: {Message}", ex.Message);
+            // TODO: How to tell Kafka this message was not successful so that it can be picked up again.
+            _logger.LogError(ex, "Consumer error: {message}", ex.Message);
             this.IsReady = (!OnError?.Invoke(this, new ErrorEventArgs(ex)) ?? true) && this.IsReady;
         }
+        catch (HttpClientRequestException ex)
```
Fosol, line 170: Capture these errors separately for a different log message.

openshift/templates/tekton/pipelines/complete.yaml
```diff
@@ -141,6 +141,8 @@ objects:
               value: $(params.IMAGE_TAG)
 
         - name: build-api
+          runAfter:
```
Fosol, line 144: Sequential build

services/net/content/ContentManager.cs
```diff
@@ -114,26 +108,15 @@ public override async Task RunAsync()
                     var kafkaTopics = this.KafkaAdmin.ListTopics();
                     topics = topics.Except(topics.Except(kafkaTopics)).ToArray();
 
-                    // If the topics change we need to subscribe to the new topics.
-                    if (!ContainsAll(this.Topics, topics))
```
Fosol, line 118: No need for this as it's performed by the `this.Consumer.Subscribe(topics)` function.

jdtoombs (approved):

PR #271, opened 2022-09-15
TNO-427: Tooltip not disappearing on dropdown

jdtoombs:
I had tried a handful of different libraries with the same issue, as well as trying to find a better solution for "globally" handled tooltips. Was not able to make progress there.

This is a bit of a hacky fix; however, allows us to keep our tooltips the way they are throughout the application. I believe we can also get rid of the rebuild statements if we restructure where the tooltip is being rendered; but that may be outside of this ticket.

tooltip

### DIFF
--- a/app/editor/src/components/form/select/Select.tsx
+++ b/app/editor/src/components/form/select/Select.tsx
@@ -1,6 +1,7 @@
 import React, { Ref } from 'react';
 import { ActionMeta, GroupBase, Props } from 'react-select';
 import ReactSelect from 'react-select/dist/declarations/src/Select';
+import ReactTooltip from 'react-tooltip';
 import { FieldSize } from 'tno-core';
 import { Row } from 'tno-core/dist/components/flex';
 
@@ -51,6 +52,8 @@ export interface ISelectProps<OptionType> extends SelectProps {
   ref?: Ref<ReactSelect<OptionType, boolean, GroupBase<OptionType>>>;
 }
 
+type TooltipRef = { tooltipRef: null } | null;
+
 /**
  * Select component provides a bootstrapped styled button element.
  * @param param0 Select element attributes.
@@ -79,7 +82,7 @@ export const Select = <OptionType extends IOptionItem>({
 }: ISelectProps<OptionType>) => {
   const selectRef = React.useRef(null);
   const inputRef = React.useRef<HTMLInputElement>(null);
-
+  const tip = React.useRef(null);
   return (
     <styled.Select className="frm-in">
       {label && (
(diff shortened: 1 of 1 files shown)


app/editor/src/components/form/select/Select.tsx
```diff
@@ -51,6 +52,8 @@ export interface ISelectProps<OptionType> extends SelectProps {
   ref?: Ref<ReactSelect<OptionType, boolean, GroupBase<OptionType>>>;
 }
 
+type TooltipRef = { tooltipRef: null } | null;
```
jdtoombs, line 55: Is there a better place to put this? Maybe utils? Wasn't sure where; I believe we will not need to use a ref outside of the select component.

app/editor/src/components/form/select/Select.tsx
```diff
@@ -79,7 +82,7 @@ export const Select = <OptionType extends IOptionItem>({
 }: ISelectProps<OptionType>) => {
   const selectRef = React.useRef(null);
   const inputRef = React.useRef<HTMLInputElement>(null);
-
+  const tip = React.useRef(null);
```
jdtoombs, line 85: Probably should make a better name for this.. will adjust in a.m

app/editor/src/components/form/select/Select.tsx
```diff
@@ -93,8 +96,13 @@ export const Select = <OptionType extends IOptionItem>({
             onClear?.();
           }
         }}
-        data-for="main-tooltip"
+        data-for="select-tooltip"
         data-tip={tooltip}
+        onClick={() => {
```
jdtoombs, line 101: Ensures the tooltip is hidden when an option is selected

app/editor/src/components/form/select/Select.tsx
```diff
@@ -147,6 +155,7 @@ export const Select = <OptionType extends IOptionItem>({
         />
       )}
       {error && <p role="alert">{error}</p>}
+      <ReactTooltip ref={tip} id="select-tooltip" effect="float" type="light" place="top" />
```
jdtoombs, line 158: Select will not use the global tooltip so we can easily access the ref

app/editor/src/components/form/select/Select.tsx
jdtoombs: Also note that `onBlur` is not triggered after selecting until the user click the screen again (a second time)

Fosol (commented):
What happens when there are multiple select components?

jdtoombs:
> What happens when there are multiple select components?

@Fosol: Works as intended - can see in the GIF above with the two different tooltips on the dropdowns. If that is what you mean!

Fosol (approved):

PR #272, opened 2022-09-15
TNO-418: Change error validation

jdtoombs:
Updated the error validation on the clips "form", did not use Yup as this is not technically behaving like a Formik form but follows the same conditional error message display format.

Also created a small component used elsewhere in the application to display errors.

clip-errors

### DIFF
--- a/app/editor/src/components/form/error/Error.tsx
+++ b/app/editor/src/components/form/error/Error.tsx
@@ -1,0 +1,15 @@
+import { Show } from 'tno-core';
+
+export interface IErrorProps {
+  /** The error message that needs to be displayed */
+  error?: string;
+}
+
+/** Simple component to conditionally display an error message */
+export const Error: React.FC<IErrorProps> = ({ error }) => {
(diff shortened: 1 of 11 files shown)


app/editor/src/components/form/error/Error.tsx
```diff
@@ -0,0 +1,15 @@
+import { Show } from 'tno-core';
+
+export interface IErrorProps {
+  /** The error message that needs to be displayed */
+  error?: string;
+}
+
+/** Simple component to conditionally display an error message */
+export const Error: React.FC<IErrorProps> = ({ error }) => {
```
Fosol, line 9: You could extend this component and make one for Formik.  It would just need the `name` property to extract the appropriate error.

Fosol (approved):

app/editor/src/components/form/error/Error.tsx
jdtoombs: Good call

PR #282, opened 2022-09-23
Fix Tag Admin

Fosol:
The Tag admin forms were invalid.  They didn't correctly handle the `id` property.

image

### DIFF
--- a/app/editor/src/features/admin/tags/TagList.tsx
+++ b/app/editor/src/features/admin/tags/TagList.tsx
@@ -35,7 +35,7 @@ export const TagList: React.FC = () => {
           <IconButton
             iconType="plus"
             label={`Add new tag`}
-            onClick={() => navigate(`/admin/tags/NEW`)}
+            onClick={() => navigate(`/admin/tags/***`)}
           />
         </Row>
         <GridTable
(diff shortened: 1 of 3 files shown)


app/editor/src/features/admin/tags/TagList.tsx
```diff
@@ -35,7 +35,7 @@ export const TagList: React.FC = () => {
           <IconButton
             iconType="plus"
             label={`Add new tag`}
-            onClick={() => navigate(`/admin/tags/NEW`)}
+            onClick={() => navigate(`/admin/tags/***`)}
```
Fosol, line 38: Less chance of this being a valid tag identifier

app/editor/src/features/admin/tags/TagsForm.tsx
```diff
@@ -55,14 +56,21 @@ export const TagsForm: React.FC = () => {
...
       >
-        {({ isSubmitting, values }) => (
+        {({ isSubmitting, values, errors }) => (
           <div className="form-container">
             <Col className="form-inputs">
+              <FormikText
+                width={FieldSize.Large}
+                name="id"
+                label="Code"
+                disabled={id !== '***'}
```
Fosol, line 71: Cannot change a primary key value after it has been saved.

PR #293, opened 2022-09-28
TNO-445 Fix Kafka consumers

Fosol:
The Kafka consumer services were using a `Task.Factory.StartNew(...)`, but the tasks it created would not await and continued to run to completion (even though they were still running!).  The fix creates the thread in a simpler way and appears to correctly await until the process completes before attempting to create a new thread.

Also updated the Infrastructure as Code to resolve transcription service configuration issues in Openshift.

### DIFF
--- a/libs/net/kafka/KafkaListener`.cs
+++ b/libs/net/kafka/KafkaListener`.cs
@@ -128,6 +128,7 @@ public class KafkaListener<TKey, TValue> : IKafkaListener<TKey, TValue>, IDispos
     /// <summary>
     /// Listen for a message from Kafka.
     /// Place this in a loop if you want to receive more than one message.
+    /// Wrapped in a try+catch so that consuming failures should never result in a dead thread.
     /// </summary>
     /// <param name="action"></param>
     /// <param name="cancellationToken"></param>
@@ -150,12 +151,12 @@ public class KafkaListener<TKey, TValue> : IKafkaListener<TKey, TValue>, IDispos
             // This provides a way to keep trying to handle the current result because of some kind of intermittent failure.
             if (_currentResult == null)
             {
-                _logger.LogDebug("Waiting to receive message from Kafka topics:'{topic}'", String.Join(", ", this.Consumer?.Subscription ?? new List<string>()));
+                _logger.LogDebug("Waiting to receive message from Kafka topics: '{topic}'", String.Join(", ", this.Consumer?.Subscription ?? new List<string>()));
 
                 // I don't understand why Kafka didn't use an async+await pattern here, but this function blocks until it receives a message.
                 _currentResult = this.Consumer!.Consume(cancellationToken);
 
-                _logger.LogInformation("Message received from Kafka topic:'{topic}' key:'{key}'", _currentResult.Topic, _currentResult.Message.Key);
+                _logger.LogInformation("Message received from Kafka topic: '{topic}' key:'{key}'", _currentResult.Topic, _currentResult.Message.Key);
 
                 if (!this.IsPaused)
                 {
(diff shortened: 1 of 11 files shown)


libs/net/kafka/KafkaListener`.cs
```diff
@@ -164,18 +165,25 @@ public async Task ConsumeAsync(Func<ConsumeResult<TKey, TValue>, Task<ConsumerAc
...
                 }
             }
+            else
+            {
+                _logger.LogDebug("Retrying prior message from Kafka topic: '{topic}' key:'{key}'", _currentResult.Topic, _currentResult.Message.Key);
+            }
 
-            proceed = await action(_currentResult);
-            if (proceed == ConsumerAction.Proceed)
+            if (_currentResult != null)
```
Fosol, line 173: Somehow it was getting here with a null value.  I don't know how that is even possible and was unable to debug and find a way.

openshift/kustomize/services/transcription/base/deploy.yaml
```diff
@@ -37,9 +37,9 @@ spec:
         component: transcription-service
     spec:
       volumes:
-        - name: av-storage
+        - name: file-storage
```
Fosol, line 40: Transcription service needs to have access to the same volume as the API.

services/net/clip/ClipAction.cs
```diff
@@ -155,7 +155,7 @@ protected override async Task StopProcessAsync(ICommandProcess process, Cancella
     private ContentReferenceModel CreateContentReference(IngestModel ingest, ScheduleModel schedule)
     {
         var today = GetLocalDateTime(ingest, DateTime.UtcNow);
-        var publishedOn = new DateTime(today.Year, today.Month, today.Day, 0, 0, 0, DateTimeKind.Local) + schedule.StartAt;
+        var publishedOn = new DateTime(today.Year, today.Month, today.Day, 0, 0, 0, today.Kind) + schedule.StartAt;
```
Fosol, line 158: I believe this should result in the correct date displayed in the UI

services/net/content/ContentManager.cs
```diff
@@ -150,7 +150,7 @@ private void ConsumeMessages()
         if (_consumer == null || _notRunning.Contains(_consumer.Status))
         {
             _cancelToken = new CancellationTokenSource();
-            _consumer = Task.Run(ConsumerHandlerAsync, _cancelToken.Token);
```
Fosol, line 153: Can result in task running to completion before it's really done.

services/net/indexing/IndexingManager.cs
```diff
@@ -149,20 +149,20 @@ private void ConsumeMessages()
         if (_consumer == null || _notRunning.Contains(_consumer.Status))
         {
             _cancelToken = new CancellationTokenSource();
-            _consumer = Task.Factory.StartNew(() => ConsumerHandler(), _cancelToken.Token);
```
Fosol, line 152: Can result in task running to completion before it's really done.

services/net/nlp/NLPManager.cs
```diff
@@ -126,20 +126,20 @@ private void ConsumeMessages()
         if (_consumer == null || _notRunning.Contains(_consumer.Status))
         {
             _cancelToken = new CancellationTokenSource();
-            _consumer = Task.Factory.StartNew(() => ConsumerHandler(), _cancelToken.Token);
```
Fosol, line 129: Can result in task running to completion before it's really done.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -117,20 +117,20 @@ private void ConsumeMessages()
         if (_consumer == null || _notRunning.Contains(_consumer.Status))
         {
             _cancelToken = new CancellationTokenSource();
-            _consumer = Task.Factory.StartNew(() => ConsumerHandler(), _cancelToken.Token);
```
Fosol, line 120: Can result in task running to completion before it's really done.

jdtoombs (approved):

PR #294, opened 2022-09-28
TNO-621, 611

jdtoombs:
- Slight DOM structure change to `ContentListView` and min-height adjusted to `fit-content`
- Commentary will now only appear once (as indented) for content types other than snippet/print

### DIFF
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -96,11 +96,16 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType: initCont
   const enableNext = indexPosition < (page?.items.length ?? 0) - 1;
 
   const determineActions = () => {
-    if (contentType === ContentTypeName.Snippet)
-      return (a: IActionModel) => a.valueType === ValueType.Boolean;
-    if (contentType === ContentTypeName.PrintContent)
-      return (a: IActionModel) =>
-        a.valueType === ValueType.Boolean && a.name !== ActionName.NonQualified;
+    switch (contentType) {
+      case ContentTypeName.Snippet:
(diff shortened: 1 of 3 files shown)


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -96,11 +96,16 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType: initCont
...
-        a.valueType === ValueType.Boolean && a.name !== ActionName.NonQualified;
+    switch (contentType) {
+      case ContentTypeName.Snippet:
+        return (a: IActionModel) => a.valueType === ValueType.Boolean;
+      case ContentTypeName.PrintContent:
+        return (a: IActionModel) =>
+          a.valueType === ValueType.Boolean && a.name !== ActionName.NonQualified;
+      // TODO: Determine actions for remaining content types
+      default:
+        return (a: IActionModel) => a.valueType === ValueType.Boolean;
```
Fosol, line 107: Same as top, rewrite.
```
case ContentTypeName.Snippet:
default:
   return ...
```

Fosol (changes requested):

app/editor/src/features/content/form/ContentForm.tsx
jdtoombs: For some reason ES-Lint does not like the fall through

jdtoombs: Still works though

Fosol: Interesting linting rule to stop fall throughs... Makes me wonder why use a switch statement if you can't use a fall through?

Fosol (approved):

PR #296, opened 2022-09-28
Transcription video service

AlessiaYChen:


services/net/transcription/TranscriptionManager.cs
```diff
@@ -228,14 +228,21 @@ private async Task UpdateTranscriptionAsync(ContentModel content)
         // TODO: Handle different storage locations.
         // Remote storage locations may not be easily accessible by this service.
         var path = content.FileReferences.FirstOrDefault()?.Path;
-        var safePath = Path.Join(_options.VolumePath, path.MakeRelativePath());
+        var safePath = Path.Join(_options.FilePath, path.MakeRelativePath());
```
Fosol, line 231: This should be `_options.VolumePath`.  The configuration is to map the volume.  The path of the file is everything after the volume path.

Fosol: There should be no `_options.FilePath` in the source code.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -259,10 +266,15 @@ private async Task UpdateTranscriptionAsync(ContentModel content)
                 // The content is no longer available for some reason.
                 this.Logger.LogError("Content no longer exists. Content ID: {Id}", content.Id);
             }
+
         }
         else
         {
-            this.Logger.LogError("File does not exist for content. Content ID: {Id}, Path: {path}", content.Id, path);
+            if (mediaType == SourceMediaType.Video) {
+                this.Logger.LogError("Video File failed to converted to audio. Content ID: {Id}, Path: {path}", content.Id, safePath);
```
Fosol, line 274: At this point it's not a valid error message.  You should remove it.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -228,14 +228,21 @@ private async Task UpdateTranscriptionAsync(ContentModel content)
         // TODO: Handle different storage locations.
         // Remote storage locations may not be easily accessible by this service.
         var path = content.FileReferences.FirstOrDefault()?.Path;
-        var safePath = Path.Join(_options.VolumePath, path.MakeRelativePath());
+        var safePath = Path.Join(_options.FilePath, path.MakeRelativePath());
+
+        var mediaType = await IsVideoAsync(safePath) ? SourceMediaType.Video : SourceMediaType.Audio;
+        if (mediaType == SourceMediaType.Video)
```
Fosol, line 234: Move this within the `if (File.Exists(safePath)` statement.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -327,5 +339,49 @@ private async Task<string> RequestTranscriptionAsync(byte[] data, string languag
...
+        await process.WaitForExitAsync();
+        return !String.IsNullOrWhiteSpace(output);
+    }
+
+    /// <summary>
+    /// video to audio
+    /// </summary>
+    /// <param name="file">video file</param>
+    /// <returns></returns>
+    private async Task<bool> Video2Audio(string srcFile, string destFile)
```
Fosol, line 370: Because your logic above merely changes the extension to `mp3` I'd recommend not including the `destFile` parameter, and add the logic here.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -327,5 +339,49 @@ private async Task<string> RequestTranscriptionAsync(byte[] data, string languag
...
+        process.StartInfo.FileName = "/bin/sh";
+        process.StartInfo.Arguments = $"-c \"ffmpeg -i {srcFile} -y {destFile} \"";
+        process.StartInfo.UseShellExecute = false;
+        process.StartInfo.RedirectStandardOutput = true;
+        process.StartInfo.CreateNoWindow = true;
+        process.Start();
+
+        var output = await process.StandardOutput.ReadToEndAsync();
+        await process.WaitForExitAsync();
+        return !String.IsNullOrWhiteSpace(output);
```
Fosol, line 384: Not sure returning a boolean value makes a lot of sense here.  Return the full path to the new file.
If the process fails, you need to capture the error and propagating it.

Fosol (changes requested):

services/net/transcription/TranscriptionManager.cs
AlessiaYChen: I moved this session out of the **if (File.Exists(safePath)** since i overwrote the safe path after the convert. I'll do another if statement to check whether converted file exists or not if it make more sense to move it inside **if (File.Exists(safePath))**

Fosol: You only overwrite the `safePath` if (a) it's a video, and (b) the convert is successful.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -229,13 +229,22 @@ private async Task UpdateTranscriptionAsync(ContentModel content)
         // Remote storage locations may not be easily accessible by this service.
         var path = content.FileReferences.FirstOrDefault()?.Path;
         var safePath = Path.Join(_options.VolumePath, path.MakeRelativePath());
+
+        // convert to audio if it's video file
+        var isVideo = await IsVideoAsync(safePath);
```
Fosol, line 234: What happens if the file doesn't exist for this method `IsVideoAsync(...)`?

services/net/transcription/TranscriptionManager.cs
```diff
@@ -327,5 +337,50 @@ private async Task<string> RequestTranscriptionAsync(byte[] data, string languag
...
+        var process = new System.Diagnostics.Process();
+        process.StartInfo.Verb = $"Stream Type";
+        process.StartInfo.FileName = "/bin/sh";
+        process.StartInfo.Arguments = $"-c \"ffmpeg -i {file} 2>&1 | grep Video | awk '{{print $0}}' | tr -d ,\"";
+        process.StartInfo.UseShellExecute = false;
+        process.StartInfo.RedirectStandardOutput = true;
+        process.StartInfo.CreateNoWindow = true;
+        process.Start();
+
+        var output = await process.StandardOutput.ReadToEndAsync();
```
Fosol, line 358: Any errors will be lost, we need to log them.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -327,5 +337,50 @@ private async Task<string> RequestTranscriptionAsync(byte[] data, string languag
...
+        var process = new System.Diagnostics.Process();
+        process.StartInfo.Verb = $"Stream Type";
+        process.StartInfo.FileName = "/bin/sh";
+        process.StartInfo.Arguments = $"-c \"ffmpeg -i {srcFile} -y {destFile} 2>&1 \"";
+        process.StartInfo.UseShellExecute = false;
+        process.StartInfo.RedirectStandardOutput = true;
+        process.StartInfo.CreateNoWindow = true;
+        process.Start();
+
+        var output = await process.StandardOutput.ReadToEndAsync();
```
Fosol, line 380: Any errors will be lost, we need to log them.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -232,40 +232,50 @@ private async Task UpdateTranscriptionAsync(ContentModel content)
...
             this.Logger.LogInformation("Transcription requested.  Content ID: {Id}", content.Id);
 
-            var original = content.Body;
-            var fileBytes = File.ReadAllBytes(safePath);
-
-            var transcript = await RequestTranscriptionAsync(fileBytes); // TODO: Extract language from data source.
+            // convert to audio if it's video file
+            var isVideo = Path.GetExtension(safePath).ToLower() == ".mp4";
+            if (isVideo) {
+                var convertFilePath = await Video2Audio(safePath);
```
Fosol, line 243: You can simplify this to `safePath = await Video2Audio(safePath)` which reduces the need of an additional variable.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -232,40 +232,50 @@ private async Task UpdateTranscriptionAsync(ContentModel content)
         // Remote storage locations may not be easily accessible by this service.
         var path = content.FileReferences.FirstOrDefault()?.Path;
         var safePath = Path.Join(_options.VolumePath, path.MakeRelativePath());
+       
         if (File.Exists(safePath))
         {
             this.Logger.LogInformation("Transcription requested.  Content ID: {Id}", content.Id);
```
Fosol, line 238: This should be moved down to where a request has actually been made.  At this point it may still not be performed.

services/net/transcription/TranscriptionManager.cs
```diff
@@ -232,40 +232,50 @@ private async Task UpdateTranscriptionAsync(ContentModel content)
...
+            if (isVideo) {
+                var convertFilePath = await Video2Audio(safePath);
+                safePath = convertFilePath;
+            }
 
-            // Fetch content again because it may have been updated by an external source.
-            // This can introduce issues if the transcript has been edited as now it will overwrite what was changed.
-            var result = await _api.FindContentByIdAsync(content.Id);
-            if (result != null && !String.IsNullOrWhiteSpace(transcript))
+            if (!isVideo || !String.IsNullOrEmpty(safePath))
```
Fosol, line 247: If `Video2Audio` returns an empty string then we wouldn't perform this process.  Simplify to `!String.IsNullOrEmpty(safePath)`.

Fosol (changes requested):

Fosol (approved):

PR #297, opened 2022-09-29
Fix kafka consumers

Fosol:
The Kafka consumers were not correctly awaiting the completion of the handle message task.  This was originally resulting in the creation of new threads any time a long async process ran.  My last fix introduced a different bug, which was it no longer ran asynchronously.  

This fix appears to correct the prior issues.

### DIFF
--- a/services/net/content/ContentManager.cs
+++ b/services/net/content/ContentManager.cs
@@ -149,8 +149,11 @@ public class ContentManager : ServiceManager<ContentOptions>
     {
         if (_consumer == null || _notRunning.Contains(_consumer.Status))
         {
+            // Make sure the prior task is cancelled before creating a new one.
+            if (_cancelToken?.IsCancellationRequested == false)
+                _cancelToken?.Cancel();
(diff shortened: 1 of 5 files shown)


services/net/content/ContentManager.cs
```diff
@@ -149,8 +149,11 @@ private void ConsumeMessages()
     {
         if (_consumer == null || _notRunning.Contains(_consumer.Status))
         {
+            // Make sure the prior task is cancelled before creating a new one.
+            if (_cancelToken?.IsCancellationRequested == false)
+                _cancelToken?.Cancel();
             _cancelToken = new CancellationTokenSource();
-            _consumer = ConsumerHandlerAsync();
+            _consumer = Task.Run(ConsumerHandlerAsync, _cancelToken.Token);
```
Fosol, line 156: I originally switch away from `Task.Run(...)`, but I don't remember the reason.  `Factory.StartNew(...)` seemed to be a good choice, but it was also not awaiting a Task that performed async methods inside it.  So far my testing appears to show this working as intended.

PR #302, opened 2022-09-29
TNO-607, 457: Product store not updating correctly & Combined view not updating list 

jdtoombs:
**TNO-607**

**TNO-457 - List View Now Updates**

### DIFF
--- a/app/editor/src/hooks/useLookupSync.ts
+++ b/app/editor/src/hooks/useLookupSync.ts
@@ -6,7 +6,7 @@ import { useLookupStore } from 'store/slices';
 /**
  * Hook that syncs the lookups between the admin store and lookup store. Lookup store will need to be occasionally updated when an admin updates the admin store.
  */
-export const useLookupSync = () => {
+export const useLookupSync = (onSubmit?: Function) => {
   const adminProducts = useSelector((state: any) => state.admin.products);
   const lookupProducts = useSelector((state: any) => state.lookup.products);
 
@@ -16,5 +16,5 @@ export const useLookupSync = () => {
     // the following line indicates that the admin store has been updated and the lookups need to be synced
     if (!_.isEqual(adminProducts, lookupProducts)) store.storeProducts(adminProducts);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [store]);
+  }, [store, onSubmit]);
 };
--- a/app/editor/src/hooks/useLookupSync.ts
+++ b/app/editor/src/hooks/useLookupSync.ts
@@ -1,0 +1,20 @@
+import _ from 'lodash';
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { useLookupStore } from 'store/slices';
+
+/**
+ * Hook that syncs the lookups between the admin store and lookup store. Lookup store will need to be occasionally updated when an admin updates the admin store.
+ */
+export const useLookupSync = () => {
+  const adminProducts = useSelector((state: any) => state.admin.products);
+  const lookupProducts = useSelector((state: any) => state.lookup.products);
+
+  const [, store] = useLookupStore();
+
+  React.useEffect(() => {
+    // the following line indicates that the admin store has been updated and the lookups need to be synced
+    if (!_.isEqual(adminProducts, lookupProducts)) store.storeProducts(adminProducts);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [store]);
+};
--- a/app/editor/src/store/hooks/content/useContent.ts
+++ b/app/editor/src/store/hooks/content/useContent.ts
@@ -45,7 +45,17 @@ export const useContent = (props?: IContentProps): [IContentState, IContentContr
       return await dispatch('add-content', () => api.addContent(content));
     },
     updateContent: async (content: IContentModel) => {
-      return await dispatch('update-content', () => api.updateContent(content));
+      const result = await dispatch('update-content', () => api.updateContent(content));
+      if (state.content) {
+        actions.storeContent({
+          ...state.content,
+          items: state.content.items.map((i: IContentModel) => {
+            if (i.id === content.id) return content;
+            return i;
+          }),
+        });
+      }
+      return result;
     },
     deleteContent: async (content: IContentModel) => {
       return await dispatch('delete-content', () => api.deleteContent(content));
--- a/app/editor/src/store/hooks/content/useContent.ts
+++ b/app/editor/src/store/hooks/content/useContent.ts
@@ -42,7 +42,11 @@ export const useContent = (props?: IContentProps): [IContentState, IContentContr
       return result;
     },
     addContent: async (content: IContentModel) => {
-      return await dispatch('add-content', () => api.addContent(content));
+      const result = await dispatch('add-content', () => api.addContent(content));
+      if (state.content) {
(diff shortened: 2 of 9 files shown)


app/editor/src/store/hooks/admin/products/useProducts.ts
```diff
@@ -1,7 +1,7 @@
 import { IProductModel, useApiAdminProducts } from 'hooks/api-editor';
 import React from 'react';
 import { useApiDispatcher } from 'store/hooks';
-import { IAdminState, useAdminStore } from 'store/slices';
+import { ILookupState, useLookupStore } from 'store/slices';
```
Fosol, line 4: Incorrect.  The admin section is separate by design.  You will however need to update lookups appropriately when changes occur.  Note that there are two different types of challenges here.
a) Admin edits values and their current session should update their local lookups
b) Other users have no idea that things have changed unless they do a hard refresh, as their lookups are cached for a time.

Fosol (changes requested):

app/editor/src/store/hooks/admin/products/useProducts.ts
jdtoombs: Sounds good - thanks for the help!

app/editor/src/hooks/useLookupSync.ts
```diff
@@ -0,0 +1,20 @@
+import _ from 'lodash';
```
jdtoombs, line 1: Can eventually add the other lookups edited in the admin store

app/editor/src/store/hooks/admin/products/useProducts.ts
Fosol: You shouldn't have to worry about issue (b).  If a user has some stale content they'll have to live with that until their cache expires.

app/editor/src/hooks/useLookupSync.ts
jdtoombs: Maybe this should be moved to the admin areas so it is updated as the user submits - versus when they load the list view..

app/editor/src/store/hooks/content/useContent.ts
```diff
@@ -45,7 +45,17 @@ export const useContent = (props?: IContentProps): [IContentState, IContentContr
       return await dispatch('add-content', () => api.addContent(content));
     },
     updateContent: async (content: IContentModel) => {
-      return await dispatch('update-content', () => api.updateContent(content));
+      const result = await dispatch('update-content', () => api.updateContent(content));
+      if (state.content) {
```
jdtoombs, line 49: We just weren't adding content to the redux store when it was updated

app/editor/src/store/hooks/admin/products/useProducts.ts
jdtoombs: @Fosol updated and also added TNO-457 :)

app/editor/src/utils/getSortableOptions.ts
```diff
@@ -14,5 +15,5 @@ export const getSortableOptions = <T extends ISortableModel<any>>(
   prepend: IOptionItem[] = [],
   map: (item: T) => IOptionItem = (i) => new OptionItem(i.name, i.id),
 ) => {
-  return prepend.concat([...items].sort(sortSortable).map(map));
+  return prepend.concat([...filterEnabled(items)].sort(sortSortable).map(map));
```
Fosol, line 18: Not exactly what I was thinking, but should be fine for now.  I was thinking the `getSortableOptions` probably should be extended to have an option filter by enabled.  However, that may have undesirable outcomes too.

app/editor/src/store/hooks/lookup/utils/filterEnabled.ts
```diff
@@ -0,0 +1,3 @@
+export const filterEnabled = (data: any[]) => {
```
Fosol, line 1: We should document our methods more often.  I'm guilty of the same in javascript presently.  Trying to get work done quickly.

app/editor/src/store/hooks/content/useContent.ts
```diff
@@ -45,7 +45,17 @@ export const useContent = (props?: IContentProps): [IContentState, IContentContr
...
+          ...state.content,
+          items: state.content.items.map((i: IContentModel) => {
+            if (i.id === content.id) return content;
+            return i;
+          }),
+        });
+      }
+      return result;
     },
     deleteContent: async (content: IContentModel) => {
```
Fosol, line 60: Need to update the store here appropriately I believe.

app/editor/src/store/hooks/content/useContent.ts
```diff
@@ -45,7 +45,17 @@ export const useContent = (props?: IContentProps): [IContentState, IContentContr
       return await dispatch('add-content', () => api.addContent(content));
```
Fosol, line 45: Need to add to store here appropriately too I believe.

Fosol (changes requested):

app/editor/src/store/hooks/lookup/utils/filterEnabled.ts
jdtoombs: Will add some in now

jdtoombs:
@Fosol, updated again!

app/editor/src/hooks/useLookupSync.ts
Fosol: Yes it should be in the admin, but for now this will do.

app/editor/src/store/hooks/content/useContent.ts
```diff
@@ -42,13 +42,34 @@ export const useContent = (props?: IContentProps): [IContentState, IContentContr
       return result;
     },
     addContent: async (content: IContentModel) => {
-      return await dispatch('add-content', () => api.addContent(content));
+      const result = await dispatch('add-content', () => api.addContent(content));
+      if (state.content) {
```
Fosol, line 46: I don't think `state.content` would be positive if a user was adding their first content item.

app/editor/src/store/hooks/content/useContent.ts
```diff
@@ -42,13 +42,34 @@ export const useContent = (props?: IContentProps): [IContentState, IContentContr
...
+            if (i.id === content.id) return content;
+            return i;
+          }),
+        });
+      }
+      return result;
     },
     deleteContent: async (content: IContentModel) => {
-      return await dispatch('delete-content', () => api.deleteContent(content));
+      const result = await dispatch('delete-content', () => api.deleteContent(content));
```
Fosol, line 65: When content is loaded into the form does `state.content` get set?  If not, this would never be true.

app/editor/src/store/hooks/content/useContent.ts
jdtoombs: I believe it still returns something like:
`
...
items: [],
total: 0
`
when there are no items

jdtoombs: Hmm ok works again, but found a weird issue with `publishedOn` that happens occasionally

app/editor/src/store/hooks/content/useContent.ts
jdtoombs: Yup, seems to be working!

Fosol (approved):

PR #303, opened 2022-09-30
Update image service

AlessiaYChen:
update path and log exceptions

### DIFF
--- a/services/net/image/ImageAction.cs
+++ b/services/net/image/ImageAction.cs
@@ -167,7 +167,7 @@ public class ImageAction : IngestAction<ImageOptions>
     protected string GetInputPath(IngestModel ingest)
     {
         var currentDate = GetLocalDateTime(ingest, DateTime.Now);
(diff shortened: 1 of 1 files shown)


services/net/image/ImageAction.cs
```diff
@@ -166,7 +172,8 @@ protected string GetOutputPath(IngestModel ingest)
     /// <returns></returns>
     protected string GetInputPath(IngestModel ingest)
     {
-        return Path.Combine(ingest.SourceConnection?.GetConfigurationValue("path") ?? "", $"{GetLocalDateTime(ingest, DateTime.Now):yyyy/MM/dd/}");
+        var currentDate = GetLocalDateTime(ingest, DateTime.Now);
+        return Path.Combine(ingest.SourceConnection?.GetConfigurationValue("path") ?? "", currentDate.Year.ToString(), currentDate.Month.ToString("00"), currentDate.Day.ToString("00"));
```
Fosol, line 176: Wouldn't these result in the same value?

Fosol (approved):

PR #304, opened 2022-09-30
Fix Kafka Producers

Fosol:
Our ingest services update a `Content Reference` to reduce the possibilities of duplicates.  However, a race condition is occurring where the ingest service is posting content to Kafka and the `Content Service` has already imported it before the ingest service has even finished updating the `Content Reference`.

To resolve this ingest service `Kafka Producers` will make an additional request to the `Content Reference` to determine if it has already been imported.

Additionally I have added default "Other Arguments" for `ffmpeg` when the ingest is configured for an audio stream.

### DIFF
--- a/app/editor/src/features/admin/ingests/configurations/Audio.tsx
+++ b/app/editor/src/features/admin/ingests/configurations/Audio.tsx
@@ -9,7 +9,7 @@ import { ServiceTypes } from './constants';
 import * as styled from './styled';
 
 export const Audio: React.FC = (props) => {
-  const { values } = useFormikContext<IIngestModel>();
+  const { values, setFieldValue } = useFormikContext<IIngestModel>();
 
   const serviceType = ServiceTypes.find((t) => t.value === values.configuration.serviceType);
 
@@ -20,6 +20,12 @@ export const Audio: React.FC = (props) => {
         name="configuration.serviceType"
         options={ServiceTypes}
         value={serviceType}
+        onChange={(newValue: any) => {
+          // If an audio ingest, set the configuration other args.
+          if (values.id === 0 && !values.configuration.otherArgs && newValue.value === 'stream') {
+            setFieldValue('configuration.otherArgs', '-acodec mp3 -ab 257k');
+          }
+        }}
       />
       <Show visible={values.configuration.serviceType === 'stream'}>
         <AudioStream />
--- a/app/editor/src/features/admin/ingests/configurations/AudioStream.tsx
+++ b/app/editor/src/features/admin/ingests/configurations/AudioStream.tsx
@@ -58,6 +58,7 @@ export const AudioStream: React.FC = (props) => {
         label="Other Arguments"
         name="configuration.otherArgs"
         tooltip="Any other arguments to pass to the command"
+        placeholder="-acodec mp3 -ab 257k"
       />
     </styled.IngestType>
   );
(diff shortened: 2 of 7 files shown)


app/editor/src/features/admin/ingests/configurations/Audio.tsx
```diff
@@ -20,6 +20,12 @@ export const Audio: React.FC = (props) => {
         name="configuration.serviceType"
         options={ServiceTypes}
         value={serviceType}
+        onChange={(newValue: any) => {
+          // If an audio ingest, set the configuration other args.
+          if (values.id === 0 && !values.configuration.otherArgs && newValue.value === 'stream') {
+            setFieldValue('configuration.otherArgs', '-acodec mp3 -ab 257k');
```
Fosol, line 26: Set default value

app/editor/src/features/admin/ingests/configurations/AudioStream.tsx
```diff
@@ -58,6 +58,7 @@ export const AudioStream: React.FC = (props) => {
         label="Other Arguments"
         name="configuration.otherArgs"
         tooltip="Any other arguments to pass to the command"
+        placeholder="-acodec mp3 -ab 257k"
```
Fosol, line 61: Based on blogs this seems to be a good idea... I have no idea really.  I'm just testing if it's better, but so far it seems to work the same as without it.

services/net/capture/CaptureAction.cs
```diff
@@ -97,9 +97,15 @@ public CaptureAction(IKafkaMessenger producer, IApiService api, IOptions<Capture
...
 
-                    // Assuming some success at this point, even though a stop command can be called for different reasons.
-                    reference.Status = (int)WorkflowStatus.Received;
-                    await this.Api.UpdateContentReferenceAsync(reference);
+                    // The content service will often already have imported this content before we can update the content reference.
+                    reference = await this.Api.FindContentReferenceAsync(reference.Source, reference.Uid);
+                    if (reference != null)
+                    {
+                        // Assuming some success at this point, even though a stop command can be called for different reasons.
+                        if (reference.Status != (int)WorkflowStatus.Imported)
```
Fosol, line 105: If it's already imported, we just want to update the Kafka reference information.

services/net/clip/ClipAction.cs
```diff
@@ -478,9 +486,7 @@ private static string GetOtherArgs(IngestModel ingest)
     private static string GetCopy(IngestModel ingest)
     {
         var value = ingest.GetConfigurationValue("copy");
-        // Attempting to use this copy command is much faster, but it corrupts the output file...
-        // return String.IsNullOrWhiteSpace(value) ? " -c:v copy -c:a copy" : $" {value}";
-        return String.IsNullOrWhiteSpace(value) ? "" : $" {value}";
+        return String.IsNullOrWhiteSpace(value) ? " -c:v copy -c:a copy" : $" {value}";
```
Fosol, line 489: Put it back the way it was as this is far more efficient.  The real issue appears to be related to the stream missing content.

services/net/image/ImageAction.cs
```diff
@@ -134,8 +134,15 @@ public ImageAction(IKafkaMessenger producer, IApiService api, IOptions<ImageOpti
...
                     }
 
-                    reference.Status = (int)WorkflowStatus.Received;
-                    await this.Api.UpdateContentReferenceAsync(reference);
+                    // The content service will often already have imported this content before we can update the content reference.
+                    reference = await this.Api.FindContentReferenceAsync(reference.Source, reference.Uid);
+                    if (reference != null)
+                    {
+                        // Assuming some success at this point, even though a stop command can be called for different reasons.
+                        if (reference.Status != (int)WorkflowStatus.Imported)
```
Fosol, line 142: If it's already imported, we just want to update the Kafka reference information.

PR #307, opened 2022-10-03
Fix Kafka Producers

Fosol:
Kafka producers used in our ingest services were running into a race condition with the Content Service.  We maintain a reference to content being ingested to reduce duplication.  However, it serves as a lock, or a single point of entry.  Which when processes are being performed by different processes at different rates it can get out of sequence.

I've updated the logic to check if an optimistic concurrency error has occurred.  If it has, it can then correctly update the kafka position in the reference without changing the status.


libs/net/services/Actions/IngestAction.cs
```diff
@@ -47,5 +53,64 @@ public IngestAction(IApiService api, IOptions<TOptions> options) : base(api, opt
...
+    }
+
+    /// <summary>
+    /// Send AJAX request to api to update content reference.
+    /// This content reference has been successfully received by Kafka.
+    /// Handles optimistic concurrency issue resulting from a race condition with the Content Service.
+    /// </summary>
+    /// <param name="reference"></param>
+    /// <returns></returns>
+    protected virtual async Task<ContentReferenceModel?> UpdateContentReferenceAsync(ContentReferenceModel reference)
```
Fosol, line 85: Shared function to reduce duplication.

PR #309, opened 2022-10-03
Mapping paper names to source configuration

sturufous:
This PR is pretty straightforward for anyone who understands the image service as it migrates functionality from that service into the filemonitor service.  

One thing to note...I had to change a single line in the image service to run with the new connection functionality.

### DIFF
--- a/services/net/filemonitor/FilemonitorAction.cs
+++ b/services/net/filemonitor/FilemonitorAction.cs
@@ -582,7 +582,7 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
     /// </summary>
     /// <param name="ingest"></param>
     /// <returns></returns>
-    /// <exception cref="ArgumentException"></exception>
+    /// <exception cref="FormatException"></exception>
     private Dictionary<string, string> GetSourcesForIngester(IngestModel ingest)
     {
         var sources = new Dictionary<string, string>();
@@ -599,7 +599,7 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
             }
             else
             {
-                throw new FormatException($"Invalid source value in ingest configuration. Source '{source}' for ingest '{ingest.Name}'");
+                throw new ConfigurationException($"Invalid source value in ingest configuration. Source '{source}' for ingest '{ingest.Name}'");
             }
         }
 
--- a/services/net/filemonitor/FilemonitorAction.cs
+++ b/services/net/filemonitor/FilemonitorAction.cs
@@ -7,11 +7,15 @@ using Microsoft.Extensions.Options;
 using TNO.API.Areas.Services.Models.ContentReference;
 using TNO.API.Areas.Services.Models.Ingest;
 using TNO.Core.Http;
+using TNO.Core.Extensions;
 using TNO.Entities;
 using TNO.Kafka.Models;
 using TNO.Kafka;
+using Renci.SshNet;
+using Renci.SshNet.Sftp;
 using TNO.Models.Extensions;
 using TNO.Services.Actions;
+using TNO.Services.Actions.Managers;
 using TNO.Services.FileMonitor.Config;
 using TNO.Core.Exceptions;
 using System.Net;
@@ -64,7 +68,8 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
     public override async Task PerformActionAsync<T>(IIngestServiceActionManager manager, string? name = null, T? data = null, CancellationToken cancellationToken = default) where T : class
     {
         _logger.LogDebug($"Performing ingestion service action for ingest '{manager.Ingest.Name}'");
-        var dir = Path.Join(_options.Value.VolumePath, manager.Ingest.GetConfigurationValue(Fields.ImportDir));
+
+        var dir = GetMonitorPath(manager.Ingest);
         var articles = new List<SourceContent>();
 
         if (String.IsNullOrEmpty(dir))
@@ -72,6 +77,7 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
             throw new InvalidOperationException($"No import directory defined for ingest '{manager.Ingest.Name}'");
         }
 
+        await FetchFiles(manager);
         var format = manager.Ingest.GetConfigurationValue(Fields.FileFormat);
         var selfPublished = manager.Ingest.GetConfigurationValue<bool>(Fields.SelfPublished);
         var sources = selfPublished ? new Dictionary<string, string>() : GetSourcesForIngester(manager.Ingest);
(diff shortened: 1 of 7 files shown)


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -88,6 +98,129 @@ public FileMonitorAction(IHttpRequestClient httpClient, IApiService api, IKafkaM
         await ImportArticlesAsync(manager, articles);
     }
 
+    /// <summary>
+    /// Retrieve files for this connection from the remote server.
+    /// </summary>
+    /// <param name="manager"></param>
+
+    /// <returns></returns>
+    public async Task FetchFiles(IIngestServiceActionManager manager)
```
Fosol, line 107: You don't need to change but at some point I suspect we really should not be copying these files over.  However, we'll need to figure out to best read them and determine if the file has already been parsed.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -88,6 +98,129 @@ public FileMonitorAction(IHttpRequestClient httpClient, IApiService api, IKafkaM
...
+        filePattern = filePattern.Replace("<date>", $"{GetLocalDateTime(manager.Ingest, DateTime.Now.AddDays(manager.Ingest.GetConfigurationValue<double>("dateOffset"))):yyyyMMdd}");
+        Regex match = new Regex(filePattern);
+        if (String.IsNullOrWhiteSpace(filePattern)) throw new ConfigurationException($"Ingest '{manager.Ingest.Name}' is missing a 'filePattern'.");
+
+        // TODO: create new account to access server
+        // Extract the source connection configuration settings.
+        var remotePath = manager.Ingest.SourceConnection?.GetConfigurationValue("path");
+        var username = manager.Ingest.SourceConnection?.GetConfigurationValue("username");
+        var keyFileName = manager.Ingest.SourceConnection?.GetConfigurationValue("keyFileName");
+        var hostname = manager.Ingest.SourceConnection?.GetConfigurationValue("hostname");
```
Fosol, line 120: Good use of specific type of exception.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -88,6 +98,129 @@ public FileMonitorAction(IHttpRequestClient httpClient, IApiService api, IKafkaM
...
+            foreach (var file in files)
+            {
+                await CopyFile(client, manager.Ingest, Path.Combine(remotePath, file.Name));
+            }
+
+            client.Disconnect();
+        }
+        else
+        {
+            _logger.LogError("SSH Private key file does not exist: {file}", sshKeyFile);
```
Fosol, line 151: Should this throw an exception or is ignoring the issue and just logging suitable?

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -88,6 +98,129 @@ public FileMonitorAction(IHttpRequestClient httpClient, IApiService api, IKafkaM
...
+    }
+
+    /// <summary>
+    /// Convert to timezone and return as local.
+    /// Dates should be stored in the timezone of the data source.
+    /// </summary>
+    /// <param name="ingest"></param>
+    /// <param name="date"></param>
+    /// <returns></returns>
+    protected DateTime GetLocalDateTime(IngestModel ingest, DateTime date)
```
Fosol, line 219: We need to at some point make this a shared function as we use it in a few places now.  Probably belongs in a parent class.  You don't need to change it now though.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -323,21 +457,28 @@ private string GetFmsData(string story, string key, IngestModel ingest)
...
-                    item.Page = GetXmlData(story, Fields.Page, ingest);
-                    item.Section = GetXmlData(story, Fields.Section, ingest);
-
-                    item.Authors = GetAuthorList(GetXmlData(story, Fields.Author, ingest));
-                    item.PublishedOn = GetPublishedOn(GetXmlData(story, Fields.Date, ingest), ingest);
-
-                    articles.Add(item);
+                    var code = GetItemSourceCode(ingest, papername, sources);
+
+                    if (!string.IsNullOrEmpty(code)) // This is a valid newspaper source
```
Fosol, line 463: If there is no code we seem to ignore it.  Shouldn't we at least log it?

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -323,21 +457,28 @@ private string GetFmsData(string story, string key, IngestModel ingest)
...
+                        item.ProductId = ingest.ProductId;
+
+                        item.Authors = GetAuthorList(GetXmlData(story, Fields.Author, ingest));
+                        item.PublishedOn = GetPublishedOn(GetXmlData(story, Fields.Date, ingest), ingest);
+
+                        articles.Add(item);
+                    }
                 }
             }
             catch (Exception ex)
```
Fosol, line 484: Doesn't seem right to log a warning and then throw the exception.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -386,22 +528,27 @@ private List<SourceContent> GetFmsArticles(string dir, IIngestServiceActionManag
...
-                    item.Section = GetFmsData(filtered, Fields.Section, ingest);
-
-                    item.Tags = GetTagList(filtered, ingest);
-                    item.Authors = GetAuthorList(GetFmsData(filtered, Fields.Author, ingest));
-                    item.PublishedOn = GetPublishedOn(GetFmsData(filtered, Fields.Date, ingest), ingest);
-
-                    articles.Add(item);
+                    code = string.IsNullOrEmpty(code) ? GetItemSourceCode(ingest, papername, sources) : code;
+
+                    if (!string.IsNullOrEmpty(code)) // This is a valid newspaper source
```
Fosol, line 533: If there is no code we seem to ignore it. Shouldn't we at least log it?

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -386,22 +528,27 @@ private List<SourceContent> GetFmsArticles(string dir, IIngestServiceActionManag
...
+
+                        item.Tags = GetTagList(filtered, ingest);
+                        item.Authors = GetAuthorList(GetFmsData(filtered, Fields.Author, ingest));
+                        item.PublishedOn = GetPublishedOn(GetFmsData(filtered, Fields.Date, ingest), ingest);
+
+                        articles.Add(item);
+                    }
                 }
             }
             catch (Exception ex)
```
Fosol, line 554: Doesn't seem right to log a warning and then throw the exception.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -415,35 +562,67 @@ private List<SourceContent> GetFmsArticles(string dir, IIngestServiceActionManag
...
+        {
+            var pair = source.Split('=');
+
+            if (pair.Length == 2)
+            {
+                sources.Add(pair[0], pair[1]);
+            }
+            else
+            {
+                throw new ArgumentException($"Invalid source value in ingest configuration. Source '{source}' for ingest '{ingest.Name}'");
```
Fosol, line 619: Probably should be a `ConfigurationException` like the other similar errors.

services/net/image/ImageAction.cs
```diff
@@ -73,7 +73,7 @@ public ImageAction(IKafkaMessenger producer, IApiService api, IOptions<ImageOpti
         this.Logger.LogDebug("Performing ingestion service action for data source '{name}'", manager.Ingest.Name);
 
         // Extract the ingest configuration settings.
-        var inputFileCode = String.IsNullOrWhiteSpace(manager.Ingest.GetConfigurationValue("code")) ? manager.Ingest.Source?.Code : manager.Ingest.GetConfigurationValue("code");
+        var inputFileCode = String.IsNullOrWhiteSpace(manager.Ingest.SourceConnection?.GetConfigurationValue("code")) ? manager.Ingest.Source?.Code : manager.Ingest.SourceConnection?.GetConfigurationValue("code");
```
Fosol, line 76: I don't think this is correct.  The code would be specific to an `Ingest.Configuration`, not the `SourceConnection.Configuration`.  Or at least it makes more sense to me that the connection is more generic about where files are, and the ingest is all about how to ingest those files.

Fosol (changes requested):

services/net/filemonitor/FilemonitorAction.cs
sturufous: Changed it to an exception, as the service can't do anything without the key file and should stop operating after max retries.

services/net/filemonitor/FilemonitorAction.cs
sturufous: A warning is logged in GetItemSourceCode().

services/net/filemonitor/FilemonitorAction.cs
sturufous: I'm doing this because the exception that is thrown will not identify the name of the ingest being processed and that info might aid in troubleshooting. Do you think there's value in that? If not I can remove the try/catch block and let the exception bubble up.

services/net/filemonitor/FilemonitorAction.cs
sturufous: See previous comment.

services/net/filemonitor/FilemonitorAction.cs
sturufous: Fixed.

services/net/image/ImageAction.cs
sturufous: Your concern likely relates to Alessia's interpretation of what `code` means in the connection configuration. This is what she sent me as example content for the image service's connection.configuration field:

{"path":"/dsk98/binaryroot/2022/09/27","username":"ckayfish","hostname":"scharnhorst.tno.gov.bc.ca","keyFileName":"id_rsa","code":"sv-GLB"}

In this case, she used `code` as the file name pattern to filter on. Does this provide clarity?

services/net/filemonitor/FilemonitorAction.cs
Fosol: Then throw an exception with the appropriate information in it.  No need to log here, if it will be logged somewhere else.

services/net/image/ImageAction.cs
Fosol: Then I believe my understanding is correct.  The `code` should exist in the `Ingest.Configuration` as it is specific to the ingest, not the connection.

services/net/filemonitor/FilemonitorAction.cs
sturufous: If I catch the exception that is generated by C#, then create and throw a new exception with a descriptive error, will I not lose the information associated with the original exception? This is why I log the warning, ensuring that information is saved, and then throw the original exception. I could also incorporate the original error in the message that I pass to my new exception so that both are reported when the new exception is thrown.

services/net/image/ImageAction.cs
sturufous: Reverted it to the way it worked before.

services/net/filemonitor/FilemonitorAction.cs
Fosol: You include the original in the new.  An exception has an innerException.

sturufous: Fixed it.

services/net/image/ImageAction.cs
```diff
@@ -73,7 +73,7 @@ public ImageAction(IKafkaMessenger producer, IApiService api, IOptions<ImageOpti
         this.Logger.LogDebug("Performing ingestion service action for data source '{name}'", manager.Ingest.Name);
 
         // Extract the ingest configuration settings.
-        var inputFileCode = String.IsNullOrWhiteSpace(manager.Ingest.GetConfigurationValue("code")) ? manager.Ingest.Source?.Code : manager.Ingest.GetConfigurationValue("code");
+        var inputFileCode = String.IsNullOrWhiteSpace(manager.Ingest.GetConfigurationValue("code")) ? manager.Ingest.Source?.Code : manager.Ingest.SourceConnection?.GetConfigurationValue("code");
```
Fosol, line 76: Am I misunderstanding the `code` property?  I would think this is should be `manager.Ingest.GetConfigurationValue("code")`.

Fosol (changes requested):

services/net/image/ImageAction.cs
sturufous: I reverted the line to what you suggested here just before my last push. Not sure why it didn't apply the change. I'll check.

Fosol (approved):

PR #310, opened 2022-10-04
TNO-590: Content list view screen real estate

jdtoombs:
- hover headline for full title
- disable page/section column when `filter.printContent` is not selected
- fix bug where the print content filter checkbox was not "checking"
- reduce size of font in table body

With `setHiddenColumns` now exposed we can easily add a way to disable more columns - if that is desired. The ticket is pretty open ended so showing what I have for now, and welcome to input! 

tno-590


libs/npm/core/src/components/grid-table/GridTable.tsx
```diff
@@ -184,6 +190,12 @@ export const GridTable = <T extends object>({
     pageSize,
   };
 
+  // The user / system disables a column
+  React.useEffect(() => {
+    if (hiddenColumns) setHiddenColumns(hiddenColumns);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Fosol, line 196: Every time I see this I cringe.  It almost always means we're doing something wrong.  What causes the infinite loop here if you include `setHiddenColumns`?

app/editor/src/features/content/list-view/ContentFilter.tsx
```diff
@@ -139,6 +139,7 @@ export const ContentFilter: React.FC<IContentFilterProps> = ({ search }) => {
                     ...filter,
                     pageIndex: 0,
                     contentType: e.target.checked ? ContentTypeName.PrintContent : undefined,
+                    printContent: e.target.checked,
```
Fosol, line 142: The filter shouldn't have a `printContent` property anymore.  The `contentType==ContentTypeName.PrintContent` is the actual filter.  We don't want to properties to do the same thing.

Fosol (changes requested):

libs/npm/core/src/components/grid-table/GridTable.tsx
jdtoombs: I was trying to debug why `setHiddenColumns` was triggering infinite loop - had no luck yet. Will spend a little more time on it. I know `setState` normally causes a similar issue. Will do some more research

app/editor/src/features/content/list-view/ContentFilter.tsx
jdtoombs: Makes sense, just added it in because `value` and `checked` were relying on it. Will adjust those to `contentType==ContentTypeName.PrintContent`

app/editor/src/features/content/list-view/ContentFilter.tsx
```diff
@@ -132,8 +132,8 @@ export const ContentFilter: React.FC<IContentFilterProps> = ({ search }) => {
                 name="isPrintContent"
                 label="Print Content"
                 tooltip="Newspaper content without audio/video"
-                value={filter.printContent}
-                checked={filter.printContent}
+                value={filter.contentType === ContentTypeName.PrintContent}
```
Fosol, line 135: This throws me off, why do we need to set both `value` and `checked`?  Wouldn't the value just be `ContentTypeName.PrintContent`?

Fosol (approved):

app/editor/src/features/content/list-view/ContentFilter.tsx
jdtoombs: We don't technically need the value as we are just using if it is checked or not to determine the filter (and we are not submitting) - I could remove the value attribute from the rest of the checkboxes if that is desired! Just wasn't sure if there was a reason behind the scenes

libs/npm/core/src/components/grid-table/GridTable.tsx
jdtoombs: Can't figure this one out - some people say ignore the linting warning, others say there are ways around it. This one is specifically hard as `setHiddenColumns` is under the hood with `react-table` 😖

PR #320, opened 2022-10-06
Fix content form layout

Fosol:
The Content Form layout should expand and have vertical scrolling when it is on its own page.  When it is part of the combined page it should fit the page.

image


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -460,7 +460,7 @@ export const ContentForm: React.FC<IContentFormProps> = ({ contentType: initCont
                 <Row>
                   <Show visible={isSnippetForm(contentType)}>
                     <Tabs
-                      className={`tabs ${size === 1 ? 'small' : 'large'}`}
```
Fosol, line 463: Don't need to have a duplicate `tabs` class name.

PR #325, opened 2022-10-07
TNO-408: Sitewide file upload

jdtoombs:
Can now upload to storage and navigate away without stopping the upload. Will add to clip form in a.m.


app/editor/src/upload-context/UploadContext.ts
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
Fosol, line 1: Move files associated with upload-context to `/app/editor/src/components/upload`.  It's a shared component that will eventually be moved to `tno-core`.

app/editor/src/upload-context/UploadContext.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { IUploadFile, IUploadManager } from './interfaces';
+
+export const UploadContext = React.createContext<IUploadManager>({
+  upload: () => {
+    throw Error('UploadContext has no Provider');
+  },
+});
+
+export const UploadFilesContext = React.createContext<IUploadFile>({
```
Fosol, line 11: Add some docs

app/editor/src/upload-context/UploadContextWrapper.tsx
```diff
@@ -0,0 +1,64 @@
...
+      const data = new FormData();
+      data.append('file', file);
+
+      setFile({
+        name: file.name,
+        status: 'pending',
+        loaded: 0,
+        total: file.size,
+      });
+      storage
```
Fosol, line 29: For now this will be okay, but we can't have the multi-upload tightly coupled with the storage hook.

app/editor/src/upload-context/interfaces/FetchState.ts
```diff
@@ -0,0 +1 @@
+export type FetchState = 'idle' | 'pending' | 'success' | 'failure';
```
Fosol, line 1: I prefer enum values over string values.

app/editor/src/upload-context/interfaces/IUploadFile.ts
```diff
@@ -0,0 +1,8 @@
+import { FetchState } from './FetchState';
+
+export interface IUploadFile {
```
Fosol, line 3: This interface name appears to be a little misleading.  It seems to be `IUploadState` would reflect what it is better.

Fosol (changes requested):

app/editor/src/upload-context/interfaces/FetchState.ts
jdtoombs: Switching!

Fosol (approved):

app/editor/src/upload-context/UploadContext.ts
jdtoombs: I thought I had it all working but I must have had some cached items. But for some reason I cannot get the context wrapper to behave as expected when I move it to `/app/editor/src/components/upload`. The wrapper all of a sudden acts as a block to the rest of the context providers. Do you have any idea why that may be? Has been driving me mental. Works fine when I leave the folder as is.

Fosol: Shouldn't matter where you put it.  But some times hot reload gets cache mixed up.  `make refresh n=editor`

jdtoombs: I'm not sure whether this is a local issue or not but I cannot get rid of it - don't want to waste too much more time on it. Makes no sense. I change the same things when dragging it to and from its current location. Works from src no issue...

PR #333, opened 2022-10-12
TNO-680 Fix Print Content Actions

Fosol:
Print Content has been fixed so that the actions work again.

image

### DIFF
--- a/app/editor/src/features/content/form/ContentActions.tsx
+++ b/app/editor/src/features/content/form/ContentActions.tsx
@@ -43,29 +43,47 @@ export const ContentActions: React.FC<IContentActionsProps> = ({
   const [hidden, setHidden] = React.useState<IContentActionCheckbox[]>([]);
 
   React.useEffect(() => {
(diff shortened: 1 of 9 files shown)


app/editor/src/features/content/form/ContentActions.tsx
```diff
@@ -42,30 +42,48 @@ export const ContentActions: React.FC<IContentActionsProps> = ({
...
+      return formActions.map((a) => {
+        const found = state.find((i) => i.id === a.id);
+        return { id: a.id, value: !!a.value ? true : found?.value ?? false };
+      });
+    });
   }, [formActions]);
 
   const options = actions.filter(filter).map((a) => {
     const index = formActions.findIndex((ca) => ca.id === a.id);
-    if (contentType === ContentTypeName.PrintContent) {
```
Fosol, line 65: @jdtoombs You can't set state values this way.  They are read-only.  You have to use the `setFieldValues(...)` to change a `formActions[...].value`.

PR #335, opened 2022-10-13
TNO-649: Pager no longer freezes user out

jdtoombs:
The logic was not fully working as I had forgotten to cast `e.target.value` to a number when checking for it's truthy-ness. Also ensure the user may not enter the negatives.


libs/npm/core/src/components/grid-table/Pager.tsx
```diff
@@ -149,10 +149,11 @@ export const Pager: React.FC<IPagerProps> = ({
               if (Number(e.target.value) > 100) {
                 setPageSize(100);
               }
-              if (!!e.target.value) {
+              if (!!Number(e.target.value)) {
```
jdtoombs, line 152: Sneaky culprit

Fosol (approved):

PR #354, opened 2022-10-21
TNO-715 Prepare for Keycloak Gold Realm CSS

Fosol:
To prepare for the Keycloak Gold Realm our authorization implementation needed to be updated.  Tables related to `Role` and `Claim` have been deleted.  This simplifies our DB and API.

The API now has Keycloak client role authorization attributes which will return 403 if the user token does not include the appropriate claim.

There is still additional work that will be required to integrate with Keycloak Gold Realm, but this PR resolves a number of items which will make the full implementation much easier.

> Requires you to `make renew`, which will nuke both your databases (keycloak, postgres).  I will need to refresh the DEV and TEST environments.

> Update your API `.env` file with the following key `Keycloak__ClientId=89ff6cf4-3755-4329-adab-ccfb74052c97`


Makefile
```diff
@@ -131,9 +131,9 @@ remove: ## Remove all containers
 
 renew: ## Refresh all relevant services that were impacted by prior Pull Request.
 	$(info Refresh all relevant services that were impacted by prior Pull Request.)
-	@make db-refresh
+	@make db-nuke
```
Fosol, line 134: This is important because the Keycloak database needs to be rebuilt too.

api/net/Areas/Admin/Controllers/ActionController.cs
```diff
@@ -1,19 +1,19 @@
...
 using TNO.Entities.Models;
+using TNO.Keycloak;
 
 namespace TNO.API.Areas.Admin.Controllers;
 
 /// <summary>
 /// ActionController class, provides Action endpoints for the api.
 /// </summary>
-[Authorize]
+[ClientRoleAuthorize(ClientRole.Administrator)]
```
Fosol, line 16: This authorizes only users with the client role `administrator`.

api/net/Areas/Admin/Models/User/UserModel.cs
```diff
@@ -104,9 +104,7 @@ public UserModel(Entities.User entity) : base(entity)
         this.EmailVerified = entity.EmailVerified;
         this.LastLoginOn = entity.LastLoginOn;
         this.Note = entity.Note;
-
-        this.Roles = entity.RolesManyToMany.Where(r => r.Role != null).Select(r => new RoleModel(r.Role!));
-        this.Roles = entity.Roles.Select(r => new RoleModel(r));
+        this.Roles = entity.Roles.Split(",").Select(r => r[1..^1]);
```
Fosol, line 107: We store the roles in the database in the following format `"[administrator],[editor]"`.  We do this so that we have a simple way to search for them.  It's not ideal, but for a quick solution it will work for now.

api/net/Areas/Editor/Controllers/LookupController.cs
```diff
@@ -111,16 +112,17 @@ public class LookupController : ControllerBase
...
+
         var actions = _actionService.FindAll();
         var categories = _categoryService.FindAll();
-        var claims = _claimService.FindAll();
         var products = _productService.FindAll();
         var sources = _sourceService.FindAll();
         var license = _licenseService.FindAll();
         var ingestTypes = _ingestTypeService.FindAll();
-        var roles = _roleService.FindAll();
+        var roles = (await _keycloakService.GetRolesAsync(_keycloakOptions.ClientId.Value)).Select(r => r.Name!);
```
Fosol, line 125: We fetch the roles from keycloak instead of our database.  This simplifies managing and syncing these objects.

app/editor/Dockerfile
```diff
@@ -5,6 +5,7 @@ COPY .yarn ./.yarn
 COPY package*.json yarn.lock .yarnrc.yml ./
 
 RUN yarn set version berry
+RUN npx browserslist@latest --update-db
```
Fosol, line 8: When we built the image in Openshift it kept giving a warning stating we should do this.

app/editor/src/features/navbar/NavBar.tsx
```diff
@@ -81,16 +81,16 @@ export const NavBar: React.FC = () => {
...
-            <NavBarItem navigateTo="/admin/sources" label="Sources" claim={Claim.administrator} />
-            <NavBarItem navigateTo="/admin/products" label="Products" claim={Claim.administrator} />
-            <NavBarItem navigateTo="/admin/series" label="Series" claim={Claim.administrator} />
             <NavBarItem
               navigateTo="/admin/categories"
               label="Categories"
               claim={Claim.administrator}
             />
             <NavBarItem navigateTo="/admin/tags" label="Tags" claim={Claim.administrator} />
+            <NavBarItem navigateTo="/admin/series" label="Series" claim={Claim.administrator} />
```
Fosol, line 90: Reorganized the order based on the most used by Scott

libs/net/dal/Services/UserService.cs
```diff
@@ -50,7 +50,7 @@ public IPaged<User> Find(UserFilter filter)
             $"{keyword}%") || EF.Functions.Like(c.LastName.ToLower(), $"{keyword}%"));
         }
         if (!String.IsNullOrWhiteSpace(filter.RoleName))
-            query = query.Where(c => c.Roles.Any(r => r.Name.ToLower() == filter.RoleName.ToLower()));
+            query = query.Where(c => EF.Functions.Like(c.Roles.ToLower(), $"%[{filter.RoleName.ToLower()}]%"));
```
Fosol, line 53: I don't love this solution for searching, but I couldn't think of simple way to search two different data sources (keycloak, tno), and provide paging.

libs/net/entities/User.cs
```diff
@@ -104,14 +104,10 @@ public class User : AuditColumns
...
+    /// get/set - Comma separated list of roles assigned to this user (i.e. "[admin],[editor]").
     /// </summary>
-    public virtual List<Role> Roles { get; } = new List<Role>();
-
-    /// <summary>
-    /// get/set - A collection of roles this user belongs to (many-to-many).
-    /// </summary>
-    public virtual List<UserRole> RolesManyToMany { get; } = new List<UserRole>();
+    [Column("roles")]
+    public string Roles { get; set; } = "";
```
Fosol, line 110: Simplifies the management and syncing between keycloak and tno.

libs/net/keycloak/ClientRole.cs
```diff
@@ -0,0 +1,25 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace TNO.Keycloak;
+
+/// <summary>
+/// ClientRole enum, provides the possible Keycloak client roles.
+/// </summary>
+public enum ClientRole
```
Fosol, line 8: This is a hardcoded list of client roles that must be created in Keycloak for the solution to work.

openshift/kustomize/tekton/base/tasks/build-component.yaml
```diff
@@ -67,7 +67,7 @@ spec:
           memory: 250Mi
           cpu: 250m
         limits:
-          memory: 1Gi
+          memory: 2Gi
```
Fosol, line 70: Needed more memory to build

tools/scripts/gen-env-files.sh
```diff
@@ -189,6 +189,7 @@ echo \
 "ASPNETCORE_ENVIRONMENT=Development
 ASPNETCORE_URLS=http://+:8080
 
+Keycloak__ClientId=89ff6cf4-3755-4329-adab-ccfb74052c97
```
Fosol, line 192: The client Id isn't a secret, but you'll need to add it to your `.env` file.

PR #360, opened 2022-10-24
TNO-712 - Ingest content with unknown source name

sturufous:
git rebase -i still not working. I'll troubleshoot in the morning.

### DIFF
--- a/services/net/filemonitor/FilemonitorAction.cs
+++ b/services/net/filemonitor/FilemonitorAction.cs
@@ -580,7 +580,10 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
             else
             {
                 var defaultSource = ingest.GetConfigurationValue("defaultSource");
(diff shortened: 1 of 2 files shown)


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -579,8 +579,9 @@ private string GetItemSourceCode(IngestModel ingest, string paperName, Dictionar
             }
             else
             {
+                var defaultSource = ingest.GetConfigurationValue("defaultSource");
                 _logger.LogWarning("Paper '{paperName}' not in configuration string for ingest '{ingest.Name}'.", paperName, ingest.Name);
```
Fosol, line 583: If they've configured a default source then it will use it.  There is no reason to log a warning in this case.  Only log the warning if they have not configured a default source and one wasn't found.

Fosol (changes requested):

services/net/filemonitor/FilemonitorAction.cs
sturufous: Fixed.

Fosol (approved):

PR #362, opened 2022-10-26
TNO-711: Double scroll bar fix

jdtoombs:
- Tables now scroll rather than the full form page (like the content list view), also stops the page from jumping around when a pagination value is changed
- Work orders slightly adjusted to match other admin areas
- Double scroll bar no longer on content list view

formpagefix

### DIFF
--- a/app/editor/src/components/form/formpage/styled/FormPage.tsx
+++ b/app/editor/src/components/form/formpage/styled/FormPage.tsx
@@ -13,7 +13,7 @@ export const FormPage = styled.div<IFormPageProps>`
 
   div[role='rowgroup'] {
     min-height: 100px;
-    max-height: calc(100vh - 600px);
+    max-height: calc(100vh - 400px);
     overflow-y: scroll;
     overflow-x: hidden;
   }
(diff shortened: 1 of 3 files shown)


app/editor/src/features/admin/work-orders/WorkOrderList.tsx
```diff
@@ -74,23 +74,21 @@ export const WorkOrderList = () => {
 
   return (
     <styled.WorkOrderList>
-      <FormPage>
```
jdtoombs, line 77: Just removing `FormPage` from here and adding it to the base styled component

app/editor/src/components/form/formpage/styled/FormPage.tsx
```diff
@@ -9,6 +9,12 @@ export const FormPage = styled.div<IFormPageProps>`
   ${(props) => (props.maxWidth !== '' ? `max-width: ${props.maxWidth ?? '1200px'}` : '')};
   padding: 0.5em 2em 0 2em;
   margin: 0px auto;
-  overflow-y: overlay;
-  overflow-x: hidden;
+  overflow: hidden;
```
Fosol, line 12: This always worries me that we'll be cutting things off.

Fosol (approved):

PR #364, opened 2022-10-26
TNO-699: replace p tag with new line in content service

AlessiaYChen:
The Content Service will extract <p> from all content.

### DIFF
--- a/services/net/content/ContentManager.cs
+++ b/services/net/content/ContentManager.cs
@@ -250,8 +250,8 @@ public class ContentManager : ServiceManager<ContentOptions>
                     Headline = String.IsNullOrWhiteSpace(model.Title) ? "[TBD]" : model.Title,
                     Uid = model.Uid,
                     Page = "", // TODO: Provide default page from Data Source config settings.
-                    Summary = String.IsNullOrWhiteSpace(model.Summary) ? "[TBD]" : model.Summary,
-                    Body = !String.IsNullOrWhiteSpace(model.Body) ? model.Body : model.ContentType == ContentType.Snippet ? "" : model.Summary,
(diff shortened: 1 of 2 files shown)


Fosol (approved):

services/net/content/ContentManager.cs
```diff
@@ -250,8 +250,8 @@ private async Task HandleMessageAsync(ConsumeResult<string, SourceContent> resul
                     Headline = String.IsNullOrWhiteSpace(model.Title) ? "[TBD]" : model.Title,
                     Uid = model.Uid,
                     Page = "", // TODO: Provide default page from Data Source config settings.
-                    Summary = String.IsNullOrWhiteSpace(model.Summary) ? "[TBD]" : model.Summary,
-                    Body = !String.IsNullOrWhiteSpace(model.Body) ? model.Body : model.ContentType == ContentType.Snippet ? "" : model.Summary,
+                    Summary = String.IsNullOrWhiteSpace(model.Summary) ? "[TBD]" : StringExtensions.SanitizeContent(model.Summary, "<p>|</p>", "\r\n"),
```
Fosol, line 253: Does this keep the body text of the <p>?

AlessiaYChen: It only removes <p> or </p> tag and add new line without removing any content

AlessiaYChen: If we consider `<p class="XXX">`, we may need to use `<p(?:\\s[^>]*)?>|</p>` instead of `<p>|</p>`. What do you think? Also i find that there are some content with broken tags (the one Carolynn sent), it only has `<p>` without the closing tag `</p>`.

Fosol: Probably have to handle these other scenarios too.

Fosol (approved):

PR #365, opened 2022-10-27
First stab at end-to-end support for product id override on source

sturufous:
Due to a glitch on my end, the commit message is wrong, but this shouldn't affect the content. 

I rejigged the insert statements for the daily print papers in 01-Source.sql to avoid having to edit the other 246 entries.

### DIFF
--- a/app/editor/src/features/admin/sources/SourceDetails.tsx
+++ b/app/editor/src/features/admin/sources/SourceDetails.tsx
@@ -19,6 +19,7 @@ export const SourceDetails: React.FC<ISourceDetailsProps> = () => {
 
   const users = getUserOptions(lookups.users);
   const licenses = getSortableOptions(lookups.licenses);
+  const products = getSortableOptions(lookups.products);
 
   React.useEffect(() => {
     // Ensures the connection settings can display the correct form on initial load.
@@ -63,6 +64,12 @@ export const SourceDetails: React.FC<ISourceDetailsProps> = () => {
           name="ownerId"
           tooltip="The user that manages this content"
           options={users}
+        />
+        <FormikSelect
+          label="Product Designation Override"
+          name="productId"
+          tooltip="The product designation the source content will be assigned (overrides the value in the ingest)"
+          options={products}
         />
       </Col>
       <Col>
--- a/libs/net/dal/Migrations/20221028170850_Initial.cs
+++ b/libs/net/dal/Migrations/20221028170850_Initial.cs
@@ -465,6 +465,12 @@ namespace TNO.DAL.Migrations
                         principalTable: "user",
                         principalColumn: "id",
                         onDelete: ReferentialAction.Restrict);
+                    table.ForeignKey(
+                        name: "FK_source_product_product_id",
+                        column: x => x.product_id,
+                        principalTable: "product",
+                        principalColumn: "id",
+                        onDelete: ReferentialAction.Restrict);
                 });
 
             migrationBuilder.CreateTable(
--- a/libs/net/entities/Source.cs
+++ b/libs/net/entities/Source.cs
@@ -41,6 +41,12 @@ public class Source : BaseType<int>
     public int? OwnerId { get; set; }
 
     /// <summary>
+    /// get/set - Foreign key to product.
+    /// </summary>
+    [Column("product_id")]
+    public int? ProductId { get; set; }
+
+    /// <summary>
     /// get/set - The user who owns this source.
     /// </summary>
     public virtual User? Owner { get; set; }
--- a/services/net/content/ContentManager.cs
+++ b/services/net/content/ContentManager.cs
@@ -245,7 +245,7 @@ public class ContentManager : ServiceManager<ContentOptions>
                     SourceId = source?.Id,
                     OtherSource = model.Source,
                     ContentType = model.ContentType,
-                    ProductId = model.ProductId,
+                    ProductId = productId,
                     LicenseId = source?.LicenseId ?? 1,  // TODO: Default license by configuration.
                     SeriesId = null, // TODO: Provide default series from Data Source config settings.
                     OtherSeries = null, // TODO: Provide default series from Data Source config settings.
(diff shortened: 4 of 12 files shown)


app/editor/src/features/admin/sources/SourceDetails.tsx
```diff
@@ -64,6 +65,12 @@ export const SourceDetails: React.FC<ISourceDetailsProps> = () => {
           tooltip="The user that manages this content"
           options={users}
         />
+        <FormikSelect
```
Fosol, line 68: This currently doesn't allow them to select None.

libs/net/dal/Configuration/SourceConfiguration.cs
```diff
@@ -15,6 +15,7 @@ public override void Configure(EntityTypeBuilder<Source> builder)
         builder.Property(m => m.ShortName).IsRequired().HasMaxLength(100).HasDefaultValueSql("''");
         builder.Property(m => m.Description).IsRequired().HasMaxLength(2000).HasDefaultValueSql("''");
         builder.Property(m => m.OwnerId);
+        builder.Property(m => m.ProductId);
```
Fosol, line 18: You should define the relationship as this is a foreign key.  See the examples below.

services/net/content/ContentManager.cs
```diff
@@ -229,7 +229,8 @@ private async Task HandleMessageAsync(ConsumeResult<string, SourceContent> resul
                 // TODO: Failures after receiving the message from Kafka will result in missing content.  Need to handle this scenario.
                 // TODO: Handle e-tag.
                 var source = await _api.GetSourceForCodeAsync(model.Source);
-                if (model.ProductId == 0)
+                var productId = source.ProductId ?? model.ProductId;
```
Fosol, line 232: Do not put this here.  The ingest that found this content should be making this decision on what the valid `ProductId` should be.

Fosol (changes requested):

services/net/content/ContentManager.cs
sturufous: Fixed.

app/editor/src/features/admin/sources/SourceDetails.tsx
sturufous: I followed the example of the other two dropdowns on the page. If the value of the field is null it displays "Select...". Once it's set you can change the value but can't set it back to null. What's the best way to set it to None in React?

app/editor/src/features/admin/sources/SourceDetails.tsx
```diff
@@ -19,6 +19,7 @@ export const SourceDetails: React.FC<ISourceDetailsProps> = () => {
 
   const users = getUserOptions(lookups.users);
   const licenses = getSortableOptions(lookups.licenses);
+  const products = getSortableOptions(lookups.products);
```
Fosol, line 22: There is an override argument that allows you to prepend options.  You can prepend an option that handles `undefined` value.

Fosol (changes requested):

app/editor/src/features/admin/sources/SourceDetails.tsx
Fosol: I've left another comment about how to prepend an option to handle `undefined`

libs/net/dal/Configuration/SourceConfiguration.cs
```diff
@@ -15,11 +15,13 @@ public override void Configure(EntityTypeBuilder<Source> builder)
...
         builder.Property(m => m.LicenseId).IsRequired();
         builder.Property(m => m.AutoTranscribe).IsRequired();
         builder.Property(m => m.DisableTranscribe).IsRequired();
 
         builder.HasOne(m => m.Owner).WithMany().HasForeignKey(m => m.OwnerId).OnDelete(DeleteBehavior.Restrict);
+        builder.HasOne(m => m.Category).WithMany().HasForeignKey(m => m.ProductId).OnDelete(DeleteBehavior.Restrict);
```
Fosol, line 24: It's not a `Category`

libs/net/entities/Source.cs
```diff
@@ -45,6 +45,17 @@ public class Source : BaseType<int>
...
+    /// <summary>
+    /// get/set - Foreign key to product.
+    /// </summary>
+    [Column("product_id")]
+    public int? ProductId { get; set; }
+
+    /// <summary>
+    /// get/set - The product associated with this data source..
+    /// </summary>
+    public Product? Category { get; set; }
```
Fosol, line 57: It's not a `Category`

Fosol (changes requested):

Fosol:
I believe once you rebase some of the comments I've added will already be resolved as long as you merge correctly.

libs/net/entities/Source.cs
sturufous: I realized this wasn't an ideal name, but I couldn't call it Product as the class is Product. What name do you suggest?

libs/net/dal/Migrations/20221028170850_Initial.cs
```diff
@@ -465,6 +465,12 @@ protected override void Up(MigrationBuilder migrationBuilder)
                         principalTable: "user",
                         principalColumn: "id",
                         onDelete: ReferentialAction.Restrict);
+                    table.ForeignKey(
```
Fosol, line 468: Remove your changes from commit.

libs/net/entities/Source.cs
```diff
@@ -45,6 +45,12 @@ public class Source : BaseType<int>
     /// </summary>
     public virtual User? Owner { get; set; }
 
+    /// <summary>
+    /// get/set - Foreign key to product.
+    /// </summary>
+    [Column("product_id")]
```
Fosol, line 51: Remove your changes from commit.

Fosol (changes requested):

libs/net/dal/Configuration/SourceConfiguration.cs
Fosol: Remove this change from your commit.

Fosol (changes requested):

PR #370, opened 2022-10-31
TNO-710 Raspberry Pi integration

Fosol:
To support the ability to run services in remote locations a number of changes were implemented to the database schema.  We now have Data Locations, which provide a way to identify where services are running and how to connect to these data locations.  Additionally, there is an updated UI for Ingest management that provides a way to configure FFmpeg for a Raspberry Pi.

> Requires `make renew` to refresh the database.  DEV and TEST will need to be refreshed too.

## Summary

- Enable ARM builds
- Update Initial DB Migration
- Add data location admin
- Fix created audit values
- Removed Kafka dependency for ingest Services
- Update ingest admin

image


libs/net/dal/Services/BaseService`.cs
```diff
@@ -46,10 +47,26 @@ public virtual TEntity Update(TEntity entity)
     {
         if (entity == null) throw new ArgumentNullException(nameof(entity));
 
-        this.Context.Entry(entity).State = EntityState.Modified;
+        var entry = this.Context.Entry(entity);
+        if (entry.State == EntityState.Detached && entity is AuditColumns audit)
```
Fosol, line 51: Fixed issue where created audit information was being overwritten

libs/net/entities/ConnectionType.cs
```diff
@@ -10,38 +10,38 @@ public enum ConnectionType
     /// </summary>
     LocalVolume = 0,
 
-    /// <summary>
-    /// Files are stored on a remote volume.
-    /// </summary>
-    RemoteVolume = 1,
```
Fosol, line 16: This is removed as it is meaningless

libs/net/entities/Source.cs
```diff
@@ -45,6 +45,17 @@ public class Source : BaseType<int>
     /// </summary>
     public virtual User? Owner { get; set; }
 
+    /// <summary>
+    /// get/set - Foreign key to product that content will be defaulted to.
+    /// </summary>
+    [Column("product_id")]
+    public int? ProductId { get; set; }
```
Fosol, line 52: @sturufous I've added the property to the table, so you shouldn't need to worry about updating the Initial migration.  You will still need to update all the relevant models.

libs/net/services/Runners/IngestService.cs
```diff
@@ -7,7 +7,7 @@ namespace TNO.Services.Runners;
 /// BaseService abstract class, provides a console application that runs a service, and an api.
 /// The main purpose of the IngestService is to configure ingest dependency injection.
 /// </summary>
-public abstract class IngestService : KafkaProducerService
+public abstract class IngestService : BaseService
```
Fosol, line 10: Ingest services no longer require Kafka.

services/net/clip/ClipManager.cs
```diff
@@ -37,19 +37,12 @@ public class ClipManager : IngestManager<ClipIngestActionManager, ClipOptions>
...
-    /// </summary>
-    /// <param name="ingest"></param>
-    /// <returns></returns>
-    private static bool IsClip(IngestModel ingest)
-    {
-        return ingest.GetConfigurationValue("serviceType") == "clip";
+        var serviceType = !String.IsNullOrWhiteSpace(this.Options.ServiceType) ? this.Options.ServiceType : "clip";
+        var hostname = System.Environment.GetEnvironmentVariable("HOSTNAME");
+        return ingests.Where(i =>
+            i.GetConfigurationValue("serviceType") == serviceType &&
```
Fosol, line 43: Provides a way to only run ingests that are configured for the service hostname.  Each RPi will have it's own configuration.

PR #372, opened 2022-10-31
TNO-439: More tag fixes

jdtoombs:
- tags would change order in the tags field based on filtering/matching against existing tags in the DB, the end result would be an alphabetical order of the tags
- as the tags would reset each time `onBlur` would occur it would result in tags being rearranged in order, until the user clicked save they would fall back to alphabetical 

- added a check that ensures the tag order stays the same to avoid confusion to the user

### DIFF
--- a/app/editor/src/features/content/form/ContentSummaryForm.tsx
+++ b/app/editor/src/features/content/form/ContentSummaryForm.tsx
@@ -7,6 +7,7 @@ import { getIn, useFormikContext } from 'formik';
 import { useCombinedView } from 'hooks';
 import { ContentTypeName, IUserModel } from 'hooks/api-editor';
 import { useModal } from 'hooks/modal';
+import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


app/editor/src/features/content/form/ContentSummaryForm.tsx
```diff
@@ -283,8 +294,7 @@ export const ContentSummaryForm: React.FC<IContentSummaryFormProps> = ({
             variant={ButtonVariant.danger}
             className="top-spacer"
             onClick={() => {
-              const regex = /\[.*\]/; // TODO: This is far too eager and could remove valuable content.
-              setFieldValue('summary', values.summary.replace(regex, ''));
+              setFieldValue('summary', values.summary.replace(tagMatch, ''));
```
jdtoombs, line 297: should be less eager now as it only checks the last line

Fosol (approved):

PR #373, opened 2022-11-01
TNO-679: Allow admins to edit user information

jdtoombs:
Everything was setup already, just needed to check if the user was an admin or not

### DIFF
--- a/app/editor/src/features/admin/users/UserForm.tsx
+++ b/app/editor/src/features/admin/users/UserForm.tsx
@@ -21,6 +21,7 @@ import { getEnumStringOptions } from 'utils';
 
 import { defaultUser } from './constants';
 import * as styled from './styled';
+import { isAdmin } from './utils/isAdmin';
 
 /**
  * Provides a User Form to manage, create, update and delete a user.
@@ -40,7 +41,7 @@ export const UserForm: React.FC = () => {
     lookups.roles.map((r) => new OptionItem(r.name, r.id)),
   );
 
-  const isLinkedToKeycloak = user.key !== '00000000-0000-0000-0000-000000000000';
+  const isAdminUser = isAdmin(user);
   const statusOptions = getEnumStringOptions(UserStatusName);
 
   React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


app/editor/src/features/admin/users/UserForm.tsx
```diff
@@ -40,7 +41,7 @@ export const UserForm: React.FC = () => {
     lookups.roles.map((r) => new OptionItem(r.name, r.id)),
   );
 
-  const isLinkedToKeycloak = user.key !== '00000000-0000-0000-0000-000000000000';
+  const isAdminUser = isAdmin(user);
```
Fosol, line 44: You don't need this.  The admin is the only user who should be able to get to this form.

app/editor/src/features/admin/users/utils/isAdmin.ts
```diff
@@ -0,0 +1,10 @@
+import { IUserModel } from 'hooks';
+
+/**
+ * Check whether the user is an admin or not.
+ * @param user the user to check
+ * @returns  true if the user is an admin, false otherwise
+ */
+export const isAdmin = (user: IUserModel) => {
```
Fosol, line 8: Not required

app/editor/src/features/admin/users/UserForm.tsx
```diff
@@ -40,7 +41,7 @@ export const UserForm: React.FC = () => {
     lookups.roles.map((r) => new OptionItem(r.name, r.id)),
   );
 
-  const isLinkedToKeycloak = user.key !== '00000000-0000-0000-0000-000000000000';
```
Fosol, line 43: This logic was to protect users that existed in Keycloak.  The reality is that in the future we won't be able to edit users in keycloak anyway.  So I'm wondering if there is any value in making this change.  For now you can simply comment this line of code out, as something similar will need to be added back.

Fosol (changes requested):

app/editor/src/features/admin/users/UserForm.tsx
jdtoombs: Sounds good!

app/editor/src/features/admin/users/UserForm.tsx
jdtoombs: Figured that may be the case... will get rid of

PR #380, opened 2022-11-03
Change api Dockerfile image

Fosol:
This may not work, but will attempt to use a prebuilt image for aspnet:6.0.  This image is stored in our image registry.

I do not know if our image registry is accessible without authentication.  I'll test with my laptop shortly if the pipeline is able to complete the build.

### DIFF
--- a/api/net/Dockerfile
+++ b/api/net/Dockerfile
@@ -1,5 +1,5 @@
 ARG BUILD_CONFIGURATION=Release
-FROM mcr.microsoft.com/dotnet/aspnet:6.0 AS base
+FROM image-registry.apps.silver.devops.gov.bc.ca/9b301c-tools/aspnet:6.0 AS base
 EXPOSE 443 8080
 
 # Copy csproj and restore as distinct layers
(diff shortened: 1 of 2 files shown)


api/net/Dockerfile.aspnet
```diff
@@ -0,0 +1,5 @@
+FROM mcr.microsoft.com/dotnet/aspnet:6.0 AS base
```
Fosol, line 1: New image that we build locally and push to our image registry manually.  This is to hopefully reduce the size of the overall build.  We need to find a way to reduce the size because Buildah is limited to 10Gb and we're out of drive space...

api/net/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG BUILD_CONFIGURATION=Release
-FROM mcr.microsoft.com/dotnet/aspnet:6.0 AS base
+FROM image-registry.apps.silver.devops.gov.bc.ca/9b301c-tools/aspnet:6.0 AS base
```
Fosol, line 2: I don't know if this requires authentication to pull.

PR #383, opened 2022-11-03
Correct Front page published on date

Fosol:
Front pages are often delivered the evening before their published on date.  There is now a way to add a regex to parse the published on date from the file name.

image


libs/net/dal/Migrations/Initial/Up/PostUp/02-Ingest.sql
```diff
@@ -462,6 +462,7 @@ INSERT INTO public.ingest (
   , frontPageId -- product_id
   , '{ "path": "binaryroot",
       "fileName": "sv-GLB",
+      "publishedOnExpression": "^sv-GLB-[A-Za-z]{3}-(?<day>[0-9]{2})(?<month>[0-9]{2})(?<year>[0-9]{4})",
```
Fosol, line 465: Update DB migration scripts to include the regex statement.

PR #386, opened 2022-11-04
TNO-710 Update Clip service to run remotely

Fosol:
The Clip Service needs access to the capture files.  Which means it either needs to copy a large file across the network, or the Clip Service needs to be run in the same location as the capture files.  For performance, the best option is to run it in the same location.

## Summary
- Clip service creates a temporary file if the stream is a mkv file, so that it can calculate duration.
- Fixed a bug where remote path include `~` for the current user's home directory.
- Fixed bug when saving a new Ingest configuration with a schedule.

### DIFF
--- a/app/editor/src/features/admin/connections/configurations/AzureConfiguration.tsx
+++ b/app/editor/src/features/admin/connections/configurations/AzureConfiguration.tsx
@@ -34,6 +34,7 @@ export const AzureConfiguration = () => {
             label="Username"
             name="configuration.username"
             value={values.configuration?.username}
+            autoComplete="off"
           />
         </Col>
         <Col flex="1 1 0">
@@ -42,7 +43,7 @@ export const AzureConfiguration = () => {
             name="configuration.password"
             value={values.configuration?.password}
             type="password"
-            autoComplete="off"
+            autoComplete="new-password"
           />
         </Col>
         <Col flex="1 1 0">
--- a/app/editor/src/features/admin/ingests/configurations/AudioClip.tsx
+++ b/app/editor/src/features/admin/ingests/configurations/AudioClip.tsx
@@ -22,15 +22,10 @@ export const AudioClip: React.FC = (props) => {
         options={TimeZones}
         value={timeZone}
       />
-      <FormikText
-        label="Format"
-        name="configuration.format"
-        tooltip="Format of the clip"
-        placeholder="mp3"
-      />
+      <FormikText label="Format" name="configuration.format" tooltip="Format of the clip" />
       <p>Use "{'{schedule.Name}'}.mp3" to name the file with the schedule name.</p>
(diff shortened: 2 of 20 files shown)


app/editor/src/features/admin/connections/configurations/AzureConfiguration.tsx
```diff
@@ -42,7 +43,7 @@ export const AzureConfiguration = () => {
             name="configuration.password"
             value={values.configuration?.password}
             type="password"
-            autoComplete="off"
+            autoComplete="new-password"
```
Fosol, line 46: Hopefully the browser will not ask to save these passwords.

app/editor/src/features/admin/ingests/configurations/AudioClip.tsx
```diff
@@ -22,15 +22,10 @@ export const AudioClip: React.FC = (props) => {
         options={TimeZones}
         value={timeZone}
       />
-      <FormikText
-        label="Format"
-        name="configuration.format"
-        tooltip="Format of the clip"
-        placeholder="mp3"
-      />
+      <FormikText label="Format" name="configuration.format" tooltip="Format of the clip" />
```
Fosol, line 25: Removed placeholder to reduce confusion when using different formats.  FFmpeg can determine the format by the output file name.

services/net/clip/ClipAction.cs
```diff
@@ -398,9 +415,34 @@ private async Task<long> GetDurationAsync(string inputFile)
...
+        return float.TryParse(output, out float value) ? (long)Math.Floor(value) : 0;
+    }
+
+    /// <summary>
+    /// Repackage the file into a different type.
+    /// </summary>
+    /// <param name="inputFile"></param>
+    /// <param name="format"></param>
+    /// <returns></returns>
+    private static async Task<string> RepackageFileAsync(string inputFile, string format = "mp4")
```
Fosol, line 428: Need to convert `mkv` into `mp4` to get the duration of the video regrettably.  There are warnings thrown and an error during this process that need to be reviewed, but it appears to function correctly.

services/net/content/ContentManager.cs
```diff
@@ -385,7 +385,7 @@ private async Task<ContentModel> CopyFileWithSSHAsync(DataLocationModel dataLoca
         Stream? file = null;
         try
         {
-            var fullPath = path.CombineWith(model.FilePath.MakeRelativePath());
+            var fullPath = path.CombineWith(model.FilePath.MakeRelativePath()).Replace("~/", $"{client.WorkingDirectory}/");
```
Fosol, line 388: If the connection is a relative path for the current user, it will change this so that the client can find the file.  For some reason the SSH client doesn't know what a `~` is.

PR #390, opened 2022-11-09
Changes for TNO-709 in new branch TNO-709-2

sturufous:
This should contain all the changes in the original branch TNO-709, plus the following:

1. Code to handle the situation where some downloaded newspaper import files are empty. I had to do this now, rather than opening another ticket, as I could not test the FileMonitor service without this change. Monday November 7th is the only day that empty files have been provided on the Globe and Mail FTP server.
2. As the GLOBE ingest is the only self-published ingest and it's a daily print newspaper, I added a statement in the migration to set the product_id for the ingest to "Daily Print". This is required for the correct product designation to display for Globe and Mail articles.

### DIFF
--- a/api/net/Areas/Admin/Models/Ingest/SourceModel.cs
+++ b/api/net/Areas/Admin/Models/Ingest/SourceModel.cs
@@ -47,6 +47,11 @@ public class SourceModel : AuditColumnsModel
     /// get/set -
     /// </summary>
     public int? OwnerId { get; set; }
+
+    /// <summary>
+    /// get/set -
+    /// </summary>
+    public int? ProductId { get; set; }
     #endregion
 
     #region Constructors
@@ -69,6 +74,7 @@ public class SourceModel : AuditColumnsModel
         this.IsEnabled = entity.IsEnabled;
         this.LicenseId = entity.LicenseId;
         this.OwnerId = entity.OwnerId;
+        this.ProductId = entity.ProductId;
     }
     #endregion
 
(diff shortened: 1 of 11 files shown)


libs/net/dal/Migrations/20221028170850_Initial.cs
```diff
@@ -453,18 +453,18 @@ protected override void Up(MigrationBuilder migrationBuilder)
                         principalTable: "license",
```
Fosol, line 453: Remove this file from your commit

libs/net/entities/Source.cs
```diff
@@ -41,15 +41,15 @@ public class Source : BaseType<int>
     public int? OwnerId { get; set; }
 
     /// <summary>
-    /// get/set - The user who owns this source.
+    /// get/set - Foreign key to product.
```
Fosol, line 44: Remove your changes from this commit.

services/net/filemonitor/appsettings.json
```diff
@@ -14,7 +14,8 @@
     "VolumePath": "/data",
     "IngestTypes": "Paper",
     "TimeZone": "Pacific Standard Time",
-    "PrivateKeysPath": "keys"
+    "PrivateKeysPath": "keys",
+    "Datalocation": "Openshift"
```
Fosol, line 18: Remove change from this commit

Fosol (changes requested):

Fosol (approved):

PR #398, opened 2022-11-15
TNO-730: Clear button on react-select

jdtoombs:
TODO: 
- [x] Cleanup styling for various forms 

- Can now click an "x" within the react-select to clear input, works for both formik and non-formik
- Also removed some dead CSS
tno-730

### DIFF
--- a/app/editor/src/components/form/select/styled/Select.tsx
+++ b/app/editor/src/components/form/select/styled/Select.tsx
@@ -6,6 +6,24 @@ export const Select = styled(Col)`
     content: ' *';
     color: ${(props) => props.theme.css.dangerColor};
   }
+  margin-right: 0.5em;
+  .select-container {
+    width: 100%;
+  }
+  .input-container {
+    position: relative;
+    .clear {
+      :hover {
+        color: ${(props) => props.theme.css.dangerColor};
+        cursor: pointer;
+      }
+      position: absolute;
+      right: 55px;
+      align-self: center;
+      font-weight: 500;
+      color: #cccccc;
+    }
+  }
 
   & > input {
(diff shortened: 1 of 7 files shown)


app/editor/src/features/content/list-view/ContentFilter.tsx
```diff
@@ -90,9 +90,10 @@ export const ContentFilter: React.FC<IContentFilterProps> = ({ search }) => {
           label="Product Designation"
           options={productOptions}
           value={productOptions.find((mt) => mt.value === filter.productId)}
+          width={FieldSize.Big}
           defaultValue={productOptions[0]}
           onChange={(newValue) => {
-            var productId = (newValue as IOptionItem).value ?? 0;
+            var productId = !!newValue ? (newValue as IOptionItem).value : 0;
```
jdtoombs, line 96: Fixed bug since you can clear value now

app/editor/src/components/form/select/styled/Select.tsx
```diff
@@ -1,11 +1,29 @@
 import styled from 'styled-components';
 import { Col } from 'tno-core/dist/components/flex';
 
-export const Select = styled(Col)`
+export const Select = styled<any>(Col)`
```
Fosol, line 4: Don't use `any`, it causes problems later and is hard to mainain.

Fosol (changes requested):

app/editor/src/components/form/select/styled/Select.tsx
jdtoombs: Oops thought I removed that, was playing with props at one point

Fosol (approved):

PR #405, opened 2022-11-17
Change default list length in admin views from 10 to 20 items and fix editing of row count in pager

sturufous:


libs/npm/core/src/components/grid-table/Pager.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 import { BiFirstPage, BiLastPage } from 'react-icons/bi';
 
-import { Button, ButtonVariant } from '../button';
-import { Row } from '../flex';
-import { Text } from '../form/text';
-import { Show } from '../show';
+import { Button, ButtonVariant } from 'tno-core';
```
Fosol, line 4: Cannot self reference

Fosol (changes requested):

libs/npm/core/src/components/grid-table/Pager.tsx
sturufous: Those changes weren't supposed to be there. The commit now shows how I wanted it to look.

Fosol (commented):
We still need a new version of the `tno-core` library to be published and referenced.  Looking forward to seeing this Monday.

Fosol (approved):

libs/npm/core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tno-core",
   "description": "TNO shared library",
-  "version": "0.0.187",
+  "version": "0.0.190",
```
Fosol, line 4: Does not match

Fosol (changes requested):

Fosol:
I've included this change in tno-core:0.0.193

PR #414, opened 2022-11-24
TNO-456 Add morning report

Fosol:
I've added a basic Morning Report page.  It is not meant to be fully functional presently, as I don't want to duplicate effort with the main Content Form.

Also published a new version of `tno-core:0.0.193` with updates to the GridTable, PagedTable, Button components.

> Requires a DB refresh, run `make renew`

- Update Initial DB migration, which includes changes made in DEV
- Able to update multiple content items by applying an action to be performed on each item

### TODO

- For some reason unit-tests are failing.  @jdtoombs is looking into this.
- The performance of the `react-table` component we use for the table is starting to be problematic.  It may be how we're using it however.  I may reduce the amount of information returned by the API.

image


api/net/Areas/Admin/Controllers/ActionController.cs
```diff
@@ -86,7 +86,7 @@ public IActionResult FindById(int id)
     [SwaggerOperation(Tags = new[] { "Action" })]
     public IActionResult Add(ActionModel model)
     {
-        var result = _service.Add(model.ToEntity());
+        var result = _service.AddAndSave(model.ToEntity());
```
Fosol, line 89: `AddAndSave(...)` replaces the old `Add(...)` methods as the default we've been using.  I've made this change so that we can choose to update multiple entities in a single transaction now.

app/editor/src/components/form/options/interfaces/OptionItem.ts
```diff
@@ -24,11 +24,10 @@ export class OptionItem<T extends string | number | undefined = string | number
   }
 
   toInterface(): IOptionItem<T> {
-    return {
-      discriminator: this.discriminator,
-      label: this.label,
-      value: this.value,
-      isEnabled: this.isEnabled,
-    };
+    return this as IOptionItem<T>;
```
Fosol, line 27: No need to create a new object as it is the same shape.

app/editor/src/components/form/select/Select.tsx
```diff
@@ -23,7 +23,7 @@ export interface ISelectBaseProps {
   /**
    * The styled variant.
    */
-  variant?: SelectVariant;
+  variant?: SelectVariant | string;
```
Fosol, line 26: Gives us more flexibility.

app/editor/src/components/form/select/Select.tsx
```diff
@@ -104,7 +123,13 @@ export const Select = <OptionType extends IOptionItem>({
           {label} {tooltip && <FontAwesomeIcon icon={faInfoCircle} />}
         </label>
       )}
-      <Row className="select-container">
+      <Row
```
Fosol, line 126: This is to replace the advanced search Enter keypress to cause a search.

app/editor/src/components/form/select/styled/Select.tsx
```diff
@@ -2,15 +2,19 @@ import styled from 'styled-components';
...
     color: ${(props) => props.theme.css.dangerColor};
   }
-  margin-right: 0.5em;
+
   .select-container {
     width: 100%;
   }
+
   .input-container {
+    width: 100%;
```
Fosol, line 17: @jdtoombs This is to make the select component expand within its space.

PR #415, opened 2022-11-24
Add manual page length selector to PagedTable and increase the default list length in UserList to 20

sturufous:


libs/npm/core/src/components/grid-table/PagedTable.tsx
```diff
@@ -21,6 +21,7 @@ export const PagedTable = <CT extends object = Record<string, unknown>>({
   return (
     <GridTable
       data={page.items}
+      manualPageSize
```
Fosol, line 24: Don't do this.  It will always be true.  Remove this property as it will already be passed down if the implementation requires it through the `{...rest}`

Fosol (changes requested):

libs/npm/core/src/components/grid-table/PagedTable.tsx
sturufous: Will get to this on Monday.

Fosol (changes requested):
Remove extra zip file from commit.

Fosol (approved):

PR #416, opened 2022-11-25
TNO-764, TNO-765: Select Field Tweaks

jdtoombs:
- slight change to text field on content form
- able to clear some previously un clearable fields (via x button)

### DIFF
--- a/app/editor/src/features/content/form/ContentSummaryForm.tsx
+++ b/app/editor/src/features/content/form/ContentSummaryForm.tsx
@@ -165,7 +165,6 @@ export const ContentSummaryForm: React.FC<IContentSummaryFormProps> = ({
                 label="Event of Day Category"
                 width={FieldSize.Medium}
                 options={categoryOptions}
-                clearValue={[]}
                 value={
                   values.categories.length
                     ? categoryOptions.find((c) => c.value === values.categories[0].id) ?? []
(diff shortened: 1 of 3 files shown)


app/editor/src/features/content/form/ContentSummaryForm.tsx
```diff
@@ -165,15 +165,17 @@ export const ContentSummaryForm: React.FC<IContentSummaryFormProps> = ({
...
                   values.categories.length
                     ? categoryOptions.find((c) => c.value === values.categories[0].id) ?? []
                     : []
                 }
                 onChange={(e: any) => {
                   // only supports one at a time right now
-                  const value = categories.find((c) => c.id === e.value);
+                  let value;
+                  if (!!e?.value) {
+                    value = categories.find((c) => c.id === e.value) ?? [];
```
Fosol, line 177: This is messy I can't understand what's going on.
The following should return an array with one item, or an empty array.
```
const value = categories.filter((c) => c.id === e.value);
setFieldValue('categories', value);
```

Fosol (changes requested):

app/editor/src/features/content/form/ContentSummaryForm.tsx
jdtoombs: The above throws an error when clearing the field, will try to clean it up though!

Fosol (approved):

PR #418, opened 2022-11-29
New toolbar and layout

jdtoombs:
- components not moved to tno-core yet as some tweaks may be needed in terms of props etc with next sprints stories
- new vertical stack layout instead of horizontal panes
- currently clicking the calendar does nothing as that is to be implemented next sprint
image


app/editor/src/features/content/tool-bar/ContentToolBar.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * Uses the ToolBar component to create a toolbar specific to the content filter
+ * @param onSearch Inform the parent of the request to search.
+ * @returns A content filter toolbar
+ */
+export const ContentToolBar: React.FC<IContentToolBarProps> = ({ onSearch }) => {
+  // eslint-disable-next-line no-empty-pattern
+  const [{}, { storeFilter, storeFilterAdvanced }] = useContent();
```
jdtoombs, line 25: linting failed with empty object

app/editor/src/components/form/formpage/FormPage.tsx
```diff
@@ -7,6 +7,8 @@ export interface IFormPageProps extends InputHTMLAttributes<HTMLDivElement> {
   minWidth?: string;
   /** bypass the default max width */
   maxWidth?: string;
+  /** remove padding from the FormPage */
+  noPadding?: boolean;
```
Fosol, line 11: Every time we add component properties for UI layout we're doing it wrong.  CSS specificity should be applied to override default behaviour as much as possible.

app/editor/src/components/toggle-group/ToggleGroup.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import * as styled from './styled';
+
+type IToggleOption = {
+  label: string;
+  onClick?: () => void;
+};
+
+export interface IToggleGroupProps {
+  options?: IToggleOption[];
```
Fosol, line 11: Would you ever use this without options?

app/editor/src/components/toggle-group/styled/ToggleGroup.tsx
```diff
@@ -0,0 +1,34 @@
+import styled from 'styled-components';
+import { IRowProps, Row } from 'tno-core';
+
+export interface IStyledToggleGroupProps extends IRowProps {
```
Fosol, line 4: Don't put an interface in this file.  Keep it consistent with other styled files.

app/editor/src/components/toggle-group/ToggleGroup.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+
+import * as styled from './styled';
+
+type IToggleOption = {
```
Fosol, line 5: Why is this a `type` not an `interface`?

app/editor/src/components/tool-bar/ToolBarSection.tsx
```diff
@@ -0,0 +1,27 @@
+import { Row } from 'tno-core';
+
+import * as styled from './styled';
+
+export interface IToolBarSectionProps {
+  actions: React.ReactNode;
```
Fosol, line 6: If I'm not mistaken, this should follow common standards and be named `children`

app/editor/src/components/tool-bar/styled/ToolBarSection.tsx
```diff
@@ -0,0 +1,51 @@
...
+    .action-icon {
+      align-self: center;
+      padding-left: 0.5em;
+      padding-right: 0.25em;
+      color: #003366;
+      height: 100%;
+      width: 1.25em;
+      cursor: pointer;
+    }
+    background-color: #f2f2f2;
```
Fosol, line 37: It would make it a lot easier to change all our theme colours if we kept them as theme variables instead of hardcoding them.

app/editor/src/features/content/tool-bar/ContentToolBar.tsx
Fosol: Don't use an empty object.  `[,{ storeFilter, storeFilterAdvanced }]`

app/editor/src/features/content/tool-bar/sections/CreateNewSection.tsx
```diff
@@ -0,0 +1,26 @@
...
+import * as styled from './styled';
+
+/**
+ * Section containing the create new content buttons
+ * @returns Section with three separate create content buttons
+ */
+export const CreateNewSection: React.FC = () => {
+  const navigate = useNavigate();
+  return (
+    <styled.CreateNewSection
```
Fosol, line 14: I don't see a lot of value of these components being wrapped in a styled component, when the styled component is actually the component they're being wrapped in `ToolBarSection`.  It's confusing and presently there is no reason to do it.  Just wrap these components in the `ToolBarSection`.

app/editor/src/features/content/tool-bar/sections/FilterContentSection.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const [productOptions, setProductOptions] = React.useState<IOptionItem[]>([]);
+
+  React.useEffect(() => {
+    setProductOptions([new OptionItem<number>('Any', 0), ...pOptions]);
+  }, [pOptions]);
+
+  const [{ userInfo }] = useApp();
+
+  return (
+    <styled.FilterContentSection
```
Fosol, line 33: Just wrap these components in the `ToolBarSection`.

Fosol (changes requested):

app/editor/src/features/content/tool-bar/ContentToolBar.tsx
jdtoombs: D'oh

app/editor/src/components/toggle-group/ToggleGroup.tsx
jdtoombs: Nope, forgot to remove that. Was for the compiler at one point

app/editor/src/components/form/formpage/FormPage.tsx
jdtoombs: Just need some clarification on the above @Fosol

app/editor/src/components/toggle-group/styled/ToggleGroup.tsx
jdtoombs: Don't even need that prop anymore :)

jdtoombs:
@Fosol, changes made! And can be found [here ](https://github.com/bcgov/tno/pull/418/commits/756c18c37d018db2d2bef5f59f62764a8a9863e3)

app/editor/src/components/form/formpage/FormPage.tsx
Fosol: When we add a property in the component to control layout like this it makes it impossible for theme overrides based on css.  A component should have a standard look and feel.  If the implementation wants a different look, then they should just be able to add css to their solution and it will override the default.  You can do this through class name.  We can have a conversation about it today when you have time.

Fosol (approved):

app/editor/src/components/form/formpage/FormPage.tsx
jdtoombs: Gotcha! Makes sense now

PR #420, opened 2022-11-30
TNO-783: Infinite scroll

jdtoombs:
Ended up not using an external library. Could achieve the same without. Not the smoothest transition - but open to suggestions

infinite-scroll


Fosol (changes requested):

jdtoombs:
@Fosol - updated; but will wait until your PR is merged before publishing the package!

libs/npm/core/src/components/grid-table/PagedTable.tsx
```diff
@@ -22,8 +22,15 @@ export const PagedTable = <CT extends object = Record<string, unknown>>({
...
   React.useEffect(() => {
-    infiniteScroll && setItems(items.concat(page.items));
+    if (addToInfiniteItems) {
+      setItems((prevItems) => prevItems.concat(page.items));
+      setAddToInfiniteItems(false);
+    } else {
+      // we don't want stale items when user changes sorting or filter from something other than scrolling
+      setItems(page.items);
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
jdtoombs, line 34: Spent a couple hours on the below

libs/npm/core/src/components/grid-table/GridTable.tsx
```diff
@@ -312,23 +338,41 @@ export const GridTable = <T extends object>({
...
-              {row.cells.map((cell) => {
-                return (
-                  <div {...cell.getCellProps()}>
-                    {cell.render('Cell') as unknown as React.ReactNode}
-                  </div>
-                );
-              })}
-            </div>
-          );
+          if (page.length === index + 1 && infiniteScroll) {
```
Fosol, line 343: Is there a way to write the two outputs?  They are identical except the `ref={lastRowRef}`.  It makes it difficult to know what's really different and then will introduce maintenance issues down the road.

Fosol (changes requested):
You'll need to merge and publish a new version of `tno-core`.

app/editor/yarn.lock
```diff
@@ -13913,9 +13913,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"tno-core@npm:0.0.193":
-  version: 0.0.193
-  resolution: "tno-core@npm:0.0.193"
+"tno-core@npm:0.0.196":
```
Fosol, line 13916: Something doesn't look right here?

libs/npm/core/package.json
```diff
@@ -39,7 +39,7 @@
     "react-toastify": "^9.1.1",
     "react-tooltip": "^4.5.0",
     "styled-components": "^5.3.6",
-    "tno-core": "0.0.193"
+    "tno-core": "0.0.196"
```
Fosol, line 42: Something is wrong here.

libs/npm/core/src/components/grid-table/PagedTable.tsx
Fosol: 99% of the time it means we're doing something wrong :(  But sometimes I have to do it too lol

Fosol (changes requested):

libs/npm/core/src/components/grid-table/PagedTable.tsx
jdtoombs: Lol one day we will get it.

I found out one good thing from the process at least. Often times if we are doing something like 
`setState({ ...state, someAttribute: x })` this could lead to problems with the dependencies but we can access the previous state with `setState((previousState) => { ...previousState, someAttribute: x })`

Fosol (approved):

PR #421, opened 2022-11-30
Clearable dropdown bugs

jdtoombs:
Same look and feel, just less bugs 🤦‍♂️


app/editor/src/components/form/select/Select.tsx
```diff
@@ -149,6 +138,7 @@ export const Select = <OptionType extends IOptionItem>({
             required={required}
             width={width}
             value={value}
+            isClearable
```
jdtoombs, line 141: Found this while working on another ticket 🙃 - fixes the CSS issues since we no longer have to maintain it

Fosol (approved):

PR #424, opened 2022-12-01
TNO-8 Fix work orders controller

Fosol:
Fixing bug introduced with Keycloak Gold realm integration.

### DIFF
--- a/api/net/Areas/Admin/Controllers/WorkOrderController.cs
+++ b/api/net/Areas/Admin/Controllers/WorkOrderController.cs
@@ -32,7 +32,6 @@ public class WorkOrderController : ControllerBase
 {
     #region Variables
     private readonly IWorkOrderService _workOrderService;
-    private readonly IKeycloakHelper _keycloakHelper;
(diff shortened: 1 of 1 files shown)


api/net/Areas/Admin/Controllers/WorkOrderController.cs
```diff
@@ -41,12 +40,10 @@ public class WorkOrderController : ControllerBase
     /// Creates a new instance of a WorkOrderController object, initializes with specified parameters.
     /// </summary>
     /// <param name="workOrderService"></param>
-    /// <param name="keycloakHelper"></param>
     /// <param name="hub"></param>
-    public WorkOrderController(IWorkOrderService workOrderService, IKeycloakHelper keycloakHelper, IHubContext<WorkOrderHub> hub)
+    public WorkOrderController(IWorkOrderService workOrderService, IHubContext<WorkOrderHub> hub)
```
Fosol, line 44: We no longer have direct access to Keycloak anymore.

PR #425, opened 2022-12-01
TNO-8 Reduce security to temporarily enable keycloak gold realm

Fosol:
This is a temporary "fix" to get everything working with the Keycloak Gold realm integration.

Regrettably Keycloak Gold realm does not provide a way to grant roles to the service account.  All of our services use this account to communicate with our API.  Because they have no roles, they are not authorized to use our API.  

I have removed most of the security related to authorization in the API to allow everything to keep working for now.

### DIFF
--- a/api/net/Areas/Editor/Controllers/ContentController.cs
+++ b/api/net/Areas/Editor/Controllers/ContentController.cs
@@ -16,7 +16,6 @@ using TNO.Core.Extensions;
 using TNO.Kafka;
 using TNO.API.Config;
 using TNO.Kafka.Models;
-using TNO.Keycloak;
 using System.Net.Mime;
 
 namespace TNO.API.Areas.Editor.Controllers;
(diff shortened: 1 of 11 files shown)


api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -16,15 +16,15 @@
...
 using System.Net.Mime;
 
 namespace TNO.API.Areas.Editor.Controllers;
 
 /// <summary>
 /// ContentController class, provides Content endpoints for the api.
 /// </summary>
-[ClientRoleAuthorize(ClientRole.Editor)]
+// [ClientRoleAuthorize(ClientRole.Editor)]
+[Authorize]
```
Fosol, line 27: If a user is authenticated we allow them access to these endpoints.  Not ideal, but presently a workaround to get everything again.  I'll create a custom authorization tag for the service account when I have time.

PR #427, opened 2022-12-01
TNO-805 Integrate services with keycloak gold realm

Fosol:
The services must request a token from a different issuer and audience.  Which means the API needs to authorize multiple issuers and audiences.

### DIFF
--- a/api/net/Program.cs
+++ b/api/net/Program.cs
@@ -17,7 +17,6 @@ using Microsoft.IdentityModel.Logging;
 using Microsoft.OpenApi.Models;
 using Swashbuckle.AspNetCore.SwaggerGen;
 using Prometheus;
-using TNO.API.Keycloak;
 using TNO.API.Middleware;
 using TNO.Core.Converters;
 using TNO.Core.Http;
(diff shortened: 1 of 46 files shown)


api/net/Program.cs
```diff
@@ -112,9 +111,11 @@
         options.TokenValidationParameters = new TokenValidationParameters()
         {
             ValidateIssuerSigningKey = true,
-            ValidIssuer = section.GetValue<string>("Issuer"),
+            // ValidIssuer = section.GetValue<string>("Issuer"),
+            ValidIssuers = section.GetValue<string>("Issuer")?.Split(",") ?? Array.Empty<string>(),
```
Fosol, line 115: Multiple are allowed now.

api/net/appsettings.Development.json
```diff
@@ -7,7 +7,9 @@
     }
   },
   "Keycloak": {
-    "Authority": "https://dev.loginproxy.gov.bc.ca/auth/realms/standard"
+    "Authority": "https://dev.loginproxy.gov.bc.ca/auth/realms/standard",
+    "Audience": "media-monitoring-mmia-3671,mmia-service-account-3994",
```
Fosol, line 11: One for the app and one for the services

PR #432, opened 2022-12-05
TNO-792 Fix content service

Fosol:
Presently all services will go to sleep after a configured number of sequential failures.  While this helps the services keep running on random failures, it still results in the services sleeping more often than not.  This is because certain failures are more likely to occur sequentially and within a short period of time.

All services will now make a configured number of retries after a request to the API or authentication fails.  This will allow intermittent failures to be recovered from.  It will also delay the subsequent retry request to allow the endpoint to possibly recover  from whatever is having issues.

Lastly ingest services now can also ignore failures altogether and returned a default value.  This will be used by the Capture service presently to default back to the current ingest configuration.  This means the Capture service will keep running with the last known ingest configuration when for some reason the API stops responding.

Published `tno-core:0.0.198` which allows customizing the date format in the `Date` component.

Updated Ingest list view columns to help quickly identify problematic ingest services.

image


libs/net/services/Actions/Managers/IngestActionManager`.cs
```diff
@@ -54,7 +55,7 @@ public IngestActionManager(IngestModel ingest, IApiService api, IIngestAction<TO
     protected override async Task<bool> PreRunAsync()
     {
         // Always fetch the latest version of the ingest and update this manager with it.
-        var ingest = await _api.GetIngestAsync(this.Ingest.Id);
+        var ingest = await this.Api.HandleRequestFailure(async () => await this.Api.GetIngestAsync(this.Ingest.Id), this.Options.ReuseIngests, this.Ingest);
```
Fosol, line 58: Default to current ingest value.  This may cause some unexpected issues if the first attempt to fetch the ingest fails.

PR #434, opened 2022-12-05
TNO-792 Fix ingest services

Fosol:
Adding the additional error handling regrettably introduced a thread issue.  Fixed this, and also a logic error.

### DIFF
--- a/libs/net/core/Http/HttpRequestClient.cs
+++ b/libs/net/core/Http/HttpRequestClient.cs
@@ -16,7 +16,7 @@ namespace TNO.Core.Http
     public class HttpRequestClient : IHttpRequestClient
     {
         #region Variables
-        private readonly JsonSerializerOptions _serializeOptions;
+        protected readonly JsonSerializerOptions _serializeOptions;
         private readonly ILogger<HttpRequestClient> _logger;
         #endregion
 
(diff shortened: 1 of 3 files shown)


libs/net/core/Http/HttpRequestClient.cs
```diff
@@ -16,7 +16,7 @@ namespace TNO.Core.Http
     public class HttpRequestClient : IHttpRequestClient
     {
         #region Variables
-        private readonly JsonSerializerOptions _serializeOptions;
+        protected readonly JsonSerializerOptions _serializeOptions;
```
Fosol, line 19: Grant these options to all inherited classes.

libs/net/core/Http/OpenIdConnectRequestClient.cs
```diff
@@ -70,8 +70,8 @@ public class OpenIdConnectRequestClient : HttpRequestClient, IOpenIdConnectReque
 
             if (response.IsSuccessStatusCode)
             {
-                using var responseStream = await response.Content.ReadAsStreamAsync();
-                return await responseStream.DeserializeAsync<Models.OpenIdConnectModel>();
+                var body = await response.Content.ReadAsStringAsync();
```
Fosol, line 73: The stream is not accessible in other threads which was causing errors.

libs/net/services/Helpers/ApiService.cs
```diff
@@ -113,7 +113,7 @@ protected async Task<T> RetryRequestAsync<T>(Func<Task<T>> callbackDelegate)
         }
         catch (Exception ex)
         {
-            if (this.Options.RetryLimit >= ++this.FailureCount)
+            if (this.Options.RetryLimit <= ++this.FailureCount)
```
Fosol, line 116: Fixed logic error.

PR #436, opened 2022-12-05
Fix styling bug for WYSIWYG

jdtoombs:
When there were two instances of a WYSIWYG present (when expanding to full view) it caused issues for the styling. Had to provide a ref for each instance of the editor. 

This will unblock Cello :)

### DIFF
--- a/app/editor/src/components/form/wysiwyg/CustomToolbar.tsx
+++ b/app/editor/src/components/form/wysiwyg/CustomToolbar.tsx
@@ -1,11 +1,11 @@
 import { FaCode, FaExpandAlt, FaPaintBrush, FaRemoveFormat } from 'react-icons/fa';
 import { TiDelete } from 'react-icons/ti';
-
 export interface ICustomToolbarProps {
   onClickRaw: React.MouseEventHandler<HTMLButtonElement>;
   onClickRemoveFormat: React.MouseEventHandler<HTMLButtonElement>;
   onClickFormatRaw: React.MouseEventHandler<HTMLButtonElement>;
   onClickClear: React.MouseEventHandler<HTMLButtonElement>;
+  innerRef: any;
 }
 
 /** Custom toolbar for the content WYSIWYG including ability to view HTML source and clear formatting */
@@ -14,8 +14,9 @@ export const CustomToolbar: React.FC<ICustomToolbarProps> = ({
   onClickRemoveFormat,
   onClickFormatRaw,
   onClickClear,
+  innerRef,
 }) => (
-  <div id="toolbar" className="toolbar">
+  <div ref={innerRef} className="toolbar">
     <span className="ql-formats">
       <select className="ql-header" />
       <button className="ql-bold" />
--- a/app/editor/src/components/form/wysiwyg/Wysiwyg.tsx
+++ b/app/editor/src/components/form/wysiwyg/Wysiwyg.tsx
@@ -1,3 +1,5 @@
+import 'react-quill/dist/quill.snow.css';
+
 import { useFormikContext } from 'formik';
 import { IContentModel } from 'hooks';
 import { html_beautify } from 'js-beautify';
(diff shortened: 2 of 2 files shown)


app/editor/src/components/form/wysiwyg/CustomToolbar.tsx
```diff
@@ -14,8 +14,9 @@ export const CustomToolbar: React.FC<ICustomToolbarProps> = ({
   onClickRemoveFormat,
   onClickFormatRaw,
   onClickClear,
+  innerRef,
 }) => (
-  <div id="toolbar" className="toolbar">
```
jdtoombs, line 18: Multiple instances of this id would break the styling

app/editor/src/components/form/wysiwyg/CustomToolbar.tsx
```diff
@@ -1,11 +1,11 @@
 import { FaCode, FaExpandAlt, FaPaintBrush, FaRemoveFormat } from 'react-icons/fa';
 import { TiDelete } from 'react-icons/ti';
-
 export interface ICustomToolbarProps {
   onClickRaw: React.MouseEventHandler<HTMLButtonElement>;
   onClickRemoveFormat: React.MouseEventHandler<HTMLButtonElement>;
   onClickFormatRaw: React.MouseEventHandler<HTMLButtonElement>;
   onClickClear: React.MouseEventHandler<HTMLButtonElement>;
+  innerRef: any;
```
jdtoombs, line 8: Trying to get this type

app/editor/src/components/form/wysiwyg/CustomToolbar.tsx
```diff
@@ -14,8 +14,9 @@ export const CustomToolbar: React.FC<ICustomToolbarProps> = ({
   onClickRemoveFormat,
   onClickFormatRaw,
   onClickClear,
+  innerRef,
 }) => (
-  <div id="toolbar" className="toolbar">
+  <div ref={innerRef} className="toolbar">
```
Fosol, line 19: I'm no expert but this looks like it may be related to ForwardRef.  https://reactjs.org/docs/forwarding-refs.html
I don't fully understand it all, but you best read it over and make sure you're doing this correct.

app/editor/src/components/form/wysiwyg/Wysiwyg.tsx
```diff
@@ -117,29 +123,34 @@ export const Wysiwyg: React.FC<IWysiwygProps> = ({ fieldName, label, required })
         onClickRemoveFormat={stripHtml}
         onClickFormatRaw={onClickFormatRaw}
         onClickClear={() => setState({ ...state, html: '', rawHtml: '' })}
+        innerRef={setToolBarNode}
```
Fosol, line 126: I really don't understand this situation of using react state for a ref?

app/editor/src/components/form/wysiwyg/CustomToolbar.tsx
jdtoombs: Sounds good, taking a look!

app/editor/src/components/form/wysiwyg/Wysiwyg.tsx
jdtoombs: It's kind of an odd instance, and a limitation of react-quill and the way we want to use it. But pretty much this will set the toolbar container to a unique instance of a `toolBarNode`.  Which can be found on line 94:
`    const config = {
      toolbar: {
        container: toolBarNode,
      },
    };`

jdtoombs: Experimenting with forward ref and looking for other solutions still, yet to find one

jdtoombs: Tried passing as setState instead of a ref - but no luck. Think this is one of those instances where we are fighting the library 
:(

jdtoombs: But the above is supported by React `Instead of passing a ref attribute created by createRef(), you pass a function. The function receives the React component instance or HTML DOM element as its argument, which can be stored and accessed elsewhere.` - from the docs

Fosol (approved):

PR #439, opened 2022-12-07
TNO-697 Update SignalR

Fosol:
SignalR was not working as intended in Openshift.  A user now must be authenticated before connecting to our communication hub.

If a work order status changes a message will be sent out to all users.  Depending on their current URL they will reload the related content and update their UI.

- Content Form tabs will show status icon
- Content List View will show if the content has a body (aka transcript)

> You will need to refresh the API  `make refresh n=api` 

image


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -152,42 +124,37 @@ export const ContentForm: React.FC<IContentFormProps> = ({
...
 
+  const onWorkOrder = React.useCallback(
+    (workOrder: IWorkOrderModel) => {
+      if (!!id && +id === workOrder.contentId) fetchContent(workOrder.contentId);
+    },
+    [fetchContent, id],
+  );
+
+  React.useEffect(() => {
+    return hub.listen(HubMethodName.WorkOrder, onWorkOrder);
```
Fosol, line 143: Nice and simple way to start listening to events.

app/editor/src/hooks/api-editor/constants/HubMethodName.ts
```diff
@@ -0,0 +1,3 @@
+export enum HubMethodName {
+  WorkOrder = 'WorkOrder',
```
Fosol, line 2: To maintain strongly typed method names we keep them here.

app/editor/src/hooks/api-editor/editor/useApiHub.ts
```diff
@@ -0,0 +1,91 @@
...
+  LogLevel,
+} from '@microsoft/signalr';
+import React from 'react';
+import { useAppStore } from 'store/slices';
+import { useKeycloakWrapper } from 'tno-core';
+
+import { HubMethodName, Settings } from '../constants';
+
+const url = Settings.ApiPath + '/hub';
+let hub: HubConnection;
```
Fosol, line 14: Use a module singleton variable so that it's one connection for the whole application.

app/editor/src/hooks/api-editor/editor/useApiHub.ts
```diff
@@ -0,0 +1,91 @@
...
+        });
+    }
+  }, [app, auth]);
+
+  return {
+    state,
+    start: () => hub.start(),
+    on: (methodName: string, newMethod: (...args: any[]) => void) => hub.on(methodName, newMethod),
+    off: (methodName: string, method: (...args: any[]) => void) => hub.off(methodName, method),
+    stop: () => hub.stop(),
```
Fosol, line 81: At some point we should stop the connection when the user logs out.

PR #440, opened 2022-12-07
TNO-792 Fix Content service

Fosol:
The content service was having issues restarting after going to sleep.  

The reason the content service fails regrettably is due to a known challenge with deployments.  We still need a way to tell all our services to go to sleep during a deployment of the API.


libs/net/kafka/KafkaListener`.cs
```diff
@@ -304,6 +304,7 @@ public void Stop()
         if (_open)
         {
             _logger.LogInformation("Consumer is stopping");
+            this.Topics = Array.Empty<string>();
```
Fosol, line 307: This fixes the resume bug.

PR #441, opened 2022-12-07
TNO-815 Fix keycloak gold

Fosol:
A first time user login will now perform the following steps to activate their account.

1) Check if a user with the same key exists (should only occur if they have already been activated)
2) NEW - Check if a user with the same username exists in the database
3) Check if a user with the same email exists in the database

This is to solve an issue where the preapproved user email doesn't match the username.  When the user logs in the first time, it attempts to create a new account, but the username already exists.

### DIFF
--- a/api/net/Controllers/AuthController.cs
+++ b/api/net/Controllers/AuthController.cs
@@ -54,7 +54,7 @@ public class AuthController : ControllerBase
     [HttpPost("userinfo")]
     [Produces(MediaTypeNames.Application.Json)]
     [ProducesResponseType(typeof(PrincipalModel), 200)]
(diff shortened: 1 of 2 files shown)


api/net/Controllers/AuthController.cs
```diff
@@ -54,7 +54,7 @@ public AuthController(ICssHelper cssHelper, IUserService userService)
     [HttpPost("userinfo")]
     [Produces(MediaTypeNames.Application.Json)]
     [ProducesResponseType(typeof(PrincipalModel), 200)]
-    [SwaggerOperation(Tags = new[] { "health" })]
+    [SwaggerOperation(Tags = new[] { "Auth" })]
```
Fosol, line 57: Reorganize

PR #443, opened 2022-12-08
TNO-815 Fix Keycloak gold

Fosol:
The integration with Keycloak Gold realm had a number of bugs.

Published new `tno-core:0.0.199` to fix `useKeycloakWrapper()` hook.

- Preapproved user was not being correctly updated when a user logged in the first time
- The `User.Key` was not always being correctly linked to Keycloak
- After activation React state was not updating keycloak claims


api/net/Keycloak/CssHelper.cs
```diff
@@ -203,7 +203,7 @@ private async Task AddOrUpdateUserAsync(string username, Entities.User? user, TN
     public async Task<string[]> UpdateUserRolesAsync(string username, string[] roles)
     {
         var userRoles = await _cssService.GetRolesForUserAsync(username);
-        if (userRoles.Users.Length == 0) return Array.Empty<string>();
+        if (userRoles.Users.Length == 0) userRoles = new UserRoleResponseModel() { };
```
Fosol, line 206: This resolve an issue where roles were not being applied to a new user.  Keycloak Gold realm doesn't return anything if the user hasn't already been assigned a role.  Which is annoying and makes adding new users more difficult.

app/editor/src/features/access-request/AccessRequest.tsx
```diff
@@ -19,7 +23,7 @@ export const AccessRequest: React.FC = (props) => {
   React.useEffect(() => {
     // The user has been approved, redirect back to home page.
     if (keycloak.hasClaim() && location.pathname === '/welcome') navigate('/');
-  });
+  }, [keycloak, location.pathname, navigate]);
```
Fosol, line 26: This helped fix state issue after the user was activated.

app/editor/src/features/admin/users/UserForm.tsx
```diff
@@ -194,13 +194,10 @@ export const UserForm: React.FC = () => {
...
               confirmText="Yes, Remove It"
               onConfirm={async () => {
-                try {
-                  await api.deleteUser(user);
-                  toast.success(`${user.username} has successfully been deleted.`);
-                  navigate('/admin/users');
-                } finally {
-                  toggle();
-                }
+                toggle();
```
Fosol, line 197: Toggle right away for better user experience.

libs/net/dal/Services/UserService.cs
```diff
@@ -123,6 +123,7 @@ public override User UpdateAndSave(User entity)
 
         if (original != null)
         {
+            original.Key = entity.Key;
```
Fosol, line 126: Update key

libs/net/dal/Services/UserService.cs
```diff
@@ -135,6 +136,7 @@ public override User UpdateAndSave(User entity)
             original.Note = entity.Note;
             original.Code = entity.Code;
             original.Roles = entity.Roles;
+            original.LastLoginOn = entity.LastLoginOn;
```
Fosol, line 139: Update last login.

libs/npm/core/src/hooks/keycloak/useKeycloakWrapper.ts
```diff
@@ -13,11 +13,12 @@ export function useKeycloakWrapper(): IKeycloak {
 
   const [initialized, setInitialized] = React.useState(ready);
 
+  if (ready) token = keycloak.tokenParsed as IKeycloakToken;
```
Fosol, line 16: Fix state issue with token value.  The token is stored as a global module level so that state doesn't get refreshed all the time for each instance of the hook.

PR #444, opened 2022-12-08
TNO-784: add toolbar component

AlessiaYChen:


app/editor/src/components/tool-bar/ToolBarSection2.tsx
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Creates a section for the parent ToolBar component
+ * @param actions Provide buttons/ filter actions for given section
+ * @param label Provide a label for the section
+ * @returns A ToolBar section
+ */
+export const ToolBarSection2: React.FC<IToolBarSection2Props> = ({ label }) => {
+  return (
+    <styled.ToolBarSection className="details-section">
+      <Row className="title-container">Content Details</Row>
```
jdtoombs, line 16: Could maybe make this a prop (`title: string`) so we can reuse this section for other areas!

app/editor/src/components/tool-bar/ToolBarSection2.tsx
```diff
@@ -0,0 +1,21 @@
+import { Row } from 'tno-core';
```
jdtoombs, line 1: Could maybe rename this component to `TitleToolBarSection.tsx`

app/editor/src/components/tool-bar/ToolBarSection2.tsx
```diff
@@ -0,0 +1,21 @@
...
+import * as styled from './styled';
+
+export interface IToolBarSection2Props {
+  label: string;
```
jdtoombs, line 5: Maybe make this optional so that we can use it as a title section in other parts of the application if required. And then `title: string;` will be a required prop to determine the heading for the given section.

jdtoombs:
Looks good! We can probably reuse this section for other areas of the application at some point so adding title as a prop will make it a bit more generic for us. 

Also if you run `yarn lint:fix` from your `/app/editor` that should clear up the errors

app/editor/src/components/tool-bar/ToolBarSection2.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * Creates a section for the parent ToolBar component
+ * @param actions Provide buttons/ filter actions for given section
+ * @param label Provide a label for the section
+ * @returns A ToolBar section
+ */
+export const ToolBarSection2: React.FC<IToolBarSection2Props> = ({ label }) => {
+  return (
+    <styled.ToolBarSection className="details-section">
+      <Row className="title-container">Content Details</Row>
+      <Row className="status-container">{label}</Row>
```
jdtoombs, line 17: Eventually we could rename this class to something different in case we have a slightly different use case than a status to  pair with a title

app/editor/src/components/tool-bar/ToolBarSection2.tsx
jdtoombs, line 1: Maybe rename this to ContentStatusSection.tsx and move it to [tno](https://github.com/bcgov/tno)/[app](https://github.com/bcgov/tno/tree/dev/app)/[editor](https://github.com/bcgov/tno/tree/dev/app/editor)/[src](https://github.com/bcgov/tno/tree/dev/app/editor/src)/[features](https://github.com/bcgov/tno/tree/dev/app/editor/src/features)/[content](https://github.com/bcgov/tno/tree/dev/app/editor/src/features/content)/[tool-bar](https://github.com/bcgov/tno/tree/dev/app/editor/src/features/content/tool-bar)/sections/

app/editor/src/components/tool-bar/styled/ToolBarSection.tsx
```diff
@@ -48,4 +48,18 @@ export const ToolBarSection = styled(Col)`
     padding: 0.5em;
     justify-content: center;
   }
+  .title-container {
```
jdtoombs, line 51: Do you think this should be added to a separate styled component @Fosol? Or is that overkill - for example have it in `styled/ContentStatusSection.tsx` as it is specific to that component and not the general default `ToolBarSection`

app/editor/src/components/tool-bar/styled/ToolBar.tsx
```diff
@@ -14,4 +14,7 @@ export const ToolBar = styled(Row)`
   .section:last-child {
     border-right: none;
   }
+  .details-section {
+    background-color: #faf9f7;
```
jdtoombs, line 18: I will get to use this on my tickets :) Thanks!

jdtoombs:
Looks good! If you run `yarn lint:fix` at `/app/editor` it should take care of that linting error for you. Also for the css colours we have a file called `_colors.scss` that may contain what you are using if not you can add it!

To do so:

1. add variable to `_colors.scss`
2. Add that new variable to a variable export in `_variables.module.scss`
3. In the styled component use the variable like so `background-color: ${(props) => props.theme.css.variableNameHere};`

I am notorious for forgetting to use the variables - but as Jeremy always says it's much easier to maintain that way.

Can see line 37 in `stled/ToolBar.tsx` for an example on how to ingest that variable into the styled component.

app/editor/src/components/tool-bar/ToolBarSection2.tsx
Fosol: I don't think there is any value in having another generic component.  The current `ToolBarSection` component does most of this.  Just update it so that it can have an optional `icon` property.  Make a specific implementation for `ContentStatusSection` that just shows the status.  You can then override the styling locally to make the component look correct.

`/app/editor/src/features/content/tool-bar/ContentStatusSection.tsx`

```
export const ContentStatusSection: React.FC<IContentStatusSectionProps> = () => {
const { values } = useFormikContext<IContentForm>();
return (<styled.ContentStatusSection>
 <ToolBarSection label="Content Details">{values.status}</ToolBarSection/>
 </styled.ContentStatusSection>
 );
 }

app/editor/src/components/tool-bar/styled/ToolBarSection.tsx
Fosol: I think I covered this in my comment on the `ToolBarSection2.tsx` file.

Fosol (changes requested):

jdtoombs:
Closing as covered in https://github.com/bcgov/tno/pull/463

PR #470, opened 2023-01-09
TNO-781: New toning group

jdtoombs:
image

Updated toning group on the content form.


app/editor/src/components/form/toning/ToningGroup.tsx
```diff
@@ -0,0 +1,64 @@
...
+  fieldName: keyof IContentModel;
+}
+
+export const ToningGroup: React.FC<IToningGroupProps> = ({ fieldName }) => {
+  const toningOptions = [-5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5];
+  const { values, setFieldValue, errors, touched } = useFormikContext<IContentModel>();
+
+  const [active, setActive] = React.useState<number>();
+  React.useEffect(() => {
+    if (values.tonePools?.length && values.tonePools[0].value) setActive(values.tonePools[0].value);
```
Fosol, line 19: defaults to `-5`?  Is this what we want?

app/editor/src/components/form/toning/ToningGroup.tsx
```diff
@@ -0,0 +1,64 @@
...
+        {toningOptions.map((option) => (
+          <Col>
+            {determineIndicator(option)}
+            <button
+              className={active === option ? 'active' : ''}
+              type="button"
+              key={option}
+              onClick={() => {
+                setActive(option);
+                setFieldValue('tonePool', option);
```
Fosol, line 52: Why do we need to set two properties?

app/editor/src/components/form/toning/ToningGroup.tsx
jdtoombs: this is actually looking at the existing formik value of tonePools which the way we have it setup will always be tonePools[0]. it currently defaults to nothing when creating new content!

app/editor/src/components/form/toning/ToningGroup.tsx
jdtoombs: Can't remember the exact reasoning, but it was done this way in order to get the validation to work correctly

Fosol (approved):

PR #471, opened 2023-01-09
TNO-850 Setup Subscriber App

Fosol:
Setup the Subscriber application so that we can start doing development on it.

Moved most shared components to `tno-core` library.  Published new version of library `tno-core:0.0.203`.  There are components that need to be cleaned up and moved still, but not critical to this story.

This will add a new container for the subscriber app and it needs to be running for nginx to work.  I'd love to make it optional somehow.

The new URL is `http://localhost:4008`

After this is merged I'll try and find time to also move all the api hooks to the shared `tno-core` library.


jdtoombs (approved):
Thanks for doing that, probably not the most fun thing 😝

PR #477, opened 2023-01-13
TNO-626: User table size selection

jdtoombs:
- users will now have their page size selection persists within the pager
- updated package 0.0.203 -> 0.0.205 (made an accidental change in 0.0.204)


libs/npm/core/src/components/grid-table/Pager.tsx
```diff
@@ -65,6 +66,29 @@ export const Pager: React.FC<IPagerProps> = ({
...
+  let [searchParams, setSearchParams] = useSearchParams(query);
+  const qtyCookie = document.cookie
+    .split('; ')
+    .find((row) => row.startsWith('qty='))
+    ?.split('=')[1];
+
+  // update cookie when qty changes
+  React.useEffect(() => {
+    if (qtyCookie !== searchParams.get('qty') && searchParams.get('qty') !== null)
+      document.cookie = `qty=${searchParams.get('qty')}; expires=Fri, 31 Dec 9999 23:59:59 GMT`;
```
Fosol, line 80: Do you need to hardcode the date?

jdtoombs: For cookies we can either set a time limit or a far future date. By default it ends after the user closes their browser unfortunately

Fosol (approved):

PR #480, opened 2023-01-13
Adding ffprobe command for in-progress mkv capture duration

sturufous:
Also addressed the issue where the default ffprobe command returns a float in string format but the mkv command returns an integer. Effectively divided the mkv duration string value by 1000 by inserting a decimal point, making the duration represent seconds rather than milliseconds. I couldn't find a way to control the output format of the duration in ffprobe. This also ensures that the duration for all formats can be converted using float.TryParse().

### DIFF
--- a/services/net/clip/ClipAction.cs
+++ b/services/net/clip/ClipAction.cs
@@ -385,26 +385,23 @@ public class ClipAction : CommandAction<ClipOptions>
         switch (ext)
         {
             case ".mkv":
-                var tempFile = await RepackageFileAsync(inputFile);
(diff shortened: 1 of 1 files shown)


services/net/clip/ClipAction.cs
```diff
@@ -385,26 +385,23 @@ private static async Task<long> GetDurationAsync(string inputFile)
         switch (ext)
```
Fosol, line 385: Move logic to `ParseDurationAsync(...)` function.  As it's specifically related to parsing the duration.

Fosol (changes requested):

PR #484, opened 2023-01-16
TNO-854: Bold tabs

jdtoombs:
Quick PR to bold the label text within our Tab component


Fosol (approved):
The downside to minor updates to a shared library lol.  I wonder if we could come up with a way to reduce this if we had way more time to do everything?

Fosol:
@jdtoombs Looks like you didn't do a proper merge and `yarn install`.  You'll need to resolve this, and also you'll need to update the subscriber package.json to use the new version of the `tno-core` library.

jdtoombs:
@Fosol - forgot to pull down latest before my changes :( fixing now

jdtoombs:
> The downside to minor updates to a shared library lol. I wonder if we could come up with a way to reduce this if we had way more time to do everything?

Lol yes, that would be ideal. Could do some brainstorming!

jdtoombs:
Screwed up the branch

PR #490, opened 2023-01-18
TNO-906 Fix timezone

Fosol:
Applying Stuart's timezone fix in separate PR to avoid DB changes for now.

Resolved a number of build warnings.


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -755,21 +755,42 @@ private static List<Author> GetAuthorList(string author)
...
-            return dateTime.ToUniversalTime();
+            return LocalizeForIngestTimezone(dateTime, ingest, options);
         }
         else
         {
             return DateTime.Now;
         }
     }
 
+    /// <summary>
```
Fosol, line 774: Copied from Stuart's PR to address timezone issue with daily papers ingest.

PR #493, opened 2023-01-18
TNO-906 Fix duplicates

Fosol:
The File Service was posting duplicate content to Kafka when the published date was older.  I am not sure why but the daily papers appear to have a lot of duplicate content.


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -240,7 +240,7 @@ private async Task ImportArticleAsync(IIngestServiceActionManager manager, Sourc
                 reference = await AddContentReferenceAsync(manager.Ingest, content);
             }
             else if ((reference.Status == (int)WorkflowStatus.InProgress && reference.UpdatedOn?.AddMinutes(5) < DateTime.UtcNow) ||
-                    (reference.Status != (int)WorkflowStatus.InProgress && reference.PublishedOn != content.PublishedOn))
+                    (reference.Status != (int)WorkflowStatus.InProgress && reference.PublishedOn < content.PublishedOn))
```
Fosol, line 243: Only post content with a newer published date.  Technically this should be a new updated value, as the published date should not change once content has been published.  However, reviewing the source content we must do this.

PR #498, opened 2023-01-18
TNO-906 Improve ingest admin

Fosol:
The Ingest admin page has an 'Ingesting' tab to help debug.  This tab lists related content references.  I've updated this table to include all possible `source` values for the ingest.  Previously it was only filtering on the main `source`.  The papers however ingest many different sources.'

You can also now click on the row and load the content if it exists.

image

### DIFF
--- a/api/net/Areas/Admin/Controllers/ContentReferenceController.cs
+++ b/api/net/Areas/Admin/Controllers/ContentReferenceController.cs
@@ -27,17 +27,17 @@ namespace TNO.API.Areas.Admin.Controllers;
 public class ContentReferenceController : ControllerBase
 {
     #region Variables
-    private readonly IContentReferenceService _service;
+    private readonly IContentReferenceService _contentReferenceService;
     #endregion
 
     #region Constructors
     /// <summary>
     /// Creates a new instance of a ContentReferenceController object, initializes with specified parameters.
     /// </summary>
-    /// <param name="service"></param>
-    public ContentReferenceController(IContentReferenceService service)
+    /// <param name="contentReferenceService"></param>
+    public ContentReferenceController(IContentReferenceService contentReferenceService)
     {
-        _service = service;
+        _contentReferenceService = contentReferenceService;
     }
     #endregion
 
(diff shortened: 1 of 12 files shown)


api/net/Areas/Admin/Controllers/ContentReferenceController.cs
```diff
@@ -106,8 +106,23 @@ public IActionResult Update(ContentReferenceModel model)
...
+    /// <summary>
+    /// Find all the content ids for the specified 'uid'.
+    /// </summary>
+    /// <param name="uid"></param>
+    /// <returns></returns>
+    [HttpGet("content/ids")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(long[]), (int)HttpStatusCode.OK)]
+    [SwaggerOperation(Tags = new[] { "ContentReference" })]
+    public IActionResult FindContentId([FromQuery] string uid)
```
Fosol, line 122: New endpoint to find the content ids for the specified 'uid'.  It should only ever return one item, but it's possible to return many.

app/editor/src/features/admin/ingests/ContentReferenceList.tsx
```diff
@@ -81,6 +88,18 @@ export const ContentReferenceList: React.FC<IContentReferenceListProps> = (props
     [filter],
   );
 
+  const handleRowClick = React.useCallback(
```
Fosol, line 91: Open the content in a new tab.

app/editor/src/features/admin/ingests/ContentReferenceList.tsx
```diff
@@ -22,9 +23,14 @@ export const ContentReferenceList: React.FC<IContentReferenceListProps> = (props
   const [, api] = useContentReferences();
   const { values } = useFormikContext<IIngestModel>();
 
+  const sources = [
```
Fosol, line 26: Filter on all possible `source` values for this ingest.

PR #501, opened 2023-01-19
TNO-874 Fix streams

Fosol:
Fixing API endpoints that access storage/files/streams.  There was an issue where an invalid connection would result in attempting to access files in the wrong location.

Adding FFmpeg loglevel to more ingest configurations.

Fixing a few default radio station URLs.  FFmpeg doesn't support playlists.

### DIFF
--- a/api/net/Areas/Editor/Controllers/StorageController.cs
+++ b/api/net/Areas/Editor/Controllers/StorageController.cs
@@ -4,7 +4,6 @@ using System.Web;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
-using Renci.SshNet;
 using Swashbuckle.AspNetCore.Annotations;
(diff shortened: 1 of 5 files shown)


api/net/Areas/Editor/Controllers/StorageController.cs
```diff
@@ -67,11 +66,11 @@ public StorageController(IConnectionHelper connection, IOptions<StorageOptions>
     public IActionResult GetFolder([FromRoute] int? locationId, [FromQuery] string? path)
     {
         path = String.IsNullOrWhiteSpace(path) ? "" : HttpUtility.UrlDecode(path).MakeRelativePath();
-        if (locationId != null)
+        var connection = locationId.HasValue ? _connection.GetConnection(locationId.Value) : null;
```
Fosol, line 69: Presently the editor app passes an invalid connection.  The default behaviour should currently be accessing files in the API.  I suspect this is not what we want in the future however.

PR #509, opened 2023-01-23
TNO-939 Fix cron job

Fosol:
Updated `purge-service` CronJob so that it works and doesn't spam logs with invalid info.

### DIFF
--- a/openshift/kustomize/cron/Dockerfile
+++ b/openshift/kustomize/cron/Dockerfile
@@ -1,0 +1,6 @@
+FROM alpine:3.17.1
+
+RUN apk update
+RUN apk add bash
+RUN apk add bash-doc
+RUN apk add bash-completion
--- a/openshift/kustomize/cron/base/config-map.yaml
+++ b/openshift/kustomize/cron/base/config-map.yaml
@@ -16,7 +16,7 @@ metadata:
     created-by: jeremy.foster
 data:
   entrypoint.sh: |-
-    #!/bin/sh
+    #!/bin/bash
 
     # *********************************************************************************
     # Delete folders that are older than the specified number of days.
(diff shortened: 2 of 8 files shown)


openshift/kustomize/cron/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine:3.17.1
```
Fosol, line 1: Needed a base image that has `bash` installed.  Would have loved an image I didn't have to customize, but this does the trick.

openshift/kustomize/cron/base/config-map.yaml
```diff
@@ -16,47 +16,38 @@ metadata:
     created-by: jeremy.foster
 data:
   entrypoint.sh: |-
-    #!/bin/sh
+    #!/bin/bash
```
Fosol, line 19: busybox doesn't have `bash`, but `sh` doesn't have a good way to use custom functions.

openshift/kustomize/cron/base/config-map.yaml
```diff
@@ -16,47 +16,38 @@ metadata:
...
+      DELETE_S=$(($(date +%s) - (24*60*60*${2-0})))
+      FOLDER_S=$(date -d "$FOLDER" +%s)
 
-          # Find a match, mark for keep.
-          if [ $MATCH -eq 1 ]; then
-            DELETE=0
-          fi
-          i=$(( $i + 1 ))
-        done
+      if [ "$FOLDER_S" -le "$DELETE_S" ]; then
```
Fosol, line 33: Fixed delete of folder

openshift/kustomize/cron/base/config-map.yaml
```diff
@@ -16,47 +16,38 @@ metadata:
...
-        else
-          echo "Keep - $d"
-        fi
+    purge_folders () {
+      echo "************************************"
+      echo "Purging $1"
+      echo "Deleting folders older than $2 days"
 
-      done
+      find $1/*/*/ -type d -prune -exec bash -c 'delete_folder "$0" $1' {} $2 \;
```
Fosol, line 45: Common and simple way to execute command on each found item.

openshift/kustomize/cron/base/cron-job.yaml
```diff
@@ -23,14 +23,14 @@ spec:
...
-                defaultMode: 0700
+                defaultMode: 0754
                 name: purge-service
             - name: av-storage
               persistentVolumeClaim:
                 claimName: av-storage
           containers:
             - name: purge-service
-              image: busybox:1.35.0
+              image: image-registry.apps.silver.devops.gov.bc.ca/9b301c-tools/alpine:3.17.1
```
Fosol, line 33: Use local image we built for purpose

openshift/kustomize/cron/base/cron-job.yaml
```diff
@@ -23,14 +23,14 @@ spec:
           volumes:
             - name: entrypoint
               configMap:
-                defaultMode: 0700
+                defaultMode: 0754
```
Fosol, line 26: Add execute permission to script

tools/scripts/purge-folders.sh
```diff
@@ -24,12 +24,12 @@ purge_folders () {
   echo "Purging $1"
   echo "Deleting folders older than $2 days"
 
-  find $1/*/*/ -type d -exec bash -c 'delete_folder "$0" $1' {} $2 \;
+  find $1/*/*/ -type d -prune -exec bash -c 'delete_folder "$0" $1' {} $2 \;
```
Fosol, line 27: `-prune` stops the find command from throwing an error after the folder is deleted.

PR #511, opened 2023-01-27
TNO-626: Save qty cache, keep qty on refresh

jdtoombs:
Only saves to cache when page size selector is present (will not override content list view etc)
tno-626


app/editor/public/keycloak.json
```diff
@@ -1,8 +1,8 @@
 {
```
jdtoombs, line 1: will clean this up when i remove this from draft

Fosol (approved):

PR #512, opened 2023-02-01
TNO-818: Toolbar Responsiveness

jdtoombs:
- content list view filter toolbar now changes the Show Only section to a multi-select if screen real-estate is minimal. will sync up with existing values from checkboxes
- license options hidden on the content form toolbar if space is minimal 

**Sync example:**
persist-resize

**Resize example:** 
resize-toolbar

### DIFF
--- a/app/editor/src/features/content/form/constants/showOnlyOptions.ts
+++ b/app/editor/src/features/content/form/constants/showOnlyOptions.ts
@@ -5,7 +5,6 @@ import { ShowOnlyValues } from './ShowOnlyValues';
 export const showOnlyOptions = [
   new OptionItem('Print Content', ShowOnlyValues.PrintContent),
   new OptionItem('Included in EoD', ShowOnlyValues.IncludedInEod),
-  new OptionItem('On Ticker', ShowOnlyValues.PrintContent),
   new OptionItem('Commentary', ShowOnlyValues.Commentary),
   new OptionItem('Top Story', ShowOnlyValues.TopStory),
 ];
--- a/app/editor/src/features/content/form/constants/showOnlyOptions.ts
+++ b/app/editor/src/features/content/form/constants/showOnlyOptions.ts
@@ -1,10 +1,11 @@
-import { ContentTypeName } from 'hooks';
 import { OptionItem } from 'tno-core';
 
+import { ShowOnlyValues } from './ShowOnlyValues';
+
 export const showOnlyOptions = [
-  new OptionItem('Print Content', ContentTypeName.PrintContent),
-  new OptionItem('Included in EoD', 'true'),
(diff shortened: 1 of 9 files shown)


app/editor/src/features/content/tool-bar/sections/filter/utils/checkLatestShowOnlyEntry.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param newValues the current array of selected options
+ * @param onChange the function to perform on change of the filter
+ * @param filter the filter to be updated
+ */
+export const checkLatestShowOnlyEntry = (
+  newValues: IOptionItem[],
+  onChange: (filter: IContentListFilter) => void,
+  filter: IContentListFilter,
+) => {
+  if (newValues[newValues.length - 1].value === ContentTypeName.PrintContent) {
```
Fosol, line 16: Use your filter enum instead of mixing all things things.  And use a switch statement.

Fosol (changes requested):

app/editor/src/features/content/list-view/interfaces/IContentListFilter.ts
```diff
@@ -16,6 +16,7 @@ export interface IContentListFilter {
   userId: number | '';
   timeFrame: number | '';
   // Actions
+  showOnly?: string;
```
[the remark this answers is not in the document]
jdtoombs: Gotcha, just wondering if we want to make it required for the cases of things like queryToFilter as it those values will already be covered by their individual checkboxes

app/editor/src/features/content/form/constants/showOnlyOptions.ts
```diff
@@ -0,0 +1,11 @@
+import { OptionItem } from 'tno-core';
+
+import { ShowOnlyValues } from './ShowOnlyValues';
+
+export const showOnlyOptions = [
+  new OptionItem('Print Content', ShowOnlyValues.PrintContent),
+  new OptionItem('Included in EoD', ShowOnlyValues.IncludedInEod),
+  new OptionItem('On Ticker', ShowOnlyValues.PrintContent),
```
Fosol, line 8: Don't think this makes sense.  I believe also they wanted to remove 'On Ticker' anyway.

Fosol (changes requested):

app/editor/src/features/content/form/constants/showOnlyOptions.ts
jdtoombs: Woops, that was a copy paste error! Good catch - will remove on ticker from both checkboxes and dropdown while I am there though!

Fosol (approved):

PR #518, opened 2023-02-06
TNO-736 Add FFmpeg to API

Fosol:
Surprisingly FFmpeg was never released to Openshift on the API.  I have now added it to the appropriate `Dockerfile`.

I have also updated the resource requirements for the Capture and API pods.

### DIFF
--- a/api/net/Dockerfile.openshift
+++ b/api/net/Dockerfile.openshift
@@ -1,5 +1,10 @@
 ARG BUILD_CONFIGURATION=Release
 FROM image-registry.apps.silver.devops.gov.bc.ca/9b301c-tools/aspnet:7.0 AS base
+
+RUN apt-get update && apt-get -y upgrade
+RUN apt -y install curl libc6-dev libgdiplus ffmpeg
+RUN apt-get clean
+
 EXPOSE 443 8080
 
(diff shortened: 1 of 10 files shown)


api/net/Dockerfile.openshift
```diff
@@ -1,5 +1,10 @@
 ARG BUILD_CONFIGURATION=Release
 FROM image-registry.apps.silver.devops.gov.bc.ca/9b301c-tools/aspnet:7.0 AS base
+
+RUN apt-get update && apt-get -y upgrade
```
Fosol, line 4: I can't believe we never added this...

libs/net/kafka/KafkaListener`.cs
```diff
@@ -201,6 +201,14 @@ public async Task ConsumeAsync(Func<ConsumeResult<TKey, TValue>, Task> action, C
                 _logger.LogWarning("Unexpected consume containing no message");
             }
         }
+        catch (ConsumeException ex)
+        {
+            // No need to thro
+            if (ex.Message.Contains("Subscribed topic not available:"))
```
Fosol, line 207: Don't need to error out on missing topics.

PR #519, opened 2023-02-07
Fix Tekton build

Fosol:
Running into additional issues building any of the services in Openshift.  The Buildah Tekton Task originally was failing because it ran out of space.  However, after mapping to a volume with more space the new error is related to permissions.  Which would imply another Openshift 4.11 security change.

Also increased resources for a few pods.

### DIFF
--- a/api/net/Dockerfile
+++ b/api/net/Dockerfile
@@ -34,13 +34,15 @@ RUN chmod +x /app/entrypoint.sh
 RUN mkdir /data
 VOLUME /data
 RUN chgrp -R 0 /data && \
-    chmod -R g=u /data
+    chmod -R g=u /data && \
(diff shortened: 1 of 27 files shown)


api/net/Dockerfile
```diff
@@ -34,13 +34,15 @@ RUN chmod +x /app/entrypoint.sh
 RUN mkdir /data
 VOLUME /data
 RUN chgrp -R 0 /data && \
-    chmod -R g=u /data
+    chmod -R g=u /data && \
+    chown -R 1001 /data
```
Fosol, line 38: Make the local api build similar to the one running in Openshift.  This was causing permission issues with folders/files created by the root user.

PR #524, opened 2023-02-10
Fix services

Fosol:
The library that loads `.env` files had a breaking change that resulted in not loading the config.

In my efforts to debug Openshift DevOps issues a change to the Dockerfiles has been rolledback.

Templates have been updated to support the latest security requirements and space limitations.

### DIFF
--- a/libs/net/services/Runners/BaseService.cs
+++ b/libs/net/services/Runners/BaseService.cs
@@ -2,6 +2,7 @@
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using DotNetEnv.Configuration;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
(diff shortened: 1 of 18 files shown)


libs/net/services/Runners/BaseService.cs
```diff
@@ -47,9 +48,10 @@ public abstract class BaseService
     /// <param name="args"></param>
     public BaseService(string[] args)
     {
-        DotNetEnv.Env.Load();
+        DotNetEnv.Env.Load($"{Environment.CurrentDirectory}{Path.DirectorySeparatorChar}.env");
```
Fosol, line 51: Fix the missing configuration issue

openshift/kustomize/tekton/base/pipeline-runs/buildah-service.yaml
```diff
@@ -52,7 +52,7 @@ spec:
           resources:
             requests:
               storage: 5Gi
-          storageClassName: netapp-file-standard
+          storageClassName: netapp-block-standard
```
Fosol, line 55: Needs to be block to work

services/net/capture/Dockerfile
```diff
@@ -15,8 +15,8 @@ COPY services/net/command services/net/command
 COPY services/net/capture services/net/capture
 COPY libs/net libs/net
 
-# RUN fix_permissions() { while [ $# -gt 0 ] ; do chgrp -R 0 "$1" && chmod -R g=u "$1"; shift; done } && \
-#     fix_permissions "/app" "/tmp"
+RUN fix_permissions() { while [ $# -gt 0 ] ; do chgrp -R 0 "$1" && chmod -R g=u "$1"; shift; done } && \
```
Fosol, line 18: Rolled backed change

PR #527, opened 2023-02-10
TNO-1065: Hide NLP options for editors

jdtoombs:
Hide the NLP options for editors as they have been confusing and are not yet required

### DIFF
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -36,6 +36,7 @@ import {
   Area,
   Button,
   ButtonVariant,
+  Claim,
   Col,
   FieldSize,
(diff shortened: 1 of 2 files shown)


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -654,7 +657,13 @@ export const ContentForm: React.FC<IContentFormProps> = ({
                         Transcribe
                       </Button>
                     </Show>
-                    <Show visible={!!props.values.id && props.values.body.length > 0}>
+                    <Show
```
Fosol, line 660: Just remove it for now.  We don't need it presently.

Fosol (changes requested):

Fosol (approved):

PR #532, opened 2023-02-14
TNO-927 Improve error message

Fosol:
When creating a new record with a primary key value that already exists, a friendly error will be returned.

Also fixed the DB migration DevOps pipeline.

image

### DIFF
--- a/api/net/Middleware/ErrorHandlingMiddleware.cs
+++ b/api/net/Middleware/ErrorHandlingMiddleware.cs
@@ -11,7 +11,7 @@ using TNO.Core.Exceptions;
 namespace TNO.API.Middleware
 {
     /// <summary>
-    /// ErrorHandlingMidleware class, provides a way to catch and handle unhandled errors in a generic way.
+    /// ErrorHandlingMiddleware class, provides a way to catch and handle unhandled errors in a generic way.
     /// </summary>
     public class ErrorHandlingMiddleware
     {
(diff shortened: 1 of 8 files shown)


api/net/Middleware/ErrorHandlingMiddleware.cs
```diff
@@ -99,7 +99,10 @@ private async Task HandleExceptionAsync(HttpContext context, Exception ex)
             else if (ex is DbUpdateException)
             {
                 code = HttpStatusCode.BadRequest;
-                message = "A database error occurred while updating.";
+                if (ex.InnerException?.Message.Contains("23505: duplicate key value violates unique constraint") == true)
```
Fosol, line 102: I don't love manually determining a relevant error message, but I can't think of anything more secure for production.  We don't want raw errors getting out to users.

app/subscriber/src/store/hooks/useAjaxWrapper.ts
```diff
@@ -30,14 +30,17 @@ export const useAjaxWrapper = () => {
...
+          let detail = '';
           const data = ae.response?.data as any;
 
           if (ae.response?.status === 401) message = 'Authentication required.';
           else if (ae.response?.status === 403)
             message = 'Authorization required.  Your account does not have access.';
           else if (typeof data === 'string' && !!data) message = data;
-          else if (!!data?.error) message = `${data?.error}\n${data.details}`;
-          else if (!!data?.errors) {
+          else if (!!data?.error) {
```
Fosol, line 40: Need to remove all this duplication in the subscriber app.

PR #537, opened 2023-02-17
Add HDMI ingest config

Fosol:
Provide a way to configure HDMI ingest configurations for capturing audio and video.

image

### DIFF
--- a/libs/net/services/Config/IngestServiceOptions.cs
+++ b/libs/net/services/Config/IngestServiceOptions.cs
@@ -6,10 +6,10 @@ namespace TNO.Services.Config;
 public class IngestServiceOptions : ServiceOptions
 {
     #region Properties
-    /// <summary>
-    /// get/set - The type of service that will be run (i.e. stream|clip|tuner|RPi).
+    /// <summary>"
+    /// get/set - An array of service types that will be run (i.e. "stream", "clip", "tuner", "RPi", "HDMI").
     /// </summary>
-    public string ServiceType { get; set; } = "";
+    public string ServiceTypes { get; set; } = "";
 
     /// <summary>
     /// get/set - An array of ingest type this service will ingest (comma separated) (i.e. "Audio, Video").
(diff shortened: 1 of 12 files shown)


openshift/kustomize/tekton/base/tasks/db-migration-conditional.yaml
```diff
@@ -71,7 +71,7 @@ spec:
           IMAGE=image-registry.apps.silver.devops.gov.bc.ca/$(params.PROJECT_SHORTNAME)-tools/$(params.IMAGE):$(params.IMAGE_TAG)
           CONNECTION_STRING=$(oc -n $PROJECT get configmap $(params.API_NAME) -o jsonpath='{.data.CONNECTION_STRING}')
           DB_USERNAME=$(oc -n $PROJECT get secret $(params.DB_SECRET_NAME) -o jsonpath='{.data.user}' | base64 -d)
-          # DB_PASSWORD=$(oc -n $PROJECT get secret $(params.DB_SECRET_NAME) -o jsonpath='{.data.password}' | base64 -d)
+          DB_PASSWORD=$(oc -n $PROJECT get secret $(params.DB_SECRET_NAME) -o jsonpath='{.data.password}' | base64 -d)
```
Fosol, line 74: Fix DB migration task

libs/net/services/Config/IngestServiceOptions.cs
```diff
@@ -6,10 +6,10 @@ namespace TNO.Services.Config;
 public class IngestServiceOptions : ServiceOptions
 {
     #region Properties
-    /// <summary>
-    /// get/set - The type of service that will be run (i.e. stream|clip|tuner|RPi).
+    /// <summary>"
+    /// get/set - An array of service types that will be run (i.e. "stream", "clip", "tuner", "RPi", "HDMI").
     /// </summary>
-    public string ServiceType { get; set; } = "";
+    public string ServiceTypes { get; set; } = "";
```
Fosol, line 12: We can now support an array of different service types.

PR #545, opened 2023-02-22
TNO-1041 Add SignalR back-plane

Fosol:
SignalR maintains state locally and is unaware of other instances of the API that may be running.  When running the API as a cluster (more than one pod) it requires a SignalR backpane to support broadcasting messages to every pod.  A Kafka Hub backpane has been created to support clusters.

All messages sent to the Hub will be initially published to Kafka.  The API is now a Kafka Consumer which will then receive the latest message and send it to the appropriate user now mater which pod they are connected to.

For all this to work the API will need to be reconfigured as a statefulset so that each pod will have a consistent and different Kafka Consumer Group ID.  Otherwise they would not all receive each message, or they would end generating new Group ID values that would result in receiving all messages again.

In addition, we are now able to send messages directly to a single user instead of broadcasting to all users.

## Highlights

- Fixed bug with Clip Service
- Added Indexing Service to `make up p=main` because we are now using Elasticsearch for results
- Update your API with `make refresh n=api` once this PR is merged

### DIFF
--- a/api/net/Areas/Editor/Controllers/WorkOrderController.cs
+++ b/api/net/Areas/Editor/Controllers/WorkOrderController.cs
@@ -243,7 +243,6 @@ public class WorkOrderController : ControllerBase
                         workOrder.Status = WorkOrderStatus.Failed;
                         workOrder.Note = "Remote file request to Kafka failed";
                         workOrder = _workOrderService.UpdateAndSave(workOrder);
-                        await _hub.Clients.All.SendAsync("WorkOrder", new WorkOrderMessageModel(workOrder, _serializerOptions));
                         throw new BadRequestException("Remote file request failed");
                     }
                     return new JsonResult(new WorkOrderMessageModel(workOrder, _serializerOptions));
--- a/api/net/Areas/Services/Controllers/ContentController.cs
+++ b/api/net/Areas/Services/Controllers/ContentController.cs
@@ -35,6 +35,7 @@ public class ContentController : ControllerBase
     #region Variables
     private readonly IContentService _contentService;
     private readonly IFileReferenceService _fileReferenceService;
+    private readonly IUserService _userService;
     private readonly StorageOptions _storageOptions;
     private readonly IHubContext<WorkOrderHub> _hub;
     #endregion
@@ -45,16 +46,19 @@ public class ContentController : ControllerBase
     /// </summary>
     /// <param name="contentService"></param>
     /// <param name="fileReferenceService"></param>
+    /// <param name="userService"></param>
     /// <param name="hub"></param>
     /// <param name="storageOptions"></param>
     public ContentController(
         IContentService contentService,
         IFileReferenceService fileReferenceService,
+        IUserService userService,
         IHubContext<WorkOrderHub> hub,
         IOptions<StorageOptions> storageOptions)
     {
         _contentService = contentService;
         _fileReferenceService = fileReferenceService;
+        _userService = userService;
         _hub = hub;
         _storageOptions = storageOptions.Value;
     }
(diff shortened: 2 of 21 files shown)


api/net/Areas/Editor/Controllers/WorkOrderController.cs
```diff
@@ -243,7 +243,6 @@ public async Task<IActionResult> RequestFileAsync([FromRoute] int locationId, [F
                         workOrder.Status = WorkOrderStatus.Failed;
                         workOrder.Note = "Remote file request to Kafka failed";
                         workOrder = _workOrderService.UpdateAndSave(workOrder);
-                        await _hub.Clients.All.SendAsync("WorkOrder", new WorkOrderMessageModel(workOrder, _serializerOptions));
```
Fosol, line 246: Not required.  A failed request should return an appropriate error response already.

api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -149,7 +153,13 @@ public async Task<IActionResult> UploadFile([FromRoute] long id, [FromQuery] lon
 
         // Send notification to user who requested the content.
         if (content.OwnerId.HasValue)
-            await _hub.Clients.All.SendAsync("Content", new ContentMessageModel(content));
+        {
+            var owner = content.Owner ?? _userService.FindById(content.OwnerId.Value);
```
Fosol, line 157: The owner of the content will be informed the file has been uploaded.

api/net/SignalR/WorkOrderHub.cs
```diff
@@ -4,8 +4,9 @@
 namespace TNO.API.SignalR;
 
 /// <summary>
-/// The WorkOrderHub class
+/// The WorkOrderHub class, provides the SignalR hub to control messages.
 /// </summary>
+[AllowAnonymous]
```
Fosol, line 9: Resolve the error on the login page.

api/net/appsettings.json
```diff
@@ -71,6 +72,13 @@
       "MessageSendMaxRetries": 10000000,
       "BatchSize": 16384,
       "LingerMs": 1
+    },
+    "Consumer": {
+      "GroupId": "API",
```
Fosol, line 77: This value needs to be overridden in Openshift with a dynamic statefulset variable.

services/docker-compose.yml
```diff
@@ -173,6 +173,7 @@ services:
     profiles:
       - all
       - service
+      - main
```
Fosol, line 176: Indexing service is now part of the `main` profile.  This is required because we are now using Elasticsearch for results.

services/net/clip/ClipManager.cs
```diff
@@ -37,11 +37,11 @@ public class ClipManager : IngestManager<ClipIngestActionManager, ClipOptions>
     public override async Task<IEnumerable<IngestModel>> GetIngestsAsync()
     {
         var ingests = await base.GetIngestsAsync();
-        var serviceType = !String.IsNullOrWhiteSpace(this.Options.ServiceType) ? this.Options.ServiceType : "clip";
         var hostname = System.Environment.GetEnvironmentVariable("HOSTNAME");
+        var serviceTypes = this.Options.GetServiceTypes();
 
         return ingests.Where(i =>
-            i.GetConfigurationValue("serviceType") == serviceType &&
+            (!serviceTypes.Any() || serviceTypes.Any(st => st == i.GetConfigurationValue("serviceType"))) &&
```
Fosol, line 44: Fixed bug with Clip Server I introduced in recent PR.

libs/net/kafka/SignalR/KafkaHubLifetimeManager`.cs
```diff
@@ -0,0 +1,495 @@
...
+
+namespace TNO.Kafka.SignalR;
+
+/// <summary>
+/// KafkaHubLifetimeManager class, provides a Kafka back-plane to support clusters.
+/// All messages are routed to Kafka.
+/// This hub is a Kafka consumer and will action the message once it consumes it.
+/// </summary>
+/// <typeparam name="THub"></typeparam>
+public class KafkaHubLifetimeManager<THub> : DefaultHubLifetimeManager<THub>, IDisposable
```
Fosol, line 19: This is where all the magic occurs to support a Kafka backpane.  It's a simple implementation compared to other examples.

Here is an example of a far more complex implementation - https://github.com/jbendson/SignalR.Kafka
I'm not certain if my implementation (especially the serializer/deserializer) will work in all cases.

PR #552, opened 2023-02-22
Fix indexing content from service

Fosol:
Content send to the API via ingestion services should also automatically index.

### DIFF
--- a/api/net/Areas/Services/Controllers/ContentController.cs
+++ b/api/net/Areas/Services/Controllers/ContentController.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.Options;
 using Swashbuckle.AspNetCore.Annotations;
 using TNO.API.Areas.Services.Models.Content;
+using TNO.API.Config;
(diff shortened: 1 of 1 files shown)


api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -110,14 +125,25 @@ public IActionResult FindByUid([FromQuery] string uid, [FromQuery] string? sourc
...
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Content" })]
-    public IActionResult Add(ContentModel model)
+    public async Task<IActionResult> AddAsync(ContentModel model)
     {
-        var result = _contentService.AddAndSave((Content)model);
-        return new JsonResult(new ContentModel(result));
+        var content = _contentService.AddAndSave((Content)model);
+
+        if (!String.IsNullOrWhiteSpace(_kafkaOptions.IndexingTopic))
```
Fosol, line 132: Need to index the content immediately.  This is why we're missing so much content.

PR #559, opened 2023-02-23
Enhance Elasticsearch migration

Fosol:
I've updated the scripts that provide a way to automate Elasticsearch migrations.

It now supports aliases and full rollback migrations.

> Run `make update-elastic v=1.0.1` to update your local indexes.


.mailmap
```diff
@@ -0,0 +1,6 @@
+Jeremy Foster <jeremy.foster@fosol.ca> <jeremymfoster@hotmail.com>
```
Fosol, line 1: Cleans up duplicates for reporting.

PR #566, opened 2023-02-23
Toning validation hotfix

jdtoombs:
Revert schema back and update how toning group displays validation error. 

- will create a separate ticket to dissect how to get the Yup schema working correctly with an array
toning

### DIFF
--- a/app/editor/src/components/form/toning/ToningGroup.tsx
+++ b/app/editor/src/components/form/toning/ToningGroup.tsx
@@ -12,7 +12,6 @@ export interface IToningGroupProps {
 export const ToningGroup: React.FC<IToningGroupProps> = ({ fieldName }) => {
   const toningOptions = [-5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5];
   const { values, setFieldValue, touched, errors } = useFormikContext<IContentModel>();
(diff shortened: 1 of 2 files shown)


app/editor/src/components/form/toning/ToningGroup.tsx
```diff
@@ -57,7 +56,7 @@ export const ToningGroup: React.FC<IToningGroupProps> = ({ fieldName }) => {
           </Col>
         ))}
       </Row>
-      <Error error={!!fieldName && touched[fieldName] ? errors[fieldName] : ''} />
+      <Error error={!active && active !== 0 && touched[fieldName] ? 'Toning is required' : ''} />
```
Fosol, line 59: I don't understand the change here?  The original should display the error that was applied in the yup schema.

jdtoombs: The original schema is setup to track a number; however, tonePools is an array. So I changed the schema accordingly; however, and empty array counts as truthy. So I tried adding a minimum entries of 1 - but that was not triggering the validation appropriately. Going to reapproach it, but just putting this up so things can be saved in the meantime

Fosol (approved):
Temporary fix, need to fix this correctly

PR #573, opened 2023-02-24
TNO-1113 Fix delete content

Fosol:
Updated the delete content process so that the Indexing Service will no longer make a request for the content before attempting to delete it.

This does require changing the key we use in Elasticsearch.  Which means I need to delete the index and rebuild it.

### DIFF
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -138,10 +138,11 @@ export const ContentForm: React.FC<IContentFormProps> = ({
           setContentType(content.contentType);
         } else {
           toast.error('Content requested could not be found.');
+          navigate('/contents');
         }
       });
     },
-    [getContent, findWorkOrders],
+    [getContent, findWorkOrders, navigate],
   );
 
   const onWorkOrder = React.useCallback(
--- a/app/editor/src/store/hooks/useAjaxWrapper.ts
+++ b/app/editor/src/store/hooks/useAjaxWrapper.ts
@@ -39,7 +39,7 @@ export const useAjaxWrapper = () => {
           else if (typeof data === 'string' && !!data) message = data;
           else if (!!data?.error) {
             message = `${data?.error}`;
-            detail = data?.details;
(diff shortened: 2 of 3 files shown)


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -138,10 +138,11 @@ export const ContentForm: React.FC<IContentFormProps> = ({
           setContentType(content.contentType);
         } else {
           toast.error('Content requested could not be found.');
+          navigate('/contents');
         }
       });
     },
-    [getContent, findWorkOrders],
+    [getContent, findWorkOrders, navigate],
```
Fosol, line 145: Navigate back to the Content List View if there is no content to load.

app/editor/src/store/hooks/useAjaxWrapper.ts
```diff
@@ -39,7 +39,7 @@ export const useAjaxWrapper = () => {
           else if (typeof data === 'string' && !!data) message = data;
           else if (!!data?.error) {
             message = `${data?.error}`;
-            detail = data?.details;
+            detail = message.trim() !== data?.details?.trim() ? data?.details : undefined;
```
Fosol, line 42: Removes duplicate error message information.

services/net/indexing/IndexingManager.cs
```diff
@@ -219,34 +219,28 @@ private async Task HandleMessageAsync(ConsumeResult<string, IndexRequestModel> r
...
             this.Logger.LogInformation("Indexing content from Topic: {Topic}, Content ID: {Key}", result.Topic, result.Message.Key);
+            var model = result.Message.Value;
 
-            // TODO: Failures after receiving the message from Kafka will result in missing content.  Need to handle this scenario.
-
-            var content = await this.Api.FindContentByIdAsync(result.Message.Value.ContentId);
-            if (content != null)
+            if (model.Action == IndexAction.Delete)
+            {
+                await DeleteContentAsync(model.ContentId);
```
Fosol, line 226: Delete by the key only.

PR #574, opened 2023-02-27
Save without publish bug

jdtoombs:
Looks like a small bug was introduced when rearranging the buttons. This will now correctly trigger the submission


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -676,6 +676,7 @@ export const ContentForm: React.FC<IContentFormProps> = ({
                     variant={ButtonVariant.secondary}
                     disabled={props.isSubmitting}
                     onClick={() => setSavePressed(true)}
+                    type="submit"
```
Fosol, line 679: This belongs on the Publish button

Fosol (changes requested):

PR #576, opened 2023-02-27
TNO-1128 Fix indexing concurrency issue

Fosol:
Fixed bug with change to the Publish, Unpublish, and Save without publishing buttons.

Also fixed bug when the Indexing Service updates the status of the content it causes a concurrency error when the Editor attempts to save again.  SignalR now will inform the user of the change and it will automatically fetch the latest version.

> Note this change introduced tech-debt related to updating as the editor could lose their changes if the delay is more than a few seconds.  Ideally any updates would check if there are any changes and then auto resolve them, or inform the user to resolve them before refreshing.


api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -147,6 +147,39 @@ public async Task<IActionResult> AddAsync(ContentModel model)
...
+    /// Publish message to kafka to index content in elasticsearch.
+    /// </summary>
+    /// <param name="model"></param>
+    /// <returns></returns>
+    [HttpPut("{id}")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(ContentModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Content" })]
+    public async Task<IActionResult> UpdateAsync(ContentModel model)
```
Fosol, line 161: Added this to the services so that they will use SignalR to inform users of updates.

api/net/SignalR/MessageHub.cs
```diff
@@ -0,0 +1,56 @@
...
+using TNO.API.Models.SignalR;
+
+namespace TNO.API.SignalR;
+
+/// <summary>
+/// The MessageHub class, provides the SignalR hub to control messages.
+/// </summary>
+[AllowAnonymous]
+[Authorize]
+public class MessageHub : Hub
```
Fosol, line 14: Updated to include the common functions.

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -157,6 +157,18 @@ export const ContentForm: React.FC<IContentFormProps> = ({
...
+  const onContentUpdated = React.useCallback(
+    (content: IContentModel) => {
+      // TODO: Don't overwrite the user's edits.
+      if (!!id && +id === content.id) fetchContent(content.id);
+    },
+    [fetchContent, id],
+  );
+
+  React.useEffect(() => {
+    return hub.listen(HubMethodName.Content, onContentUpdated);
```
Fosol, line 169: Listen for content being updated.

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -157,6 +157,18 @@ export const ContentForm: React.FC<IContentFormProps> = ({
     return hub.listen(HubMethodName.WorkOrder, onWorkOrder);
   }, [hub, onWorkOrder]);
 
+  const onContentUpdated = React.useCallback(
+    (content: IContentModel) => {
+      // TODO: Don't overwrite the user's edits.
+      if (!!id && +id === content.id) fetchContent(content.id);
```
Fosol, line 163: If the content was updated then fetch the latest.

jdtoombs (approved):

PR #580, opened 2023-02-28
TNO-1122: Validation

jdtoombs:
Cleaned up toning validation, we had some unnecessary placeholder fields for tone. Now the component will directly change `toningPools` instead of going through various variables.  

Also added some print content validation, as well as the conditional validation of `otherSource/source` as it was a quick add.

### DIFF
--- a/app/editor/src/components/form/toning/ToningGroup.tsx
+++ b/app/editor/src/components/form/toning/ToningGroup.tsx
@@ -1,5 +1,5 @@
 import { useFormikContext } from 'formik';
-import { IContentModel } from 'hooks';
+import { IContentModel, useLookupOptions } from 'hooks';
 import React from 'react';
 import { Col, Error, Row } from 'tno-core';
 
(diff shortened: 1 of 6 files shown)


app/editor/src/components/form/toning/ToningGroup.tsx
```diff
@@ -47,16 +49,18 @@ export const ToningGroup: React.FC<IToningGroupProps> = ({ fieldName }) => {
               key={option}
               onClick={() => {
                 setActive(option);
-                setFieldValue('tonePool', option);
```
jdtoombs, line 50: Not sure what 7 months ago Jonny was thinking here

app/editor/src/features/content/validation/ContentFormSchema.ts
```diff
@@ -1,28 +1,43 @@
...
+      then: string().trim().required('Summary is a required field.'),
+    }),
+    body: string().when('contentType', {
+      is: (value: ContentTypeName) =>
+        value !== ContentTypeName.Snippet && value !== ContentTypeName.Image,
+      then: string().trim().required('Summary is a required field.'),
+    }),
+    // TODO: Headline should not be empty.
+    headline: string().required('Headline is a required field.'),
+    tonePools: array().when('contentType', {
```
jdtoombs, line 28: updated to check array for a min length

Fosol (approved):
Like this implementation a lot better than the prior.

PR #584, opened 2023-02-28
TNO-804 Add service account authorization

Fosol:
Keycloak Gold realm CSS tool now support roles for service accounts.

This allows us to secure our endpoints so that both service accounts and users follow the same permission processes.

You will need to run `make renew` to update Keycloak with the appropriate service account roles.

Won't know if it all works until it's deployed to DEV.  Locally it appears to work correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,15 +125,26 @@ refresh: ## Stop, build, and start all containers or the one specified (args: n=
 	@make build $(if $(n),n=$(n),) $(if $(p),p=$(p),)
 	@make up $(if $(n),n=$(n),) $(if $(p),p=$(p),)
 
+refresh-keycloak: ## Stop, build, drop DB and recreate
+	$(info Stop, build, drop DB and recreate)
+	@make stop n=keycloak
+	@./tools/scripts/docker-remove.sh keycloak
+	@./tools/scripts/refresh-keycloak.sh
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -125,15 +125,26 @@ refresh: ## Stop, build, and start all containers or the one specified (args: n=
 	@make build $(if $(n),n=$(n),) $(if $(p),p=$(p),)
 	@make up $(if $(n),n=$(n),) $(if $(p),p=$(p),)
 
+refresh-keycloak: ## Stop, build, drop DB and recreate
```
Fosol, line 128: New `make refresh-keycloak` to reset the keycloak container and Db.

api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -23,14 +23,14 @@
...
+using TNO.Keycloak;
 
 namespace TNO.API.Areas.Editor.Controllers;
 
 /// <summary>
 /// ContentController class, provides Content endpoints for the api.
 /// </summary>
-// [ClientRoleAuthorize(ClientRole.Editor)]
-[Authorize]
+[ClientRoleAuthorize(ClientRole.Editor)]
```
Fosol, line 33: Authorize by role again.

PR #587, opened 2023-03-03
TNO-821: New tag input

jdtoombs:
This PR includes:
- new tag component that will extract pattern matched tags from the WYSIWYG into a react-select component
- user may enter them via the WYSIWYG or by selecting from the dropdown
- a show list button has been added for a table reference including the full names of tags
- the user may position the table wherever they like when it is opened

new-tags


Fosol (approved):

jdtoombs:
Good to go now 😄 @Fosol

app/editor/package.json
```diff
@@ -35,6 +35,7 @@
     "react-datepicker": "4.10.0",
     "react-dom": "18.2.0",
     "react-drag-drop-files": "2.3.9",
+    "react-draggable": "^4.4.5",
```
Fosol, line 38: Don't use ranges on the app, it causes issues later on

Fosol (changes requested):

jdtoombs:
updated again @Fosol

Fosol (approved):

PR #589, opened 2023-03-03
TNO-1141 Add Elastic Migration Service

Fosol:
We now have a way to manage and perform both updates and rollbacks to our Elasticsearch database.

The Elastic Migration Tool works in a similar way to the Entity Framework DB Migration Tool.  We can create migrations that include both code and scripts.  There are a few conventions, mainly related to sort order of versions and files.

> You will need to run `make renew` to update your elastic container


libs/net/core/Converters/Int32ToStringJsonConverter.cs
```diff
@@ -6,14 +6,30 @@ namespace TNO.Core.Converters
...
         {
-            return reader.TokenType switch
+            var value = reader.TokenType switch
             {
                 JsonTokenType.String => reader.GetString(),
                 JsonTokenType.Number => reader.TryGetInt32(out int result) ? $"{result}" : "",
                 _ => null,
             };
+
+            if (value != null) return value;
```
Fosol, line 24: Dealing with scenario where the JSON has an object.

libs/net/dal/Elasticsearch/TNOElasticClient.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Microsoft.Extensions.Options;
+using Nest;
+using TNO.Core.Exceptions;
+using TNO.DAL.Config;
+
+namespace TNO.DAL.Elasticsearch
```
Fosol, line 6: Moved this to the DAL instead of being in the API.  We use it in the Elastic Migration Tool.

openshift/kustomize/api/base/deploy.yaml
```diff
@@ -105,6 +105,11 @@ spec:
                 configMapKeyRef:
                   name: api
                   key: KEYCLOAK_ISSUER
+            - name: Keycloak__Secret
```
Fosol, line 108: This was missing but is needed to support the CSS service account roles,

openshift/kustomize/api/base/deploy.yaml
```diff
@@ -159,7 +164,7 @@ spec:
                   name: crunchy-pguser-admin
                   key: password
 
-            - name: ELASTIC_URIS
+            - name: Elastic__Url
```
Fosol, line 167: Fixed name

tools/elastic/migration/Dockerfile
```diff
@@ -0,0 +1,35 @@
+ARG BUILD_CONFIGURATION=Release
+FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
+
+ENV DOTNET_CLI_HOME=/tmp
+ENV PATH="$PATH:/tmp/.dotnet/tools"
+
+USER 0
+
+# WORKDIR /src
```
Fosol, line 9: I may need this in Openshift.  Will know once I am able to test.

tools/elastic/migration/ExampleHostedService.cs
```diff
@@ -0,0 +1,55 @@
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace TNO.Elastic.ExampleHosted;
+
+/// <summary>
+/// ExampleHostedService class, provides an example hosted service that stops immediately.
+/// </summary>
+public class ExampleHostedService : IHostedService
```
Fosol, line 9: I found interesting information about hosted services.  After reading up on it I think I should eventually refactor the services I've built to become hosted services.

tools/elastic/migration/Migration.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            var files = Directory.GetFiles(fullPath).OrderBy(v => v).ToArray();
+
+            foreach (var file in files.Where(n => n.EndsWith(".json")))
+            {
+                // Perform each action in order.
+                builder.Logger.LogInformation("Applying migration step '{file}'", file);
+                using var stream = File.OpenRead(file);
+                var step = JsonSerializer.Deserialize<MigrationStep>(stream, builder.SerializerOptions) ?? throw new InvalidOperationException($"Failed to deserialize '{file}'");
+                if (step.Action == MigrationAction.CreateIndex)
```
Fosol, line 127: I'd love to come up with a better way to do this, but for now it does what we need.

PR #600, opened 2023-03-09
Update react-tooltip in tno-core

jdtoombs:
Our button component uses react-tooltip separate from that of our base application, the packages have been updated now in `tno-core`.

This is why we were seeing the old tooltips appearing on the buttons in the Ingest Form.


libs/npm/core/package.json
```diff
@@ -41,7 +41,7 @@
     "react-table": "^7.8.0",
     "react-text-mask": "^5.5.0",
     "react-toastify": "^9.1.1",
-    "react-tooltip": "^4.5.1",
+    "react-tooltip": "5.9.0",
```
Fosol, line 44: This needs to be a `^5.9.0` as it's a package and not the project, otherwise it causes other kinds of issues.

Fosol (changes requested):

PR #603, opened 2023-03-14
TNO-843 Fix 1.0.6 elastic migration

Fosol:
There are number of issues with our implementation of Elasticsearch that are causing issues.

- For some reason Openshift gives us intermittent gateway 502 errors.  I've had to ignore these errors and essentially just keep make requests until they succeed.  I do not know what is causing this.  It could be a combination of resourcing and load balancing issues when under load.
- Reindexing content takes a long time, which results in gateway 504 timeout issues.  To handle this I've added a looping request to determine when the reindex completes.
- The library package NEST was selected to make search requests to elasticsearch from the API.  This library does not currently support System.Text.Json.  Which is problematic because this is how standard dotnet apps configure JSON serialization.  This means that NEST does not serialize correctly.  The new version of NEST is not available yet and it will apparently support proper serialization configuration.  I attempted to rewrite all code and remote NEST, however the amount of work is extreme and I don't have time.  Instead I've provided a separate library in the Elastic Migration Tool for indexing documents which does support serialization configuration.

> You may need to run `make elastic-update` if you haven't updated to `v1.0.6` yet.


services/net/capture/Dockerfile
```diff
@@ -16,7 +16,7 @@ COPY services/net/capture services/net/capture
 COPY libs/net libs/net
 
 RUN fix_permissions() { while [ $# -gt 0 ] ; do chgrp -R 0 "$1" && chmod -R g=u "$1"; shift; done } && \
-    fix_permissions "/app" "/tmp"
+    fix_permissions "/tmp"
```
Fosol, line 19: This was throwing warnings in all our builds.

services/net/indexing/TNO.Services.Indexing.csproj
```diff
@@ -21,7 +21,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Elastic.Clients.Elasticsearch" Version="8.0.0-beta.4" />
+    <PackageReference Include="Elastic.Clients.Elasticsearch" Version="8.0.6" />
```
Fosol, line 24: This library supports System.Text.Json configuration.  However, I am not able to figure out how to do low level requests, or date range queries.

tools/elastic/migration/MigrationBuilder.cs
```diff
@@ -16,6 +17,11 @@ public class MigrationBuilder
     /// </summary>
     public IElasticClient Client { get; }
 
+    /// <summary>
+    /// get - Elastic client for indexing because the IElasticClient does not use System.Text.Json.
+    /// </summary>
+    public ElasticsearchClient IndexingClient { get; }
```
Fosol, line 23: It is annoying to have two elastic clients, but to rewrite all the functionality that came with NEST will take far too long.

tools/elastic/migration/MigrationService.cs
```diff
@@ -77,29 +78,40 @@ public async Task RunAsync(CancellationToken cancellationToken)
...
+                    _logger.LogError(createResponse.OriginalException, "Failed to create migration index.");
+                    throw createResponse.OriginalException;
+                }
             }
         }
+        catch (ElasticsearchClientException ex)
+        {
+            _logger.LogError(ex, "Failure to create index");
+            if (ex.Response.HttpStatusCode == 502)
+                await CreateMigrationIndexAsync(cancellationToken);
```
Fosol, line 111: Infinite loop until successful... Horrible implementation, but the gateway is having issues.

tools/elastic/migration/Migrations/1.0.6/up/pre/00-pipeline-content.json
```diff
@@ -0,0 +1,33 @@
+{
+  "action": "createPipeline",
```
Fosol, line 2: Pipelines are a better solution than manually reindexing content, but for now we'll use both processes.

PR #606, opened 2023-03-14
TNO-629 Fix topic score calculation

Fosol:
Now when selecting a topic if there is a score rule that can be applied it will automatically set the score value.

image


app/editor/src/features/content/form/ContentStoryForm.tsx
```diff
@@ -193,15 +159,49 @@ export const ContentSummaryForm: React.FC<IContentSummaryFormProps> = ({
                 }}
               />
             </Show>
+            <Show visible={contentType !== ContentTypeName.Image}>
```
Fosol, line 162: Updated format to reduce whitespace and poor use of vertical space.

app/editor/src/features/content/form/utils/calcTopicScore.ts
```diff
@@ -16,19 +16,37 @@ export const calcTopicScore = (values: IContentForm, rules: ITopicScoreRuleModel
     .every((r) => {
       // Eliminate any rule that doesn't match the content values.
       if (r.section !== undefined && values.section !== r.section) return true;
```
Fosol, line 18: Fixed score rule formula logic

libs/net/models/Areas/Editor/Models/TopicScoreRule/TopicScoreRuleModel.cs
```diff
@@ -83,7 +83,7 @@ public TopicScoreRuleModel(Entities.TopicScoreRule entity)
         this.SourceId = entity.SourceId;
         this.Section = entity.Section;
         this.PageMin = entity.PageMin;
-        this.PageMin = entity.PageMin;
+        this.PageMax = entity.PageMax;
```
Fosol, line 86: Fixed API bug

PR #607, opened 2023-03-14
TNO-1162 Fix transcription service

Fosol:
Resolving issue when requesting a transcript for a story that has tag values.


libs/net/models/Areas/Services/Models/Content/ContentTagModel.cs
```diff
@@ -34,9 +34,18 @@ public class ContentTagModel : AuditColumnsModel
     /// <summary>
     /// Creates a new instance of an ContentTagModel.
     /// </summary>
-    public ContentTagModel(string code) {
+    public ContentTagModel()
```
Fosol, line 37: Serialization requires default contructor.

PR #608, opened 2023-03-14
TNO-1162 Fix Transcript Tab and Tags

Fosol:
Tags now work on the transcript tab.

image


app/editor/src/features/content/form/styled/ContentTranscriptForm.tsx
```diff
@@ -4,10 +4,6 @@ export const ContentTranscriptForm = styled.div`
   margin: 1em;
   height: 90%;
 
-  div {
```
Fosol, line 7: Doesn't seem to be needed.  Removed because it was styling the tag value divs.

PR #609, opened 2023-03-15
TNO-1140 Fix content filter state

Fosol:
Content List filter was getting reset when navigating to different pages.  It will now maintain state.

Note that there is a new issue introduced with the latest Morning Report page filter changes.  Specifically with the Advanced Search, the source value is displayed even though this shared component no longer should display it.

image

### DIFF
--- a/app/editor/src/features/content/list-view/ContentListView.tsx
+++ b/app/editor/src/features/content/list-view/ContentListView.tsx
@@ -48,8 +48,10 @@ export const ContentListView: React.FC = () => {
 
   React.useEffect(() => {
     // Extract query string values and place them into redux store.
-    storeFilter(queryToFilter(filter, window.location.search));
-    storeFilterAdvanced(queryToFilterAdvanced(filterAdvanced, window.location.search));
+    if (!!window.location.search) {
+      storeFilter(queryToFilter(filter, window.location.search));
+      storeFilterAdvanced(queryToFilterAdvanced(filterAdvanced, window.location.search));
+    }
     // Only want this to run on the first load.
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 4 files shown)


app/editor/src/features/content/list-view/ContentListView.tsx
```diff
@@ -48,8 +48,10 @@ export const ContentListView: React.FC = () => {
 
   React.useEffect(() => {
     // Extract query string values and place them into redux store.
-    storeFilter(queryToFilter(filter, window.location.search));
-    storeFilterAdvanced(queryToFilterAdvanced(filterAdvanced, window.location.search));
+    if (!!window.location.search) {
```
Fosol, line 51: Only do this if there are query parameters.  This however may result in future issues if a query parameter is specified for a different reason.  There appears to be a fundamental flaw in the implementation to extract query parameters to influence state.

app/editor/src/features/content/tool-bar/sections/filter/FilterContentSection.tsx
```diff
@@ -40,23 +40,23 @@ export const FilterContentSection: React.FC<IFilterContentSectionProps> = ({
...
 
+  const search = fromQueryString(window.location.search);
   const timeFrames = [
     { label: 'TODAY', value: 0 },
     { label: '24 HRS', value: 1 },
     { label: '48 HRS', value: 2 },
     { label: 'ALL', value: 3 },
   ];
-  const timeFrameSelected = timeFrames[search.timeFrame ?? 0].label;
+  const timeFrameSelected = timeFrames[search.timeFrame ?? filter.timeFrame].label;
```
Fosol, line 51: Use the redux store if the URL has no value.

PR #610, opened 2023-03-15
Update packages

Fosol:
Updated every npm package that I could.

Was unable to updated `yup` because of breaking changes.


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -385,7 +388,7 @@ export const ContentForm: React.FC<IContentFormProps> = ({
               !request.isSilent && request.group.some((g) => g === 'content' || g === 'lookup')
             }
           >
-            {(props) => (
+            {(props: FormikProps<IContentForm>) => (
```
Fosol, line 391: Fixing lint issue

app/editor/src/features/content/validation/ContentFormSchema.ts
```diff
@@ -4,37 +4,41 @@ import { array, date, number, object, string } from 'yup';
 export const ContentFormSchema = object().shape(
   {
     productId: number().required('Product is required.').notOneOf([0], 'Product is required.'),
-    sourceId: number().when('tempSource', {
-      is: undefined,
-      then: number().required('Either source or other source is required.'),
+    sourceId: number().when('tempSource', (value: any) => {
```
Fosol, line 7: Updated to resolve `yup` breaking changes.  Rolled back `yup` however due to the number of breaking changes.  Kept this file update since it works and will be supported when we upgrade.

PR #617, opened 2023-03-16
TNO-931, TNO-1158: Default selected for a.m report

jdtoombs:
- default selected now appears when loading the morning report
- removed show/program from both print content and stories (internet)

### DIFF
--- a/app/editor/src/features/content/form/ContentStoryForm.tsx
+++ b/app/editor/src/features/content/form/ContentStoryForm.tsx
@@ -159,13 +159,7 @@ export const ContentStoryForm: React.FC<IContentStoryFormProps> = ({
                 }}
               />
             </Show>
-            <Show
-              visible={
-                contentType !== ContentTypeName.Image &&
-                contentType !== ContentTypeName.PrintContent &&
-                contentType !== ContentTypeName.Story
-              }
-            >
+            <Show visible={contentType === ContentTypeName.Snippet}>
               <FormikSelect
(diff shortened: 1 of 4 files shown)


app/editor/src/features/content/form/ContentStoryForm.tsx
```diff
@@ -159,7 +159,13 @@ export const ContentStoryForm: React.FC<IContentStoryFormProps> = ({
                 }}
               />
             </Show>
-            <Show visible={contentType !== ContentTypeName.Image}>
+            <Show
+              visible={
+                contentType !== ContentTypeName.Image &&
```
Fosol, line 164: Doesn't this mean only show when `contentType === ContentTypeName.Snippet`?

app/editor/src/features/content/morning-report/constants/defaultReportFilter.ts
```diff
@@ -12,7 +12,7 @@ export const defaultReportFilter: IMorningReportFilter = {
   sourceId: 0,
   otherSource: '',
   productIds: [],
-  sourceIds: [],
+  sourceIds: [91, 248, 249, 246, 247],
```
Fosol, line 15: This is always problematic, as there is no guarantee that these IDs mean the same thing in each environment.  Do it by code instead.  Code isn't a ton better, but at least it's human readable and manageable through the interface.

Fosol (changes requested):

PR #620, opened 2023-03-17
Fix workorder

Fosol:
The admin state changes caused an infinite loop on the Workorder page.

image


app/editor/src/features/admin/work-orders/WorkOrderList.tsx
```diff
@@ -68,8 +70,11 @@ export const WorkOrderList = () => {
   );
 
   React.useEffect(() => {
-    fetch(workOrderFilter);
-  }, [fetch, workOrderFilter]);
+    if (workOrderFilter !== filter) {
+      setFilter(workOrderFilter);
+      fetch(workOrderFilter);
+    }
+  }, [fetch, workOrderFilter, filter]);
```
Fosol, line 77: I'd love to understand how to top the dependency update issue from causing infinite loops.  Essentially the fetch method keeps getting updated every time redux state is updated from requesting workorders.  This is a common problem with react.

PR #624, opened 2023-03-20
TNO-1156 Reorg components

Fosol:
Significant refactor of shared code between the Editor and Subscriber apps.  Move as much as I could into the `tno-core` library.

Organized a few components more intuitively so that they were closer to what uses them.  If a component is tightly coupled with other components they don't belong in the generic `/components` folder.

Regrettably with the number of changes it wasn't possible to keep file history the way I would have liked it.  Careful when resolving merge conflicts.

- Published `tno-core:0.0.222`


app/editor/src/App.tsx
```diff
@@ -3,7 +3,7 @@ import 'react-tooltip/dist/react-tooltip.css';
 
 import { ReactKeycloakProvider } from '@react-keycloak/web';
 import { LayoutAnonymous } from 'components/layout';
-import { UploadContextWrapper } from 'components/upload';
+import { UploadContextWrapper } from 'features/content';
```
Fosol, line 6: Upload is tightly coupled with the content form.  At some point it would be nice to make it generic.  Until then it belongs close to the content form.

app/editor/src/components/dialog/styled/Dialog.tsx
Fosol, line 1: A lot of files were moved but didn't keep their history.  Not worth my time to correct this issue.

PR #625, opened 2023-03-21
Fix services

Fosol:
Fixing coding errors introduced in the services.


libs/net/services/Helpers/ApiService.cs
```diff
@@ -276,7 +276,7 @@ public async Task<IEnumerable<IngestModels.IngestModel>> GetIngestsForTopicAsync
     /// <returns></returns>
     public async Task<IngestModels.IngestModel?> UpdateIngestAsync(
         IngestModels.IngestModel ingest,
-        HttpRequestHeaders? headers)
+        HttpRequestHeaders? headers = null)
```
Fosol, line 279: @CelloLiu I should have caught this when you made this change.  This is how you make an optional parameter.  Your change breaks things without this.

PR #629, opened 2023-03-23
TNO-1117 Add CC script

Fosol:
Basic script to capture Closed Caption and place it into a file with timestamps.

### DIFF
--- a/tools/scripts/cc.sh
+++ b/tools/scripts/cc.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+#########################################
+# Capture Closed Caption
+#
+# Runs command to listen to the /dev/vbi# device
+# and captures the text to the specified file.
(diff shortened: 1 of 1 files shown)


tools/scripts/cc.sh
```diff
@@ -0,0 +1,33 @@
...
+# Runs command to listen to the /dev/vbi# device
+# and captures the text to the specified file.
+# Inserts timestamp every 10 seconds.
+#
+# Arguments
+# $1 = output file (default: cc.txt)
+# $2 = device to listen to (default: /dev/vbi0)
+#########################################
+
+output=${$1:-"cc.txt"}
```
Fosol, line 15: Run this on the linux box with the Hauppauge TV Capture Card `./cc.sh /path/filename.txt`

PR #634, opened 2023-03-27
Fix row/col component

Fosol:
Fixed issue introduced with prior flexbox fix.

Columns and Rows should by default wrap.

image


libs/npm/core/src/components/flex/row/styled/Row.tsx
```diff
@@ -5,7 +5,7 @@ import { IRowProps } from '..';
 export const Row = styled.div<IRowProps>`
   display: flex;
   width: ${(props) => props.width ?? 'auto'};
-  flex-wrap: ${(props) => (props.nowrap ? 'nowrap' : props.wrap ?? 'nowrap')};
+  flex-wrap: ${(props) => (props.nowrap ? 'nowrap' : props.wrap ?? 'wrap')};
```
Fosol, line 8: @CelloLiu this is what you broke when you fixed the styling.  Every Row/Col enforced `nowrap`, which was the opposite of what they were doing originally.  Highlights the issue I mentioned about your initial change.  Wish I caught it when reviewing your second change which was to fix the incorrect keyword `no-warp`.

PR #643, opened 2023-03-29
Update make pack script

jdtoombs:
Putting this up just incase it is useful for anyone else.

Can now do:
`make pack n=editor` OR `make pack n=subscriber`


libs/npm/core/scripts/package.sh
jdtoombs, line 1: Just indentation with the new if statement

Fosol (approved):

PR #645, opened 2023-03-29
Fix advanced source filter

Fosol:
This fixes a bug I introduced in the advanced search when selecting a source.

image


app/editor/src/features/content/list-view/constants/fieldTypes.ts
```diff
@@ -2,7 +2,7 @@ import { OptionItem } from 'tno-core';
 
 export const fieldTypes = [
   new OptionItem('Headline', 'headline'),
-  new OptionItem('Source', 'sourceId'),
+  new OptionItem('Source', 'sourceIds'),
```
Fosol, line 5: The filter is for multiple now.

PR #646, opened 2023-03-29
Fix product filter

Fosol:
Removes the `Any` product as it no longer works.  The product filter is an OR statement, so we would need to update the elasticsearch query to handle this if it's required.

### DIFF
--- a/app/editor/src/features/content/tool-bar/sections/filter/FilterContentSection.tsx
+++ b/app/editor/src/features/content/tool-bar/sections/filter/FilterContentSection.tsx
@@ -9,7 +9,6 @@ import {
   fromQueryString,
   getUserOptions,
   IOptionItem,
-  OptionItem,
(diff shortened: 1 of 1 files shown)


app/editor/src/features/content/tool-bar/sections/filter/FilterContentSection.tsx
```diff
@@ -55,10 +53,6 @@ export const FilterContentSection: React.FC<IFilterContentSectionProps> = () =>
     setUserOptions(getUserOptions(users.filter((u) => !u.isSystemAccount)));
   }, [users]);
 
-  React.useEffect(() => {
-    setProductOptions([new OptionItem<number>('Any', 0), ...pOptions]);
```
Fosol, line 59: Removed 'Any' as it no longer works.

PR #649, opened 2023-04-01
TNO-1188 Refersh DEV env

Fosol:
Fixed various configuration issues with DevOps after resetting DEV.

Also fixed the Elasticsearch migration pipeline so that it now works too!


openshift/kustomize/api/base/config-map.yaml
```diff
@@ -28,6 +28,6 @@ data:
 
   CONNECTION_STRING: Host=crunchy-primary:5432;Database=tno;Include Error Detail=true;Log Parameters=true;
 
-  ELASTIC_URIS: elastic:9200
+  ELASTIC_URIS: http://elastic:9200
```
Fosol, line 31: This stumped me...  Stupid mistake.

PR #657, opened 2023-04-06
TNO-1188 Limit notification to alerts

Fosol:
Resolved a number of DevOps issues.

Added validation to notification so that it only sends if it has an Alert Action set to "true".


libs/net/entities/Validation/INotificationValidator.cs
```diff
@@ -5,21 +5,30 @@ namespace TNO.Entities.Validation;
 /// </summary>
 public interface INotificationValidator
 {
+    #region Properties
+    /// <summary>
+    /// get/set - The primary key to the alert action used to validate content.
+    /// </summary>
+    public int? AlertId { get; set; }
```
Fosol, line 12: Simple way to eliminate hard coding the alert action primary key

openshift/kustomize/postgres/crunchy/base/deploy.yaml
```diff
@@ -32,7 +32,7 @@ spec:
   postgresVersion: 14
   instances:
     - name: postgres
-      replicas: 4
+      replicas: 3
```
Fosol, line 35: Gave us 3 Gb back

PR #663, opened 2023-04-14
TNO-1193 Replace react-table component

Fosol:
Created a new React table component that we can use to replace the react-table library.  Ran into too many issues with react-table to want to continue with it.  There are still pages using the react-table, but I'll remove them later.

The new component also has a headless design, but I've created a basic component that has the features we're using.  It needs a lot more time to make it truly performant, but so far doesn't seem to be slower than react-table.  There are some state management issues that do require a hard refresh when testing regrettably.

Published tno-core:0.0.232

> You'll need to run `make refresh n=editor` to setup your local environment

## Content List View

image

## Morning Papers

image


app/editor/src/features/demo/DemoPage.tsx
```diff
@@ -0,0 +1,151 @@
+import React from 'react';
+import { Button, FlexboxTable, ITableHookColumn, Row, Text } from 'tno-core';
+
+export class DemoData {
```
Fosol, line 4: Helpful page to test stuff... I'll leave it in for now.

libs/npm/core/src/components/table/FlexboxTable.tsx
```diff
@@ -0,0 +1,193 @@
+import React from 'react';
+
+import { Text } from '../form';
+import { ITableProps, SortFlag, TablePager, useTable } from '.';
+import * as styled from './styled';
+
+export const FlexboxTable = <T extends object>({
```
Fosol, line 7: This is a default implementation.  We can create other implementations because it's headless.

libs/npm/core/src/components/table/FlexboxTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+  columns,
+  data,
+  ...rest
+}: ITableProps<T>) => {
+  const table = useTable({
+    rowId,
+    columns,
+    data,
+    options: {
+      isMulti: rest.isMulti,
```
Fosol, line 18: I'd like to find a better way to do this as currently it require manually passing the values down.

libs/npm/core/src/components/table/FlexboxTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+              value={search}
+              onChange={(e) => {
+                setSearch(e.target.value);
+                table.applyFilter(e.target.value);
+              }}
+            />
+          </div>
+        </div>
+      )}
+      {table.options.showHeader && (
```
Fosol, line 79: I'd like to extend the functionality of the hook to include ways to fetch all the attributes similar to react-table.  It would make it easier to create new implementations.

libs/npm/core/src/components/table/FlexboxTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+                  >
+                    <SortFlag column={col} />
+                  </div>
+                )}
+              </div>
+            ))}
+        </header>
+      )}
+      <div className="rows">
+        {table.groupBy && (
```
Fosol, line 116: Basic grouping support.  But it doesn't support paging when grouping.

libs/npm/core/src/components/table/useTable.tsx
```diff
@@ -0,0 +1,190 @@
...
+  TableInternal,
+} from '.';
+
+/**
+ * This hook generates a table object and manages state mutations.
+ * Only the following properties will mutate externally; columns, data, options, paging, sorting.
+ * @param param0 Hook properties
+ * @returns A TableInternal object instance.
+ */
+export const useTable = <T extends object>({
```
Fosol, line 18: I am not a fan of how I implemented state management with this hook.  It seems to work okay for now though.

libs/npm/core/src/components/table/utils/initColumns.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * @returns An updated array of columns.
+ */
+export const initColumns = <T extends object>(
+  table: TableInternal<T>,
+  columns: ITableHookColumn<T>[],
+  sortOrder: ITableSort<T>[] | undefined = [],
+) => {
+  const cols = [...columns];
+  if (table.options.isMulti)
+    cols.unshift(
```
Fosol, line 25: Multi-select comes built in.

PR #669, opened 2023-04-18
Fix product filter

Fosol:
The productIds filter expects an array, however when a single value is returned it is not handled correctly.


app/editor/src/features/content/form/styled/ContentForm.tsx
```diff
@@ -2,7 +2,6 @@ import styled from 'styled-components';
 
 export const ContentForm = styled.div`
   height: 100%;
-  z-index: 1;
```
Fosol, line 5: This was overlaying the dropdown in the select.

libs/npm/core/src/utils/fromQueryString.ts
```diff
@@ -1,22 +1,39 @@
...
+}
+
 /**
  * Converts a query string to an object.
  * @param query The query string to convert to an object.
+ * @param options Query options to control the output.
  * @returns A new instance of an object.
  */
-export const fromQueryString = (query?: string): any => {
+export const fromQueryString = (query?: string, options?: IQueryOptions): any => {
```
Fosol, line 14: We can now properly handle arrays that only have a single value in the query parameters.

PR #673, opened 2023-04-19
TNO-1183 Add Reporting Service

Fosol:
Creating the base Reporting Service that will provide a way to generate and send reports.  The Reporting Service is a Kafka Consumer.

A Report is generated through two primary properties, the `Filter` and the `Template`.  The filter is an Elasticsearch query that will fetch content.  The template uses Razor syntax to dynamically generate the report output.

Additionally, a manual report can also be created by adding content to it and then make a request to send out the report.

- The report does not currently implement the Razor engine (separate story)


api/net/Areas/Services/Controllers/ReportController.cs
```diff
@@ -76,5 +76,25 @@ public IActionResult FindById(int id)
...
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(IEnumerable<Models.Content.ContentModel>), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Report" })]
+    public async Task<IActionResult> FindContentForReportIdAsync(int id)
+    {
+        var report = _service.FindById(id);
+        if (report == null) return new BadRequestResult();
+
+        // TODO: Make request to Elasticsearch for the content that matches the filter.
```
Fosol, line 95: This has been wired up, but not tested.

libs/net/dal/Services/ReportService.cs
```diff
@@ -25,11 +43,41 @@ public IEnumerable<Report> FindAll()
...
+    /// <returns></returns>
+    /// <exception cref="Exception"></exception>
+    public async Task<IEnumerable<Content>> FindContentWithElasticsearchAsync(Report report)
+    {
+        var response = await _client.SearchAsync<Content>(s =>
+        {
+            var result = s
+                .Pretty()
+                .Index(_elasticOptions.PublishedIndex) // TODO: Switch to unpublished if the report settings require it.
+                .Query(q => q.Raw(report.Filter.ToJson()));
```
Fosol, line 71: Raw JSON request.  @CelloLiu @jdtoombs This is what I want to do with the `ContentService` function too at some point.  Then we don't have to convert the `ContentFilter` into an elastic query.  However, it will mean that we need to submit an elastic query from the frontend instead.

services/net/reporting/ReportingManager.cs
```diff
@@ -0,0 +1,442 @@
...
+    /// Process the report request.
+    /// </summary>
+    /// <param name="request"></param>
+    /// <returns></returns>
+    private async Task ProcessReportAsync(ConsumeResult<string, ReportRequestModel> result)
+    {
+        var request = result.Message.Value;
+        if (request.Destination.HasFlag(ReportDestination.ReportingService))
+        {
+            if (request.ReportInstanceId.HasValue)
```
Fosol, line 259: If an `ReportInstanceId` is provided it means we're either sending a custom report or resending an existing report.

services/net/reporting/ReportingManager.cs
```diff
@@ -0,0 +1,442 @@
...
+
+    /// <summary>
+    /// Generate the output of the report with the Razor engine.
+    /// </summary>
+    /// <param name="report"></param>
+    /// <param name="content"></param>
+    /// <returns></returns>
+    private Task<string> GenerateReportSubjectAsync(API.Areas.Services.Models.ReportInstance.ReportInstanceModel instance, IEnumerable<API.Areas.Services.Models.Content.ContentModel> content)
+    {
+        var template = instance.Report?.Settings.GetDictionaryJsonValue<string>("subject");
```
Fosol, line 366: Still need to hook up the Razor enging

services/net/reporting/ReportingManager.cs
```diff
@@ -0,0 +1,442 @@
...
+
+    /// <summary>
+    /// Generate the output of the report with the Razor engine.
+    /// </summary>
+    /// <param name="report"></param>
+    /// <param name="content"></param>
+    /// <returns></returns>
+    private Task<string> GenerateReportBodyAsync(API.Areas.Services.Models.ReportInstance.ReportInstanceModel instance, IEnumerable<API.Areas.Services.Models.Content.ContentModel> content)
+    {
+        return Task.FromResult($"{instance.Report?.Template}");
```
Fosol, line 378: Still need to hook up the Razor enging

libs/net/dal/Services/ReportService.cs
jdtoombs: I like this

jdtoombs (approved):

PR #681, opened 2023-04-21
Fix advanced search

Fosol:
Resolving issues with the advanced search.  Also update functionality with the Status filter.  Still need to improve the Morning Report implementation as it's mixing with the Content List View.

image


.vscode/settings.json
```diff
@@ -21,5 +21,9 @@
     "Renci",
     "Swashbuckle"
   ],
-  "java.configuration.updateBuildConfiguration": "interactive"
+  "java.configuration.updateBuildConfiguration": "interactive",
+  "editor.codeActionsOnSave": {
```
Fosol, line 25: Added this to top VSCode performing it's own organize imports which is not compatible with what we have configured for TNO.  Some times I hate the linters.

app/editor/src/features/content/list-view/utils/makeFilter.ts
```diff
@@ -29,15 +29,11 @@ export const makeFilter = (
       ? moment(filter.startDate).toISOString()
       : setTimeFrame(filter.timeFrame as number)?.toISOString(),
     publishedEndOn: filter.endDate ? moment(filter.endDate).toISOString() : undefined,
-    fieldType: filter.searchTerm?.trim() !== '' ? filter.fieldType : undefined,
```
Fosol, line 32: I have no idea how this ever worked because the API ignores these values.

app/editor/src/features/content/tool-bar/sections/filter/AdvancedSearchSection.tsx
```diff
@@ -86,7 +89,36 @@ export const AdvancedSearchSection: React.FC<IAdvancedSearchSectionProps> = () =
               )}
             />
           </Show>
-          <Show visible={filterAdvanced.fieldType !== 'sourceIds'}>
+          <Show visible={filterAdvanced.fieldType === advancedSearchKeys.Status}>
```
Fosol, line 92: We now have better UI for status options.  I need to do the dates at some point.

PR #682, opened 2023-04-21
Fix morning papers filter

Fosol:
Extracted the shared advanced filter so that the Morning Papers works correctly.

image

### DIFF
--- a/app/editor/src/features/content/morning-reports/utils/makeFilter.ts
+++ b/app/editor/src/features/content/morning-reports/utils/makeFilter.ts
@@ -27,14 +27,10 @@ export const makeFilter = (
     sourceIds: filter.sourceIds,
     publishedStartOn: setTimeFrame(filter.timeFrame as number)?.toISOString(),
     actions: applyActions(filter),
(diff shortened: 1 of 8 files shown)


app/editor/src/store/slices/content/contentSlice.ts
```diff
@@ -36,6 +36,11 @@ export const initialContentState: IContentState = {
     searchTerm: '',
   },
   filterMorningReports: defaultMorningReportsFilter(),
+  filterMorningReportAdvanced: {
```
Fosol, line 39: They were originally sharing an advanced filter which was causing all kinds of issues.

app/editor/src/features/content/morning-reports/utils/makeFilter.ts
```diff
@@ -27,14 +27,10 @@ export const makeFilter = (
     sourceIds: filter.sourceIds,
     publishedStartOn: setTimeFrame(filter.timeFrame as number)?.toISOString(),
     actions: applyActions(filter),
-    fieldType: filter.searchTerm?.trim() !== '' ? filter.fieldType : undefined,
```
Fosol, line 30: This would have never worked because the API ignores these values.

PR #687, opened 2023-04-25
Add subscriber pipeline

jdtoombs:
Having trouble testing this.. this is the only difference b/w the subscriber and editor I could find..

### DIFF
--- a/openshift/kustomize/tekton/base/triggers/listener.yaml
+++ b/openshift/kustomize/tekton/base/triggers/listener.yaml
@@ -28,6 +28,7 @@ spec:
                 || body.head_commit.modified.exists(m, m.startsWith('app/editor'))
                 || body.head_commit.removed.exists(m, m.startsWith('app/editor'))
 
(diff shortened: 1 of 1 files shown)


openshift/templates/tekton/pipelines/subscriber.yaml
```diff
@@ -0,0 +1,91 @@
+apiVersion: tekton.dev/v1beta1
```
Fosol, line 1: We already have a `buildah-subscriber` pipeline.  What are you hoping to improve with this pipeline?  Additionally, we have a `buildah-main` pipeline that support automation through triggers that will build and deploy both the editor and subscriber when required.  

The file that controls the trigger is `/openshift/kustomize/tekton/triggers/listener.yaml`.  It looks like it may be missing the `added` check for subscriber content.  I'm not sure why it doesn't run most of the time though.

You can test all these things by temporarily creating a copy of all of these objects and running them against your own fork.

Fosol (approved):

PR #690, opened 2023-04-27
TNO-1256 Fix DevOps

Fosol:
Updating infrastructure as code to resolve issues while setting up the TEST environment.


openshift/kustomize/api/base/deploy.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
     managed-by: kustomize
     created-by: jeremy.foster
 spec:
-  replicas: 3
+  replicas: 1
```
Fosol, line 18: Defaulting to 1 because we don't have enough resources to run 3 anyway.

openshift/kustomize/services/capture/base/kustomization.yaml
```diff
@@ -5,6 +5,7 @@ kind: Kustomization
 resources:
   - pvc.yaml
   - config-map.yaml
+  - services-config-map.yaml
```
Fosol, line 8: Adding shared config to all services.  Would like to come up with a way to not have to duplicate this file.

PR #694, opened 2023-05-01
TNO-1221 Add Commentary Default Expiry Value

Fosol:
Weekdays the commentary expiry value will default to 3 hours and weekends will be 5 hours.

> I still need to integrate with Canada's free API for holidays so that it will default to 5 hours on holidays.

image

### DIFF
--- a/app/editor/src/features/content/form/ContentActions.tsx
+++ b/app/editor/src/features/content/form/ContentActions.tsx
@@ -1,4 +1,5 @@
 import { getIn, useFormikContext } from 'formik';
+import moment from 'moment';
 import React from 'react';
 import { FaHourglassHalf } from 'react-icons/fa';
 import { useLookup } from 'store/hooks';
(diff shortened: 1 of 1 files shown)


app/editor/src/features/content/form/ContentActions.tsx
```diff
@@ -124,6 +125,10 @@ export const ContentActions: React.FC<IContentActionsProps> = ({
               onChange={(e) => {
                 const checked = e.currentTarget.checked;
                 if (!checked) setFieldValue(field('value', index), '');
+                else {
+                  const weekDay = moment(values.publishedOn).weekday();
+                  setFieldValue(field('value', index), `${weekDay === 0 || weekDay === 6 ? 3 : 5}`);
```
Fosol, line 130: One issue with this logic is for those who have their local configuration set to have Monday as the first day of the week.  I don't think that will be relevant for editors however.

PR #697, opened 2023-05-02
TNO-1077 Add headline to work order

Fosol:
All future work orders for transcripts or NLP will include the headline so that it can be displayed without making an additional query.

You can now click on the headline and it will load the content.

> Existing work orders will not have the value regrettably.

image

### DIFF
--- a/app/editor/src/features/admin/work-orders/constants/columns.tsx
+++ b/app/editor/src/features/admin/work-orders/constants/columns.tsx
@@ -14,7 +14,20 @@ export const getColumns = (
     label: 'Content',
     name: 'configuration',
     width: 4,
-    cell: (cell) => <CellEllipsis>{JSON.stringify(cell.original.configuration)}</CellEllipsis>,
+    cell: (cell) => (
+      <CellEllipsis
+        className="link"
+        onClick={(e) => {
+          if (cell.original.configuration.contentId) {
+            e.preventDefault();
+            e.stopPropagation();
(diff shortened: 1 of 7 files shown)


app/editor/src/features/admin/work-orders/constants/columns.tsx
```diff
@@ -14,7 +14,20 @@ export const getColumns = (
     label: 'Content',
     name: 'configuration',
     width: 4,
-    cell: (cell) => <CellEllipsis>{JSON.stringify(cell.original.configuration)}</CellEllipsis>,
+    cell: (cell) => (
+      <CellEllipsis
+        className="link"
+        onClick={(e) => {
+          if (cell.original.configuration.contentId) {
+            e.preventDefault();
```
Fosol, line 22: Don't want it to load the work order form, we want it to go directly to the content.

app/editor/src/features/content/form/utils/getDefaultCommentaryExpiryValue.ts
```diff
@@ -15,7 +15,6 @@ import { isHoliday } from '.';
 export const getDefaultCommentaryExpiryValue = (date: Date | string, holidays: IHolidayModel[]) => {
   const value = moment(date);
   const weekDay = value.weekday();
-  console.debug(weekDay);
```
Fosol, line 18: Oops

PR #698, opened 2023-05-02
Fix workorder table column

Fosol:
Updated style so that the column would not overflow.

image


app/editor/src/features/admin/work-orders/styled/WorkOrderList.tsx
```diff
@@ -22,4 +22,8 @@ export const WorkOrderList = styled(FormPage)`
       color: ${(props) => props.theme.css.primaryLightColor};
     }
   }
+
+  div.column {
+    overflow: hidden;
```
Fosol, line 27: Possibly we would want this as the default behavior of all columns, but I don't know.

PR #702, opened 2023-05-03
TNO-1021 Fix content owner

Fosol:
Fixing issue where the current owner of the content was not saved or stored in Elasticsearch.

Now the user who creates content will be the owner, or the user who updates content without a current owner will become the owner.

### DIFF
--- a/db/kafka/README.md
+++ b/db/kafka/README.md
@@ -29,3 +29,10 @@ Fetch all messages in the specified topic.
 docker exec -it tno-broker bash
 /bin/kafka-console-consumer --bootstrap-server localhost:29092 --topic test --from-beginning
 ```
+
+Create a topic with the broker command line.
+SSH into the Kafka broker pod.
+
+```bash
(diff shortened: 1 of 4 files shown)


db/kafka/README.md
```diff
@@ -29,3 +29,10 @@ Fetch all messages in the specified topic.
 docker exec -it tno-broker bash
 /bin/kafka-console-consumer --bootstrap-server localhost:29092 --topic test --from-beginning
 ```
+
+Create a topic with the broker command line.
+SSH into the Kafka broker pod.
+
+```bash
+/bin/kafka-topics --bootstrap-server kafka-broker-0.kafka-headless:9092,kafka-broker-1.kafka-headless:9092,kafka-broker-2.kafka-headless:9092 --topic reporting --create --partitions 3 --replication-factor 1
```
Fosol, line 37: I keep forgetting how to do this, so I've documented it.

PR #712, opened 2023-05-05
Minor fixes, and update table

jdtoombs:
Leaving as draft as I need to publish the package, but just want to make sure I don't mess you up @Fosol - will publish when I get the green light! 

Also let me know if the `groupBy` makes sense will add comments below.

This PR includes:
- commentary headlines will now truncate in commentary box
- new table implemented into home page
- small changes to table component
- add page number to section/page 
- update fonts

ex of table)
image


libs/npm/core/src/components/table/interfaces/TableInternal.ts
```diff
@@ -68,11 +68,13 @@ export class TableInternal<T extends object> implements ITableInternal<T> {
 
     this.rows = sortRows(rows, this.sortOrder);
     if (this.groupBy) {
-      this.groups = groupBy(this.rows, (row) => (row.original as any)['value']).map((group) => ({
```
jdtoombs, line 71: 'value' was hardcode here so grouping was not previously working; however, works now by passing a string of the key name to the groupBy property. Let me know if this is ok

jdtoombs:
@Fosol - ready for review!

app/subscriber/src/features/home/constants/columns.tsx
```diff
@@ -1,33 +1,25 @@
...
-export const columns: Column<any>[] = [
+export const columns: ITableHookColumn<IContentModel>[] = [
   {
-    id: 'checkbox',
-    Header: () => <Checkbox />,
-    Cell: () => <Checkbox />,
-    width: 0,
+    name: 'tone',
+    label: 'TONE',
+    cell: (cell: any) => <DetermineToneIcon tone={cell.original.tonePools[0]?.value ?? 0} />,
```
Fosol, line 9: You shouldn't need `any`

app/subscriber/src/features/home/constants/columns.tsx
```diff
@@ -1,33 +1,25 @@
...
-  },
-  {
-    id: 'description',
-    Header: 'SECTION PAGE',
-    accessor: 'section',
-    Cell: (cell) => <div className="section">{cell?.value}</div>,
-    width: 30,
+    name: 'sectionPage',
+    label: 'SECTION PAGE',
+    cell: (cell: any) => (
```
Fosol, line 20: You shouldn't need `any`

app/subscriber/src/features/home/constants/columns.tsx
```diff
@@ -1,33 +1,25 @@
...
   },
   {
-    id: 'tone',
-    Header: 'TONE',
-    accessor: 'tonePools',
-    Cell: (cell) => <DetermineToneIcon tone={cell.value ? cell.value[0]?.value : 0} />,
-    width: 20,
+    name: 'headline',
+    label: 'HEADLINE',
+    cell: (cell) => <div className="headline">{cell?.original.headline}</div>,
```
Fosol, line 14: I don't think `cell` would ever be null

libs/npm/core/src/components/table/interfaces/ITableHookGrouping.ts
```diff
@@ -2,6 +2,6 @@ import { ITableGroup } from '..';
 import { ITableInternalRow } from '.';
 
 export interface ITableHookGrouping<T extends object> {
-  groupBy?: (item: T, index: number, array: T[]) => string;
```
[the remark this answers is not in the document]
Fosol: This is a predicate.  It's expecting a function.  You can use it by passing the following parameter in your FlexboxTable `groupBy={(item) => item.prop1}`.  If we wanted to also support a string we should write code to handle both scenarios.  This means groups can be a composite of property values.  `groupBy={(item) => item.prop1 + item.prop2}`

libs/npm/core/src/components/table/interfaces/TableInternal.ts
Fosol: I think the hardcoded `value` is incorrect and due to my test data having the property.  If you instead use `this.groupBy(row.original)` what happens?

Fosol (changes requested):

libs/npm/core/src/components/table/interfaces/TableInternal.ts
jdtoombs: Hmm seems to be having a few issues, will try to sort it

jdtoombs:
@Fosol updated once again just so you know!

app/editor/tno-core-0.0.241.tgz
Fosol, line 1: DELETE

app/subscriber/tno-core-0.0.241.tgz
Fosol, line 1: DELETE

app/subscriber/tno-core-0.0.242.tgz
Fosol, line 1: DELETE

Fosol (changes requested):

Fosol (approved):

PR #714, opened 2023-05-08
TNO-93 Add Report Service

Fosol:
The Report Service is a Kafka Consumer.  It waits for a request to generate a report.  After generating the report it sends the mail with CHES.

Also fixed the FlexboxTable component `groupBy` functionality.

> Published `tno-core:0.0.243`

We now have a very basic admin page for reports to perform basic testing.

image


api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -122,21 +122,21 @@ public IActionResult FindWithDatabase()
 
     /// <summary>
     /// Find a page of content for the specified query filter.
+    /// TODO: The model stored in Elasticsearch is a little confusing based on the controller using it.  Need to clean up.
     /// </summary>
     /// <returns></returns>
     [HttpGet]
     [Produces(MediaTypeNames.Application.Json)]
-    [ProducesResponseType(typeof(IPaged<ContentModel>), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(IPaged<Services.Models.Content.ContentModel>), (int)HttpStatusCode.OK)]
```
Fosol, line 130: We index content into Elasticsearch with the Indexing Service.  This service uses the above model.  Which means we generally should be using the same model to extract content from Elasticsearch.

app/editor/src/features/demo/DemoPage.tsx
```diff
@@ -132,7 +136,8 @@ export const DemoPage: React.FC = () => {
         // showFilter={true}
         // scrollSize="200px"
         // showHeader={false}
-        // groupBy={(item) => item.section}
+        // groupBy={(item) => item.original.section ?? 'any'}
```
Fosol, line 139: @jdtoombs It will now support both a predicate function or a property name string.

libs/net/dal/Services/ReportService.cs
```diff
@@ -61,14 +63,24 @@ public IEnumerable<Report> FindAll()
...
     {
-        var response = await _client.SearchAsync<Content>(s =>
+        var response = await _client.SearchAsync<API.Areas.Services.Models.Content.ContentModel>(s =>
         {
+            var json = report.Filter.ToJson();
             var result = s
                 .Pretty()
                 .Index(_elasticOptions.PublishedIndex) // TODO: Switch to unpublished if the report settings require it.
-                .Query(q => q.Raw(report.Filter.ToJson()));
+                .Query(q => q.Raw(json == "{}" ? "" : json));
```
Fosol, line 74: I really don't like the NEST library.  It doesn't allow us to send raw JSON requests to Elasticsearch.  I want to replace it soon so that we don't have the stupid issues the library is causing.

libs/npm/core/src/components/table/interfaces/ITableHookGrouping.ts
```diff
@@ -2,6 +2,8 @@ import { ITableGroup } from '..';
 import { ITableInternalRow } from '.';
 
 export interface ITableHookGrouping<T extends object> {
-  groupBy?: (item: T, index: number, array: T[]) => string;
+  groupBy?:
+    | keyof T
```
Fosol, line 6: Handles both the predicate and property name now.

PR #716, opened 2023-05-08
TNO-1239 Init topic score rules

Fosol:
Prepopulating each environment with the topic score rules and source friendly names.

Added new DB migration `1.0.14`.

> Run `make db-update` to get the latest.


libs/net/dal/Migrations/1.0.14/Up/PostUp/02-Source.sql
```diff
@@ -0,0 +1,162 @@
+DO $$
+BEGIN
+
+UPDATE public."source" SET
```
Fosol, line 4: One downside to this implementation is if the source is missing it won't be added/updated.

PR #723, opened 2023-05-10
Update action buttons

jdtoombs:
Add new hover states, tooltips, and active sate to the action toolbar.

actionbuttons

### DIFF
--- a/app/subscriber/src/App.tsx
+++ b/app/subscriber/src/App.tsx
@@ -46,7 +46,16 @@ function App() {
         </LayoutAnonymous>
       </Show>
       <ToastContainer />
-      <Tooltip variant="light" id="main-tooltip" place="top" />
+      <Tooltip
+        style={{
+          backgroundColor: '#FFFFCC',
+          color: 'black',
+          opacity: '1',
+          boxShadow: '0 0 8px #464545',
+        }}
+        id="main-tooltip"
+        place="top"
+      />
       <Tooltip variant="light" id="main-tooltip-right" place="right" />
     </BrowserRouter>
   );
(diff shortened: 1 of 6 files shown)


app/subscriber/src/App.tsx
```diff
@@ -46,7 +46,16 @@ function App() {
         </LayoutAnonymous>
       </Show>
       <ToastContainer />
-      <Tooltip variant="light" id="main-tooltip" place="top" />
+      <Tooltip
+        style={{
+          backgroundColor: '#FFFFCC',
```
jdtoombs, line 51: In line style unfortunately as per the docs suggested. Couldn't access from styled component.

app/subscriber/src/features/home/Home.tsx
```diff
@@ -32,8 +32,7 @@ export const Home: React.FC = () => {
             contentTypes:
               filter.contentTypes.length > 0 ? filter.contentTypes : [ContentTypeName.PrintContent],
             startDate: filter.startDate ? filter.startDate : new Date().toDateString(),
-            // TODO: For testing purposes leaving this commented out
-            // status: ContentStatus.Published,
+            status: ContentStatus.Published,
```
jdtoombs, line 35: Also TNO-1262 (only show published content)

Fosol: This will 'technically' filter on published content, but it isn't the correct implementation.  Subscribers should by default only be searching the index that only contains published content.  This will ensure they receive the best performance as it will have far less content in it than the unpublished index.  You don't need to fix in this PR.  Don't close TNO-1262 until the `ContentService.FindWithElasticsearchAsync(...)` has an override to use the correct index.

Fosol (changes requested):

app/subscriber/src/features/home/Home.tsx
jdtoombs: Sounds good, will create another PR for 1262

Fosol (approved):

PR #731, opened 2023-05-12
TNO-474 Add razor engine to reports

Fosol:
The Reporting Service will now use the Razor engine for templating.

## Summary

- Add Razor engine to Reporting Service
- Update Notification Service
- Publish tno-core:0.0.246
- Update yup in package
- Add model serialization validation to help debug Content Service


libs/net/dal/Migrations/1.0.16/Up/PostUp/00-NotificationRazorTemplate.sql
```diff
@@ -0,0 +1,80 @@
+DO $$
```
Fosol, line 1: I don't have a DB migration in yet, but the template originally provided was mapped to the option `Value`.  This is an incorrect implementation.

libs/net/models/Areas/Services/Models/Content/SourceModel.cs
```diff
@@ -0,0 +1,112 @@
+
+using TNO.API.Models;
+using System.Text.Json;
+
+namespace TNO.API.Areas.Services.Models.Content;
+
+/// <summary>
+/// SourceModel class, provides a model that represents an source.
+/// </summary>
+public class SourceModel : BaseTypeModel<int>
```
Fosol, line 10: The Services model was sharing a model which isn't ideal.

services/net/notification/NotificationManager.cs
```diff
@@ -93,8 +97,7 @@ public class NotificationManager : ServiceManager<NotificationOptions>
         this.Listener.IsLongRunningJob = true;
         this.Listener.OnError += ListenerErrorHandler;
         this.Listener.OnStop += ListenerStopHandler;
-        _engine = engine;
-        _notificationOptions = notificationOptions;
```
Fosol, line 97: @CelloLiu Options are already part of the base class.  You don't need to duplicate them.  Additionally, you should extract the `notificationOptions.Value` in the constructor based on examples from Microsoft.

services/net/notification/NotificationService.cs
```diff
@@ -47,7 +49,11 @@ protected override IServiceCollection ConfigureServices(IServiceCollection servi
             .AddChesSingletonService(this.Configuration.GetSection("CHES"))
             .AddSingleton(new ClaimsPrincipal(new ClaimsIdentity(new[] { new Claim(ClaimTypes.Email, "") })))
             .AddSingleton<IServiceManager, NotificationManager>()
-            .AddSingleton<ITnoRazorLightEngine, TnoRazorLightEngine>();
+            .AddSingleton<IRazorLightEngine>(
```
Fosol, line 52: We don't currently need a wrapper class that only provides the basic initialization code.  We can do that here.

services/net/notification/TnoRazorLightEngine.cs
Fosol, line 1: Not needed as it didn't provide any additional functionality.

PR #733, opened 2023-05-15
TNO-474 Debug Content Service Issue

Fosol:
Adding additional error log information to debug Content Service errors.

### DIFF
--- a/api/net/Areas/Services/Controllers/ContentController.cs
+++ b/api/net/Areas/Services/Controllers/ContentController.cs
@@ -212,7 +212,7 @@ public class ContentController : ControllerBase
 
             await _kafkaMessenger.SendMessageAsync(_kafkaHubOptions.HubTopic, new KafkaHubMessage(HubEvent.SendAll, new InvocationMessage("Content", new[] { new ContentMessageModel(content) })));
         }
-        else
+        else if (index)
             _logger.LogWarning("Kafka indexing topic not configured.");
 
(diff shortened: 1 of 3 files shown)


api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -212,7 +212,7 @@ public async Task<IActionResult> UpdateAsync(ContentModel model, bool index = fa
 
             await _kafkaMessenger.SendMessageAsync(_kafkaHubOptions.HubTopic, new KafkaHubMessage(HubEvent.SendAll, new InvocationMessage("Content", new[] { new ContentMessageModel(content) })));
         }
-        else
+        else if (index)
```
Fosol, line 215: The Content Service doesn't always index content, this error shouldn't be displayed when indexing isn't requested.

libs/net/services/Helpers/ApiService.cs
```diff
@@ -112,7 +112,7 @@ protected async Task<T> RetryRequestAsync<T>(Func<Task<T>> callbackDelegate)
             }
 
             // Wait before retrying.
-            this.Logger.LogError(ex, "Retry attempt {count}", this.FailureCount);
+            this.Logger.LogError(ex, "Retry attempt {count}.{newline}Error:{body}", this.FailureCount, Environment.NewLine, ex.Data["Body"]);
```
Fosol, line 115: Add the error information found in the response.

PR #736, opened 2023-05-15
TNO-474 Update report cache

Fosol:
The Razor engine is setup to cache the templates after they are compiled and rendered for the first time.  Without caching it would be slow to recompile each time it ran.  To overwrite cache I've added a flag to control this behaviour.


services/net/content/ContentManager.cs
```diff
@@ -251,7 +251,7 @@ private async Task HandleMessageAsync(ConsumeResult<string, SourceContent> resul
                 OwnerId = model.RequestedById ?? source?.OwnerId,
                 Headline = String.IsNullOrWhiteSpace(model.Title) ? "[TBD]" : model.Title,
                 Uid = model.Uid,
-                Page = model.Page,
+                Page = model.Page[0..Math.Min(model.Page.Length, 10)], // TODO: Temporary workaround to deal FileMonitor Service.
```
Fosol, line 254: Resolves a minor issue with some page values.

PR #737, opened 2023-05-15
TNO-474 Fix Report/Notification Template

Fosol:
Now when we update a report/notification template it will generate a new cached parsing template.

> Please note that this design has a memory leak due to RazorLight being poorly designed.  I've created a story to resolve this issue.  I have also reviewed Fluid templating engine.  However that engine doesn't support grouping (I couldn't figure out how to get it to work).

### DIFF
--- a/services/net/notification/NotificationManager.cs
+++ b/services/net/notification/NotificationManager.cs
@@ -390,13 +390,16 @@ public class NotificationManager : ServiceManager<NotificationOptions>
     /// <returns></returns>
     private async Task<string> GenerateNotificationSubjectAsync(NotificationModel notification, ContentModel content, bool updateCache = false)
     {
-        var key = $"notification_{notification.Id}_subject";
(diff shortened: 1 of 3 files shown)


services/net/notification/NotificationManager.cs
```diff
@@ -390,13 +390,16 @@ private async Task HandleChesEmailOverrideAsync(NotificationRequestModel request
     /// <returns></returns>
     private async Task<string> GenerateNotificationSubjectAsync(NotificationModel notification, ContentModel content, bool updateCache = false)
     {
-        var key = $"notification_{notification.Id}_subject";
-        var cache = this.RazorEngine.Handler.Cache.RetrieveTemplate(key);
+        // TODO: Having a key for every version is a memory leak, but the RazorLight library is junk and has no way to invalidate a cached item.
+        var key = $"notification_{notification.Id}-{notification.Version}_subject";
```
Fosol, line 394: Added `Version` to ensure a new cached item is built.

PR #738, opened 2023-05-16
Fix CP Filter

jdtoombs:
When 304 was returned sources would consist of an empty array, need to get from local storage when that is the case.

### DIFF
--- a/app/subscriber/src/features/home/home-filters/HomeFilters.tsx
+++ b/app/subscriber/src/features/home/home-filters/HomeFilters.tsx
@@ -1,7 +1,14 @@
 import React from 'react';
 import { useContent } from 'store/hooks';
 import { useSources } from 'store/hooks/admin';
-import { Button, ButtonHeight, ButtonVariant, ContentTypeName } from 'tno-core';
+import {
+  Button,
+  ButtonHeight,
+  ButtonVariant,
+  ContentTypeName,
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/home/home-filters/HomeFilters.tsx
```diff
@@ -18,6 +25,17 @@ export const HomeFilters: React.FC<IHomeFilterProps> = () => {
   const [{ filter }, { storeFilter }] = useContent();
   const [{ sources }] = useSources();
 
+  const [sourcesLookup, setSourcesLookup] = React.useState<ISourceModel[]>([]);
+
+  // if 304 is returned we need to get it from local storage
```
Fosol, line 30: I don't understand this.  304 means the content in the endpoint hasn't changed.   Which means you already have the latest content in cache.

You should have a lookup hook like we have for the editor app.  The difference is we don't want all the users being returned in the lookups.

jdtoombs: For some reason it was returning a blank array whenever it was 304, I will try out the lookup hook and maybe update the endpoints and see what happens

jdtoombs:
Having make pack issues....

jdtoombs:
Closing as branch got messy

PR #740, opened 2023-05-17
Published Index

jdtoombs:
Let me know if this makes sense @Fosol, can also make a separate service for the Subscriber App that only hits the published index if that is preferred

### DIFF
--- a/api/net/Areas/Editor/Controllers/ContentController.cs
+++ b/api/net/Areas/Editor/Controllers/ContentController.cs
@@ -134,7 +134,7 @@ public class ContentController : ControllerBase
         var uri = new Uri(this.Request.GetDisplayUrl());
         var query = Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery(uri.Query);
         var filter = new ContentFilter(query);
-        var result = await _contentService.FindWithElasticsearchAsync(filter);
+        var result = await _contentService.FindWithElasticsearchAsync("unpublished", filter);
         var page = new Paged<Services.Models.Content.ContentModel>(
             result.Items,
             result.Page,
--- a/api/net/Areas/Editor/Controllers/ContentController.cs
+++ b/api/net/Areas/Editor/Controllers/ContentController.cs
@@ -24,6 +24,7 @@ using TNO.API.SignalR;
 using Microsoft.AspNetCore.SignalR;
 using TNO.Keycloak;
 using TNO.Core.Exceptions;
+using TNO.Elastic;
 
 namespace TNO.API.Areas.Editor.Controllers;
 
@@ -55,6 +56,8 @@ public class ContentController : ControllerBase
     private readonly KafkaOptions _kafkaOptions;
     private readonly JsonSerializerOptions _serializerOptions;
     private readonly ILogger _logger;
+    private readonly ElasticOptions _elasticOptions;
+
     #endregion
 
     #region Constructors
@@ -71,6 +74,7 @@ public class ContentController : ControllerBase
     /// <param name="storageOptions"></param>
     /// <param name="kafkaMessenger"></param>
     /// <param name="kafkaOptions"></param>
+    /// <param name="elasticOptions"></param>
     /// <param name="serializerOptions"></param>
     /// <param name="logger"></param>
     public ContentController(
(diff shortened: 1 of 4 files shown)


libs/net/dal/Services/ContentService.cs
```diff
@@ -314,9 +314,14 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
 
         var response = await _client.SearchAsync<API.Areas.Services.Models.Content.ContentModel>(s =>
         {
+            
+            if(filter.Status == ContentStatus.Publish)
```
Fosol, line 318: Hmm interesting solution.  I kinda like it because it's simple.  But I don't think it's wise.  Let me think about this tonight.  I think there are some simple ways we can do this and control the behaviour in a more intuitive way.

Fosol: Add an argument to the `ContentService.FindWithElasticsearchAsync(string index, ContentFilter filter)`.  Then each place it is used it will specify the index with the configuration option `ElasticOptions.UnpublishedIndex` or `ElasticOptions.PublishedIndex`

Fosol (changes requested):

jdtoombs:
@Fosol - this has been updated just so ya know!

api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -134,7 +134,7 @@ public async Task<IActionResult> FindWithElasticsearchAsync()
         var uri = new Uri(this.Request.GetDisplayUrl());
         var query = Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery(uri.Query);
         var filter = new ContentFilter(query);
-        var result = await _contentService.FindWithElasticsearchAsync(filter);
+        var result = await _contentService.FindWithElasticsearchAsync("unpublished", filter);
```
Fosol, line 137: Don't use magic strings.  replace with `_elasticOptions.UnpublishedIndex`

api/net/Areas/Subscriber/Controllers/ContentController.cs
```diff
@@ -113,7 +113,7 @@ public async Task<IActionResult> FindWithElasticsearchAsync()
         var uri = new Uri(this.Request.GetDisplayUrl());
         var query = Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery(uri.Query);
         var filter = new ContentFilter(query);
-        var result = await _contentService.FindWithElasticsearchAsync(filter);
+        var result = await _contentService.FindWithElasticsearchAsync("published", filter);
```
Fosol, line 116: Don't use magic strings.  replace with `_elasticOptions.PublishedIndex`

libs/net/dal/Services/ContentService.cs
```diff
@@ -315,9 +315,13 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
 
         var response = await _client.SearchAsync<API.Areas.Services.Models.Content.ContentModel>(s =>
         {
+            if(index == "published")
```
Fosol, line 318: Remove magic string logic.

Fosol (changes requested):

libs/net/dal/Services/ContentService.cs
```diff
@@ -315,9 +315,13 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
 
         var response = await _client.SearchAsync<API.Areas.Services.Models.Content.ContentModel>(s =>
         {
+            if(index == _elasticOptions.PublishedIndex)
```
Fosol, line 318: remove

libs/net/dal/Services/ContentService.cs
```diff
@@ -315,9 +315,13 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
...
+                s.Index(_elasticOptions.PublishedIndex);
+            else
+                s.Index(_elasticOptions.UnpublishedIndex);
+
             var result = s
                 .Pretty()
-                .Index(_elasticOptions.UnpublishedIndex)
```
Fosol, line 320: Put this in `.Index(index)`

Fosol (changes requested):

libs/net/dal/Services/ContentService.cs
jdtoombs: d'oh

Fosol (approved):

PR #741, opened 2023-05-17
TNO-1208 Use elastic filter in reporting service

Fosol:
The Reporting Service now uses the elastic filter to fetch content.

I have also started the process of removing the NEST library.  We can submit raw JSON queries to elasticsearch with the new TNO.Elastic project.


api/net/Areas/Services/Controllers/ReportController.cs
```diff
@@ -84,17 +84,15 @@ public IActionResult FindById(int id)
     /// <returns></returns>
     [HttpGet("{id}/content")]
     [Produces(MediaTypeNames.Application.Json)]
-    [ProducesResponseType(typeof(IEnumerable<Models.Content.ContentModel>), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(Elastic.Models.SearchResultModel<API.Areas.Services.Models.Content.ContentModel>), (int)HttpStatusCode.OK)]
```
Fosol, line 87: We now return the Elastic results instead of extracting only the content.  This will enable future enhancements in reports to more easily align to elasticsearch.

libs/net/elastic/TNOElasticClient.cs
```diff
@@ -0,0 +1,85 @@
...
+using TNO.Core.Extensions;
+using TNO.Core.Http;
+using TNO.Elastic.Models;
+
+namespace TNO.Elastic
+{
+    /// <summary>
+    /// The TNOElasticClient class
+    /// </summary>
+    public class TNOElasticClient : ElasticClient, ITNOElasticClient
```
Fosol, line 17: Eventually we'll get rid of the NEST `ElasticClient` entirely.  For now we're making it so we can do both.

libs/net/elastic/TNOElasticClient.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        /// <summary>
+        /// Make a request to Elasticsearch 'index' with the specified 'query'.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="index"></param>
+        /// <param name="query"></param>
+        /// <returns></returns>
+        public async Task<SearchResultModel<T>> SearchAsync<T>(string index, JsonDocument query)
```
Fosol, line 75: This is the magic.  We can now submit raw JSON queries to Elasticsearch.  No longer do we need to hop through hoops with NEST.  I honestly don't understand why NEST is built to be so difficult to use.

PR #742, opened 2023-05-17
TNO-1037 Add columnist/pundit

Fosol:
We now have a way to manage columnist/pundits list values.  They have been added to the content form.

> Developers will need to run the `make renew` command to update their local environment.

## Summary

- Replace react-table
- Publish `tno-core:0.0.247`
- Added DB migration `1.0.16`
- Fixed notification template
- Converted all admin react-table components to the new component

image

image


app/editor/src/features/admin/ingests/IngestList.tsx
```diff
@@ -42,15 +39,31 @@ export const IngestList: React.FC<IIngestListProps> = (props) => {
...
-        columns={columns}
-        isLoading={!!requests.length}
-        sorting={{ sortBy: [{ id: 'id', desc: false }] }}
-        paging={{ pageSizeOptions: { fromLocalStorage: true } }}
+      <IngestFilter
+        setGlobalFilter={(filter) => {
+          if (filter && filter.length) {
+            const value = filter.toLocaleLowerCase();
+            setItems(
+              ingests.filter(
```
Fosol, line 47: Simple filter implementation.  I need to create a generic way to do this and apply it to the FlexboxTable at some point.

libs/npm/core/src/hooks/api-editor/interfaces/IContentModel.ts
```diff
@@ -25,12 +27,14 @@ export interface IContentModel extends IAuditColumnsModel {
   licenseId: number;
   license?: ILicenseModel;
   sourceId?: number;
-  source?: ISeriesModel;
+  source?: ISourceModel;
```
Fosol, line 30: Fixed an old bug

PR #743, opened 2023-05-18
Fix CP Filter

jdtoombs:
Very simple yet annoying fix..


app/subscriber/src/store/hooks/lookup/useLookup.ts
```diff
@@ -40,7 +41,6 @@ import {
   useApiTopicScoreRules,
   useApiUsers,
 } from 'tno-core';
-import { getFromLocalStorage } from 'utils';
```
jdtoombs, line 43: This was causing [] to be returned

Fosol (approved):

PR #749, opened 2023-05-23
TNO-1318 Add admin filters

Fosol:
Adding filters to all the admin pages.

image

### DIFF
--- a/app/editor/src/features/admin/actions/ActionFilter.tsx
+++ b/app/editor/src/features/admin/actions/ActionFilter.tsx
@@ -2,17 +2,17 @@ import React from 'react';
 import { IconButton, Row, Text } from 'tno-core';
 
 interface IAdminFilterProps {
-  setGlobalFilter: (filterValue: any) => void;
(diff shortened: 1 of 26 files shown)


app/editor/src/features/admin/actions/ActionFilter.tsx
```diff
@@ -2,17 +2,17 @@ import React from 'react';
...
-export const TagListFilter: React.FC<IAdminFilterProps> = ({ setGlobalFilter }) => {
+export const ActionFilter: React.FC<IAdminFilterProps> = ({ onFilterChange }) => {
   const [filter, setFilter] = React.useState<string>('');
   return (
     <Row className="filter-bar" justifyContent="center">
       <Text
         onChange={(e) => {
           setFilter(e.target.value);
-          setGlobalFilter(e.target.value);
+          onFilterChange(e.target.value);
```
Fosol, line 15: I tried to move this logic to a `React.useEffect` but it results in an infinite loop after entering text in the search field.

PR #750, opened 2023-05-23
TNO-1148: Build My Minister view and functionality 

jdtoombs:
tno-core update to `0.0.251`
db-migration `v1.0.19`

Includes:

- New My Minister section that allows a user to choose which minister they would like to "follow" (just one for now)
- Clicking the My Minister option will be a quick link to a filtered list of contents for said minister
- Moved several controllers to their own subscriber sections
- New column added to user table called preferences that contains a json object of preferences

Known bug (sidebar):

- on click only working when clicking on the label after allowing for more than one clickable area - ticket will be created

image


app/subscriber/src/features/my-minister/MyMinister.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const fetch = React.useCallback(
+    async (filter: IContentListFilter & Partial<IContentListAdvancedFilter>) => {
+      try {
+        setLoading(true);
+        const data = await findContent(
+          makeFilter({
+            ...filter,
+            startDate: '',
+            endDate: '',
+            keyword: localStorage.getItem('myMinister') ?? '',
```
jdtoombs, line 30: Would we want to store this on a new column in the users table instead?

app/subscriber/src/features/home/utils/IContentFilter.ts
```diff
@@ -0,0 +1,29 @@
+import { ContentStatus, ContentTypeName, ISortPageFilter, LogicalOperator } from 'tno-core';
```
jdtoombs, line 1: Moving these over to tno-core (others as well)

app/subscriber/src/components/sidebar/SelectableMenuItems.tsx
```diff
@@ -22,15 +23,26 @@ export const SelectableMenuItems: React.FC<ISelectableMenuItemProps> = () => {
...
-            }}
             className={selected === item.label ? 'selected' : ''}
             key={item.label}
             icon={item.icon}
           >
-            {item.label}
+            <Row className="label-container">
+              <div
+                onClick={() => {
+                  navigate(`landing/${item.path}` ?? '');
```
Fosol, line 33: Will this ever result in a relative path getting appended to the URL?

app/subscriber/src/features/my-minister/MyMinister.tsx
Fosol: We could add a `User.Preferences` property which would be JSONP like many of our configuration tables.

app/subscriber/src/features/my-minister/MyMinister.tsx
```diff
@@ -0,0 +1,67 @@
...
+    [findContent],
+  );
+
+  /** retrigger content fetch when change is applied */
+  React.useEffect(() => {
+    fetch({ ...filter, ...filterAdvanced });
+  }, [filter, filterAdvanced, fetch]);
+  return (
+    <styled.MyMinister>
+      {' '}
```
Fosol, line 50: Is this intentional?

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const MyMinisterSettings: React.FC = () => {
+  const [{ ministers }] = useLookup();
+  const [myMinister, setMyMinister] = React.useState<string>();
+
+  React.useEffect(() => {
+    if (!!myMinister) localStorage.setItem('myMinister', myMinister);
+  }, [myMinister]);
+
+  const options = ministers.map((m) => {
+    return {
```
Fosol, line 16: It's probably better to create an `OptionItem` rather than force casting it to a `IOptionItem`

app/subscriber/src/store/hooks/lookup/ILookupModel.ts
```diff
@@ -0,0 +1,40 @@
...
+  ITagModel,
+  ITonePoolModel,
+  ITopicModel,
+  ITopicScoreRuleModel,
+  IUserModel,
+} from 'tno-core';
+
+import { IMinisterModel } from '../subscriber/interfaces/IMinisterModel';
+
+export interface ILookupModel {
```
Fosol, line 22: This doesn't match what I saw in the API endpoint you created.

libs/net/dal/Services/ContentService.cs
```diff
@@ -223,8 +225,17 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
...
-        //     filterQueries.Add(s => s.Terms(t => t.Field(p => p.)));
+        if (!string.IsNullOrWhiteSpace(filter.Keyword))
+            filterQueries.Add(s => s.MultiMatch(m => m
+                .Fields(f => f
+                    .Field(p => p.Headline)
+                    .Field(p => p.Body)
+                    .Field(p => p.Byline)
+                    .Field(p => p.Contributor)
+                )
+                .Query(filter.Keyword.ToLower())
```
Fosol, line 236: Not sure `ToLower()` makes sense here.  All the field values would have to be lowercase too.  Is Elasticsearch case sensitive?

Fosol: I noticed Cello also did the same in the above queries.  Please test.

libs/net/models/Areas/Subscriber/Models/Content/ContentModel.cs
```diff
@@ -35,7 +35,7 @@ public class ContentModel : AuditColumnsModel
     /// <summary>
     /// get/set - The product.
     /// </summary>
-    public ProductModel? Product { get; set; }
+    public Product.ProductModel? Product { get; set; }
```
Fosol, line 38: Mixing models can lead to bugs.  It's generally better to create a copy of the model and place it in the convention location.  That way we know what models each controller is using.

libs/net/models/Areas/Subscriber/Models/Source/SourceModel.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// <summary>
+    /// get/set - Whether to show the topics on the content form.
+    /// </summary>
+    public bool UseInTopics { get; set; }
+
+    /// <summary>
+    /// get/set -
+    /// </summary>
+    public Dictionary<string, object> Configuration { get; set; } = new Dictionary<string, object>();
```
Fosol, line 84: I doubt the subscriber needs this property

Fosol (changes requested):

app/subscriber/src/features/my-minister/MyMinister.tsx
jdtoombs: Sounds like a good idea, will do!

libs/net/dal/Services/ContentService.cs
jdtoombs: Will do

libs/net/models/Areas/Subscriber/Models/Content/ContentModel.cs
jdtoombs: Good to know, will do

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -0,0 +1,112 @@
+using System.Net;
+using System.Net.Mime;
+using System.Text.Json;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
+using Swashbuckle.AspNetCore.Annotations;
+using TNO.API.Areas.Services.Models.Minister;
```
Fosol, line 8: Why does the subscriber UserController reference the services model?

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// Update the user in Keycloak if the 'Key' is linked.
+    /// </summary>
+    /// <param name="model"></param>
+    /// <returns></returns>
+    [HttpPut("{id}")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(UserModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "User" })]
+    public async Task<IActionResult> UpdateAsync(UserModel model)
```
Fosol, line 105: Why are we letting subscribers update users?

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -0,0 +1,112 @@
...
+    #region Endpoints
+    /// <summary>
+    /// Find a page of user for the specified query filter.
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(IPaged<UserModel>), (int)HttpStatusCode.OK)]
+    [SwaggerOperation(Tags = new[] { "User" })]
+    public IActionResult Find()
```
Fosol, line 65: What functionality are we adding where subscribers can search for users?

app/subscriber/src/features/my-minister/MyMinister.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const [{ filter, filterAdvanced }, { findContent }] = useContent();
+  const [homeItems, setHomeItems] = React.useState<IContentModel[]>([]);
+  const navigate = useNavigate();
+  const [user, setUser] = React.useState<IUserModel>(defaultUser);
+  const [{ userInfo }] = useApp();
+  const api = useUsers();
+
+  React.useEffect(() => {
+    if (userInfo && userInfo.id) {
+      api.getUser(userInfo.id).then((data) => {
```
Fosol, line 25: If you are fetching the current user's information you should be using the `/auth/userinfo` endpoint.  Is there something else you need?

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// Find user for the specified 'id'.
+    /// </summary>
+    /// <param name="id"></param>
+    /// <returns></returns>
+    [HttpGet("{id}")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(UserModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(string), (int)HttpStatusCode.NoContent)]
+    [SwaggerOperation(Tags = new[] { "User" })]
+    public IActionResult FindById(int id)
```
Fosol, line 84: Looking at your PR it appears you're using this endpoint to fetch the current user's info.  But you're opening the door for requesting every single user.  While it probably doesn't matter in our solution, this is essentially a security hole in nearly all other solutions.

libs/net/models/Areas/Services/Models/Minister/UserModel.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    /// <summary>
+    /// get/set - The last date and time when user logged in.
+    /// </summary>
+    public DateTime? LastLoginOn { get; set; }
+
+    /// <summary>
+    /// get/set - The preferences for this user.
+    /// </summary>
+    public Dictionary<string, object> Preferences { get; set; } = new Dictionary<string, object>();
```
Fosol, line 75: Is there a reason to send user preferences to services?

libs/net/models/Areas/Subscriber/Models/Content/ContentModel.cs
```diff
@@ -1,5 +1,6 @@
 using System.ComponentModel.DataAnnotations;
 using Microsoft.AspNetCore.Http;
+using TNO.API.Areas.Subscriber.Content;
```
Fosol, line 3: What's this reference for?

libs/npm/core/src/hooks/api-editor/interfaces/IUserInfoModel.ts
```diff
@@ -11,6 +11,7 @@ export interface IUserInfoModel {
   lastLoginOn?: Date;
   isEnabled: boolean;
   status: UserStatusName;
+  preferences?: any;
```
Fosol, line 14: This is currently incorrect.  The `IUserInfoModel` does not contain this property.

Fosol (changes requested):

api/net/Areas/Subscriber/Controllers/UserController.cs
jdtoombs: Needed in order to update the user preferences when selecting my minister

jdtoombs: (will only ever update themselves)

api/net/Areas/Subscriber/Controllers/UserController.cs
jdtoombs: Will remove

app/subscriber/src/features/my-minister/MyMinister.tsx
jdtoombs: This was in order to get the users "myMinister" preferences, maybe I can add this to userInfo?

libs/net/models/Areas/Subscriber/Models/Content/ContentModel.cs
jdtoombs: This is for some of the models like Product or License that have a copy in the `/Content` folder

api/net/Areas/Subscriber/Controllers/UserController.cs
Fosol: Add logic to only allow a user to update themselves.

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -54,43 +51,6 @@ public UserController(IUserService userService, ICssHelper cssHelper, IOptions<J
     #endregion
 
     #region Endpoints
-    /// <summary>
```
jdtoombs: Information will now be pulled from `UserInfo`

api/net/Areas/Subscriber/Controllers/UserController.cs
jdtoombs: Will do!! Was just going to ask if that was an ok work-around :)

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// <param name="model"></param>
+    /// <param name="requestorId"></param>
+
+    /// <returns></returns>
+    [HttpPut("{id}")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(UserModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "User" })]
+    public async Task<IActionResult> UpdateAsync(UserModel model, [FromQuery] int requestorId)
```
Fosol, line 68: This won't work.  I can just lie and submit a request with the same value in each.
The API always knows who the current user is because they submit a token.
Take a closer look at the `/auth/userinfo` endpoint for more information.

api/net/Areas/Subscriber/Models/UserModel.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    /// <summary>
+    /// get/set - Whether the user email is verified.
+    /// </summary>
+    public bool EmailVerified { get; set; }
+
+    /// <summary>
+    /// get/set - Whether the user is a system account.
+    /// </summary>
+    public bool IsSystemAccount { get; set; }
```
Fosol, line 65: Do not include any fields they should not be able to see or edit (IsEnabled, Status, EmailVerified, IsSystemAccount, LastLoginOn, Note, Roles).  Otherwise you're introducing security holes.

Fosol: This is why we use different models for each endpoint.

app/subscriber/src/store/hooks/subscriber/api/useApiUsers.ts
```diff
@@ -0,0 +1,35 @@
...
+    lifecycleToasts?: ILifecycleToasts;
+    selector?: Function;
+    envelope?: typeof defaultEnvelope;
+    baseURL?: string;
+  } = {},
+) => {
+  const api = useApi(options);
+
+  return React.useRef({
+    getUsers: (etag: string | undefined = undefined) => {
```
Fosol, line 20: Remove

app/subscriber/src/store/hooks/subscriber/api/useApiUsers.ts
```diff
@@ -0,0 +1,35 @@
...
+      const config = { headers: { 'If-None-Match': etag ?? '' } };
+      return api.get<IUserModel[], AxiosResponse<IUserModel[]>, any>(`/subscriber/users`, config);
+    },
+    getUser: (id: number) => {
```
Fosol, line 24: Remove

app/subscriber/src/store/hooks/subscriber/useUsers.ts
```diff
@@ -1,22 +1,40 @@
...
 }
 
 export const useUsers = (): IUserController => {
   const api = useApiUsers();
   const dispatch = useAjaxWrapper();
+  const [state, store] = useAdminStore();
 
   const controller = React.useMemo(
     () => ({
       getUser: async (id: number) => {
```
Fosol, line 20: Remove

Fosol (changes requested):

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+    #endregion
+
+    #region Endpoints
+    /// <summary>
+    /// Update user for the specified 'id'.
+    /// Update the user in Keycloak if the 'Key' is linked.
+    /// </summary>
+    /// <param name="model"></param>
+
```
Fosol, line 61: Remove extra line

api/net/Areas/Subscriber/Controllers/UserController.cs
```diff
@@ -0,0 +1,81 @@
...
+    public IActionResult Update(UserModel model)
+    {
+        var username = User.GetUsername() ?? throw new NotAuthorizedException("Username is missing");
+        var user = _userService.FindByUsername(username) ?? throw new NotAuthorizedException("User does not exist");
+
+        if(user.Id != model.Id)
+        {
+            throw new NotAuthorizedException("You are not authorized to update this user.");
+        }
+        var result = _userService.UpdateAndSave(model.ToEntity(_serializerOptions));
```
Fosol, line 77: This will result in bad behaviour regrettably.  The model will be converted to an entity and many fields will be left blank.  Then the UserService will overwrite values with blanks.  We also don't want users changing their names and other properties.

api/net/Areas/Subscriber/Models/UserModel.cs
```diff
@@ -0,0 +1,108 @@
...
+    #region Properties
+    /// <summary>
+    /// get/set - Primary key to user.
+    /// </summary>
+    public int Id { get; set; } = default!;
+
+    /// <summary>
+    /// get/set - Unique key to identify the user.
+    /// </summary>
+    public string Key { get; set; } = "";
```
Fosol, line 20: Remove.  Subscribers don't need to know their Keycloak unique key value.

api/net/Areas/Subscriber/Controllers/UserController.cs
Fosol: You really should just make this an endpoint that updates the user preferences rather than trying to make it a way to update a user.  Save yourself a lot of issues

Fosol (changes requested):

Fosol:
@jdtoombs You'll have to update your DB migration again.  Cello submitted another one and it's merged.

Fosol (approved):

PR #754, opened 2023-05-25
tno-1296 - baseline service contentmigration

kylgarmor:
- updated tno-core npm package to 0.0.249
- updated db with new migration, will require a make db-update


libs/net/dal/Migrations/1.0.17/Up/PostUp/00-Connection.sql
Fosol, line 1: Rollback and then update will fail.  Need to handle this.
Also inserting new items without validation on unique indexes will also result in failures.  Probably not an issue for our usecase, but something you should plan for.

libs/net/dal/Migrations/1.0.17/Up/PostUp/00-Connection.sql
```diff
@@ -0,0 +1,26 @@
...
+  , "is_read_only"
+  , "sort_order"
+  , "created_by"
+  , "updated_by"
+) VALUES (
+  'TNO 1.0 Database' -- 10
+  , 'Connection to TNO 1.0 Database. Only for Content Migration.' -- description
+  , true -- is_enabled
+  , 8 -- connection_type - Database Connection
+  , '{"hostname": "oracle_database", "port":1521, "sid": "", "password": "", "username": ""}' -- configuration
```
Fosol, line 19: Thanks for not including secrets

Fosol (changes requested):

Fosol (approved):

PR #759, opened 2023-05-27
TNO-1304: Table hover bug

jdtoombs:
**Blocked by #750 as I will have to update the NPM package but putting this up for review.**

Seems to have been some CSS that we don't need but wanted to confirm with you @Fosol. Rows still signify hover over but got rid of that weird behaviour from hovering the Group Title.


app/subscriber/tno-core-0.0.249.tgz
Fosol, line 1: Delete

app/subscriber/package.json
```diff
@@ -52,7 +52,7 @@
     "react-tooltip": "5.10.0",
     "redux-logger": "3.0.6",
     "styled-components": "5.3.9",
-    "tno-core": "0.0.249"
+    "tno-core": "./tno-core-0.0.249.tgz"
```
Fosol, line 55: Fix

Fosol (changes requested):
I think you're waiting until your other PR is merged before publishing a new package.

jdtoombs:
@Fosol - updated

Fosol (approved):

PR #761, opened 2023-05-27
TNO-324: Add WYSIWYG and HTML for alerts

jdtoombs:
- WYSIWYG has been added to the alerts admin section in the editor app
- both the subscriber and editor application will display the alert message as parsed HTML 

image

image


app/editor/package.json
```diff
@@ -25,6 +25,7 @@
     "dequal": "2.0.3",
     "formik": "2.2.9",
     "history": "5.3.0",
+    "html-react-parser": "^3.0.16",
```
Fosol, line 28: We don't use a range in the app because it results in different versions for different builds.

Fosol (approved):

PR #769, opened 2023-05-30
TNO-1289: Page/section should only show on print

jdtoombs:
- Section/page will now only show on print content as intended

### DIFF
--- a/app/subscriber/src/features/home/constants/columns.tsx
+++ b/app/subscriber/src/features/home/constants/columns.tsx
@@ -3,51 +3,36 @@ import { ContentTypeName, IContentModel, ITableHookColumn } from 'tno-core';
 import { DetermineToneIcon } from '../utils';
 
 export const determinecolumns = (contentType: ContentTypeName) => {
-  let cols: ITableHookColumn<IContentModel>[] = [];
+  // columns common to all content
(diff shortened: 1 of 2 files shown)


app/subscriber/src/features/home/constants/columns.tsx
```diff
@@ -1,27 +1,53 @@
...
-    label: 'SECTION PAGE',
-    cell: (cell) => (
-      <div className="section">{`${cell.original.section}: ${cell?.original.page}`}</div>
-    ),
-    width: 1,
-  },
-];
+export const determinecolumns = (contentType: ContentTypeName) => {
+  let cols: ITableHookColumn<IContentModel>[] = [];
+  if (contentType === ContentTypeName.PrintContent) {
```
Fosol, line 7: Instead of duplicating the shared columns, just add the additional column when required.  It's easier to maintain.

Fosol (changes requested):

app/subscriber/src/features/home/constants/columns.tsx
jdtoombs: Updated!

Fosol (approved):

PR #772, opened 2023-05-31
TNO-1290: Display tags on subscriber site

jdtoombs:
Adds the selected tags to the given piece of content on the subscriber site,
image

### DIFF
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -82,7 +82,7 @@ export const ViewContent: React.FC = () => {
 
   return (
     <styled.ViewContent>
-      <ViewContentToolbar />
+      <ViewContentToolbar currentContent={content && content} />
       <Row className="headline-container">
         <p>{content?.headline && content.headline}</p>
         <Row alignItems="center">
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -82,7 +82,7 @@ export const ViewContent: React.FC = () => {
 
   return (
     <styled.ViewContent>
-      <ViewContentToolbar currentContent={content && content} />
+      <ViewContentToolbar tags={content?.tags ?? []} />
       <Row className="headline-container">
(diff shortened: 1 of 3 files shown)


app/subscriber/src/features/content/view-content/ViewContentToolbar.tsx
```diff
@@ -63,7 +68,7 @@ export const ViewContentToolbar: React.FC = () => {
             />
           </div>
         </Row>
-        <Tags />
+        <Tags currentContent={currentContent && currentContent} />
```
Fosol, line 71: Not sure what the logic here means?

app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -81,7 +81,7 @@ export const ViewContent: React.FC = () => {
 
   return (
     <styled.ViewContent>
-      <ViewContentToolbar />
+      <ViewContentToolbar currentContent={content && content} />
```
Fosol, line 84: Not sure what the logic here means?

jdtoombs: Passing down the content that is being viewed so the toolbar has access to actions, tags, etc. Have to ensure that content is defined before doing so,  otherwise it will error out

Fosol: This doesn't make sense.   Sounds like you're simply writing code to bypass the linter.  Think about the runtime process.  It's going to pass `undefined` if `content is falsy`

Fosol (changes requested):

app/subscriber/src/features/content/view-content/ViewContent.tsx
jdtoombs: I believe if content is falsy the right hand side of the expression will not be passed. Will double check

Fosol (approved):

app/subscriber/src/features/content/view-content/ViewContent.tsx
jdtoombs: May have just been a bad approach, I was expecting content to be potentially undefined depending on the fetch being executed, and the prop to be updated when that content was defined. Updated the solution.

PR #774, opened 2023-05-31
Tno-1323

kylgarmor:
Initial setup for accessing a local copy/extract of the TNO database
Only the NEWS_ITEMS table has been implemented
Initial access to NEWS_ITEMS table via ContentMigration service


db/oracle/docker/scripts/04_data.sql
Fosol, line 1: @kylgarmor Can you confirm there is nothing private or secret in this file?

libs/net/services/Helpers/IngestManagerFactory`.cs
```diff
@@ -35,7 +35,7 @@ public IngestManagerFactory(IServiceProvider serviceProvider)
     /// </summary>
     /// <param name="ingest"></param>
```
Fosol, line 36: Missing parameter in docs.  If you do a `dotnet build` it should show you a number of linting issues.

libs/net/services/Managers/IngestManager`.cs
```diff
@@ -35,13 +37,15 @@ public abstract class IngestManager<TIngestServiceActionManager, TOption> : Serv
     /// <param name="options">Configuration options.</param>
     /// <param name="logger">Logging client.</param>
     public IngestManager(
+        IServiceProvider serviceProvider,
```
Fosol, line 40: Missing parameter in docs

services/net/contentmigration/ContentMigrationService.cs
```diff
@@ -34,17 +37,26 @@ public ContentMigrationService(string[] args) : base(args)
...
+
         // TODO: Figure out how to validate without resulting in aggregating the config values.
         // services.AddOptions<ContentMigrationOptions>()
         //     .Bind(this.Configuration.GetSection("Service"))
         //     .ValidateDataAnnotations();
 
+        // Build an intermediate service provider
+        var sp = services.BuildServiceProvider();
+        // This will succeed.
+        var msg = sp.GetService<MigrationSourceContext>();
```
Fosol, line 58: Don't need this anymore

Fosol (changes requested):

db/oracle/docker/scripts/04_data.sql
kylgarmor: - the only names in here appear to be original article authors/pundits which are already public  
- urls refer to original article source on the web  
- both the `TEXT` and `BINARY` fields which are of type `CLOB` and `BLOB` are empty

Fosol (approved):

PR #782, opened 2023-06-01
feat: #tno-1322 - added service config for oracle db connection

kylgarmor:


services/net/contentmigration/Config/ContentMigrationOptions.cs
```diff
@@ -10,14 +10,14 @@ public class ContentMigrationOptions : IngestServiceOptions
...
     /// </summary>
-    public string MigrationVolumePath { get; set; } = "";
+    public string VolumePath { get; set; } = "";
 
     /// <summary>
-    /// get/set - The path to the local mapped volume.
+    /// get/set - Settings to connect to the TNO 1.0 database.
     /// </summary>
-    public string VolumePath { get; set; } = "";
+    public OracleConnectionSettings? OracleConnection { get; set; } = null;
```
Fosol, line 20: Shouldn't need to default the value to null as it will already be that.

Fosol (approved):

PR #788, opened 2023-06-05
feat: #TNO-1322

kylgarmor:
- removed extra property from ContentMigrationOptions
- updated editor app with new tno-core


Fosol (approved):
Nice front-end work  :D

PR #791, opened 2023-06-05
TNO-1383: Today's Commentary View

jdtoombs:
This PR includes the Today's Commentary functionality. 

- Today's commentary now functional by selecting it from the left hand navigation menu
- Defaults to today's date
- Date for commentary is adjustable via the date filter located at the top of the table

### DIFF
--- a/app/subscriber/src/features/todays-commentary/styled/TodaysCommentary.tsx
+++ b/app/subscriber/src/features/todays-commentary/styled/TodaysCommentary.tsx
@@ -1,0 +1,45 @@
+import styled from 'styled-components';
+export const TodaysCommentary = styled.div`
+  .table {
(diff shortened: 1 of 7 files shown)


app/subscriber/src/features/todays-commentary/styled/TodaysCommentary.tsx
```diff
@@ -0,0 +1,45 @@
+import styled from 'styled-components';
+export const TodaysCommentary = styled.div`
+  .table {
+    width: 100%;
+    .group {
+      background-color: #f9f9f9 !important;
```
Fosol, line 6: Makes me think something isn't done right when I see `!important`

Fosol (approved):

PR #797, opened 2023-06-06
feat: #tno-1324 - added ability to store config

kylgarmor:
- service can push back creationDate  of Last NewsItem migrated in run

### DIFF
--- a/libs/net/services/Actions/Managers/IngestActionManager`.cs
+++ b/libs/net/services/Actions/Managers/IngestActionManager`.cs
@@ -105,6 +105,20 @@ public class IngestActionManager<TOptions> : ServiceActionManager<TOptions>, IIn
     }
 
     /// <summary>
+    /// Update Ingest config at runtime.
+    /// </summary>
+    /// <param name="propName"></param>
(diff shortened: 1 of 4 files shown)


libs/net/services/Actions/Managers/IngestActionManager`.cs
```diff
@@ -104,6 +104,20 @@ private async Task<IngestModel> UpdateIngestAsync(int failedAttempts = 0)
...
+    /// </summary>
+    /// <param name="propName"></param>
+    /// <param name="propValue"></param>
+    /// <returns></returns>
+    public override async Task UpdateIngestConfigAsync(string propName, object propValue)
+    {
+        this.Ingest.Configuration[propName] = propValue;
+        var headers = new HttpRequestMessage().Headers;
+        headers.Add("User-Agent", GetType().FullName);
+        this.Ingest = await this.Api.UpdateIngestAsync(Ingest, headers) ?? Ingest;
```
Fosol, line 118: Don't need to handle presently, but a concurrency error will result in a failed import.

Fosol (approved):

PR #798, opened 2023-06-06
Exclude SourceId Filter

jdtoombs:
New filter to exclude specified SourceIds, and a few bug fixes with the home filters
tno-core updated to `0.0.256`


Fosol:
@jdtoombs Can you fix this PR it's blocking my ability to publish another package.

PR #804, opened 2023-06-07
Home filter bug

jdtoombs:
Sorry for all the opening and closing, getting my repos set up properly. Origin was set up incorrectly when I moved my workstations a while back.

New filter to exclude specified SourceIds, and a few bug fixes with the home filters
tno-core updated to `0.0.256`


libs/net/dal/Models/ContentFilter.cs
```diff
@@ -145,6 +145,11 @@ public class ContentFilter : PageFilter
     /// </summary>
     public string? Keyword { get; set; }
 
+    /// <summary>
+    /// get/set - Get/set the source ids to exlude from the search.
+    /// </summary>
+    public long[] ExcludeSourceIds { get; set; } = Array.Empty<long>();
```
Fosol, line 151: Int not long

Fosol (changes requested):

Fosol (approved):

PR #817, opened 2023-06-13
TNO-1406 Editor Fixes

Fosol:
Tags select opens from the top instead of the bottom.  Also fixed layout of time log.

image

Time Log Modal

image


app/editor/src/features/content/form/styled/TimeLogTable.tsx
```diff
@@ -2,11 +2,11 @@ import styled from 'styled-components';
 
 export const TimeLogTable = styled.div`
   div[role='table'] {
-    width: 550px;
```
Fosol, line 5: Don't know why this was hardcoded.

PR #818, opened 2023-06-13
feat: #1325 - PaperMigrator first pass complete

kylgarmor:
further mapping will take place after refactoring SourceContent
fixed filename typo
tweaked clip and image migration
could use refactoring to reduce duplication


TNO.sln
```diff
@@ -1,92 +1,91 @@
...
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30114.105
+# Visual Studio Version 17
+VisualStudioVersion = 17.5.33516.290
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "api", "api", "{F627B24A-217D-4BF1-BC77-E1A92DBCD07F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TNO.API", "api\net\TNO.API.csproj", "{16EA028B-B4C8-416D-BE54-D73D75483668}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TNO.API", "api\net\TNO.API.csproj", "{16EA028B-B4C8-416D-BE54-D73D75483668}"
```
Fosol, line 8: Someone's setup is corrupting these values.  I normally see this when Visual Studio is being used and one developer has one set of IDs and another developer has a set of different IDs.  Can someone sort this out?  @CelloLiu @kylgarmor

services/net/contentmigration/Config/ClipMigrationOptions.cs
```diff
@@ -0,0 +1,37 @@
+using TNO.Services.Config;
+
+namespace TNO.Services.ContentMigration.Config;
+
+/// <summary>
+/// ContentMigrationOptions class, configuration options for Import
+/// </summary>
+public class ClipMigrationOptions : IngestServiceOptions
```
Fosol, line 8: Not sure I see value in having multiple identical options.  Just make a single configuration option for the service.

services/net/contentmigration/Config/ClipMigrationOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Stores a mapping to be used where the TNO 1.0 database 'Type' field value
+    /// doesn't match the Code or Name of any MMIA Source
+    /// The Key should be the "TNO 1.0 Type"
+    /// The Value should be the "MMIA Product Name"
+    /// </summary>
+    public Dictionary<string,string> ProductMappings {get; set; } = new Dictionary<string,string>()
+        {
+            { "TV", "TV / Video News" },
```
Fosol, line 33: Default values like this should be extracted to configuration file.

services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -61,207 +111,52 @@ public ContentMigrationAction(MigrationSourceContext sourceContext, IApiService
...
-        // var lookups = await this.Api.GetLookupsAsync();
+        var lookups = await this.Api.GetLookupsAsync();
+        IEnumerable<IngestModel> ingests = await this.Api.GetIngestsAsync();
 
         var skip = 0;
-        var count = 20;
-        DateTime? importDateStart = manager.Ingest.GetConfigurationValue<DateTime>("importDateStart");
-        DateTime? importDateEnd = manager.Ingest.GetConfigurationValue<DateTime>("importDateEnd");
-        DateTime? creationDateOfNewsItem = manager.Ingest.GetConfigurationValue<DateTime>("creationDateOfLastImport");
+        var count = 50;
```
Fosol, line 120: Value like this would be handle to be part of the configuration options so that it can be controlled without editing code.

services/net/contentmigration/migrators/ClipMigrator.cs
```diff
@@ -0,0 +1,265 @@
...
+            }
+            else if (reference.Status == (int)WorkflowStatus.InProgress && reference.UpdatedOn?.AddMinutes(5) < DateTime.UtcNow)
+            {
+                // If another process has it in progress only attempt to do an Migration if it's
+                // more than an 5 minutes old. Assumption is that it is stuck.
+                reference = await UpdateContentReferenceAsync(reference, WorkflowStatus.InProgress);
+                Logger.LogInformation("Updating migrated content {RSN}:{Title}", newsItem.RSN, newsItem.Title);
+            }
+
+            reference = await FindContentReferenceAsync(reference?.Source, reference?.Uid);
```
Fosol, line 176: Why make a second request for content reference here?

services/net/contentmigration/migrators/ImageMigrator.cs
```diff
@@ -0,0 +1,255 @@
...
+/// <summary>
+/// Implementation of IContentMigrator for Image
+/// </summary>
+public class ImageMigrator : IngestAction<ImageMigrationOptions>, IContentMigrator
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public IEnumerable<string> SupportedIngests => new []{
+        "TNO 1.0 - Image Content"
```
Fosol, line 28: Probably should be based on the configuration options instead of hard coded.

services/net/contentmigration/migrators/ImageMigrator.cs
```diff
@@ -0,0 +1,255 @@
...
+            }
+            else if (reference.Status == (int)WorkflowStatus.InProgress && reference.UpdatedOn?.AddMinutes(5) < DateTime.UtcNow)
+            {
+                // If another process has it in progress only attempt to do an Migration if it's
+                // more than an 5 minutes old. Assumption is that it is stuck.
+                reference = await UpdateContentReferenceAsync(reference, WorkflowStatus.InProgress);
+                Logger.LogInformation("Updating migrated content {RSN}:{Title}", newsItem.RSN, newsItem.Title);
+            }
+
+            reference = await FindContentReferenceAsync(reference?.Source, reference?.Uid);
```
Fosol, line 172: Not sure we need to make a second request for content reference here.  I wonder if this was introduced by Cello when debugging concurrency issues.

services/net/contentmigration/migrators/PaperMigrator.cs
```diff
@@ -0,0 +1,270 @@
...
+/// <summary>
+/// manages Ingestion of TNO 1.0 'Paper'
+/// </summary>
+public class PaperMigrator : IngestAction<PaperMigrationOptions>, IContentMigrator
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public IEnumerable<string> SupportedIngests => new []{
+        "TNO 1.0 - Story Content",
```
Fosol, line 28: I'd be better to pull this from the configuration options rather than hardcoded.

services/net/contentmigration/migrators/PaperMigrator.cs
```diff
@@ -0,0 +1,270 @@
...
+            }
+            else if (reference.Status == (int)WorkflowStatus.InProgress && reference.UpdatedOn?.AddMinutes(5) < DateTime.UtcNow)
+            {
+                // If another process has it in progress only attempt to do an Migration if it's
+                // more than an 5 minutes old. Assumption is that it is stuck.
+                reference = await UpdateContentReferenceAsync(reference, WorkflowStatus.InProgress);
+                Logger.LogInformation("Updating migrated content {RSN}:{Title}", newsItem.RSN, newsItem.Title);
+            }
+
+            reference = await FindContentReferenceAsync(reference?.Source, reference?.Uid);
```
Fosol, line 181: Same issue with extra request for content reference.

tools/scripts/gen-env-files.sh
```diff
@@ -448,7 +448,7 @@ echo \
 # Keep this as a .env so it doesn't get committed to source.
 kafka:
   brokers:
-    - broker:9092
+    - broker:29092
```
Fosol, line 451: Thanks!

Fosol (changes requested):

services/net/contentmigration/Config/ClipMigrationOptions.cs
kylgarmor: I would love to do this but I couldnt figure out the correct way to get DependencyInjection to work with the same "Options" being used by different migrators.  If you take a look at the `ContentMigratorFactoryExtensions.cs` you will see what I mean.

services/net/contentmigration/migrators/PaperMigrator.cs
kylgarmor: removed in all three places.  Probably a copy-paste on my side.

services/net/contentmigration/migrators/ClipMigrator.cs
```diff
@@ -0,0 +1,264 @@
...
+/// <summary>
+/// manages Ingestion of TNO 1.0 'Snippet'
+/// </summary>
+public class ClipMigrator : IngestAction<ClipMigrationOptions>, IContentMigrator
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public IEnumerable<string> SupportedIngests => new []{
+        "TNO 1.0 - Snippet Content"
```
Fosol, line 28: Probably should be based on the configuration options instead of hard coded.

services/net/contentmigration/Config/ClipMigrationOptions.cs
Fosol: You can do this a few different ways.  However, the main way is through configuration through the appsettings.jons file, or ENV variables.  Are you running each type of the Content Migration Service in a single instance?  Or will we run four different instances?  If you are doing it all in one instances then you can separate the configuration and dependency injection by passing a different path to the configuration.

```
{
  "Service": {
    "Snippet": {
      "key": "value"
    },
    "Image": {
      "key": "value"
    }
  }
}
```

Fosol: If you're doing this with four different instances then you don't need to have a different configuration path for each time.

Fosol (changes requested):

TNO.sln
kylgarmor: I've reverted the changes here.  I will keep an eye on it going forward.

services/net/contentmigration/migrators/ContentMigrator.cs
```diff
@@ -0,0 +1,235 @@
...
+using TNO.Kafka.Models;
+using TNO.Services.ContentMigration.Config;
+using TNO.Services.ContentMigration.Sources.Oracle;
+
+namespace TNO.Services.ContentMigration.Migrators;
+
+/// <summary>
+/// Interface for ContentMigrator implementations
+/// </summary>
+public interface IContentMigrator
```
Fosol, line 19: Extract interfaces into their own files.

services/net/contentmigration/migrators/ContentMigratorFactoryExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    ///
+    /// </summary>
+    /// <param name="services"></param>
+    /// <param name="configuration"></param>
+    /// <returns></returns>
+    public static IServiceCollection AddSourceMigrators(this IServiceCollection services, IConfiguration configuration)
+    {
+        services.Configure<MigratorOptions>("ClipMigrator", configuration.GetSection("Service:ClipMigrator"));
```
Fosol, line 20: I don't see anything obvious referencing the names you provide.  Is there a convention that understands how to inject the correct object?

Fosol (changes requested):

services/net/contentmigration/migrators/ContentMigratorFactoryExtensions.cs
kylgarmor: The name I'm using as a convention is the same as the name of the class which will need to consume those named options.  If you take a look at `ContentMigrator.cs` line 68 you can see it there.  I have added a comment for it now.
```
        // as a convention the Named Option should be the same as the class name which needs to consume it
        this.MigratorOptions = migratorOptions.Get(this.GetType().Name);
```

kylgarmor: arguably I could wire it a bit tighter with something like this:
```
services.Configure<MigratorOptions>(typeof(ClipMigrator).Name, configuration.GetSection("Service:ClipMigrator"));
```

Fosol (approved):

PR #821, opened 2023-06-14
TNO-1386: (1/2) Create search result page

jdtoombs:
TNO-1386 is to add a View Media button to watch videos "in line" on the search results page. This search results page has not been made yet so this PR is capturing that work. 

Will release part (2/2) in a subsequent PR. Splitting of functionality to keep things easier.

In this PR:

- users can now search with a keyword
- after searching the user will be redirected to a new page
- the user may see a quick summary and overview of the content and click the headline to navigate to the view screen

searchpage

### DIFF
--- a/app/subscriber/src/components/search-with-logout/SearchWithLogout.tsx
+++ b/app/subscriber/src/components/search-with-logout/SearchWithLogout.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
 import { BiLogOut } from 'react-icons/bi';
 import { FaSearch, FaUserCircle } from 'react-icons/fa';
+import { useNavigate, useParams } from 'react-router-dom';
 import { Button, Row, Show, Text, useKeycloakWrapper, useWindowSize } from 'tno-core';
 
 import * as styled from './styled';
-import { useNavigate, useParams } from 'react-router-dom';
 
 /** The component containing the search bar and the logout button for the subscriber application. Responsive and adjusts to the viewing device*/
 export const SearchWithLogout: React.FC = () => {
--- a/app/subscriber/src/components/search-with-logout/SearchWithLogout.tsx
+++ b/app/subscriber/src/components/search-with-logout/SearchWithLogout.tsx
@@ -1,32 +1,43 @@
 import React from 'react';
 import { BiLogOut } from 'react-icons/bi';
-import { FaUserCircle } from 'react-icons/fa';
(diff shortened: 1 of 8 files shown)


app/subscriber/src/components/search-with-logout/SearchWithLogout.tsx
```diff
@@ -1,32 +1,43 @@
...
 import * as styled from './styled';
 
 /** The component containing the search bar and the logout button for the subscriber application. Responsive and adjusts to the viewing device*/
 export const SearchWithLogout: React.FC = () => {
   const keycloak = useKeycloakWrapper();
-  const [, { storeFilterAdvanced }] = useContent();
+  const [searchItem, setSearchItem] = React.useState<string>('');
+  const navigate = useNavigate();
   const { width } = useWindowSize();
+  const { id } = useParams();
```
Fosol, line 15: Where is the `id` parameter coming from?  The path you have below is `/search/${searchItem}`.

app/subscriber/src/components/search-with-logout/SearchWithLogout.tsx
```diff
@@ -1,32 +1,43 @@
...
+
+  // update search item when id changes on search page
+  React.useEffect(() => {
+    if (window.location.pathname.includes('search')) {
+      if (!!id) setSearchItem(id ?? '');
```
Fosol, line 20: Why is the parameter 'id' when you're expecting some random text?

Fosol: Also why even bother using the state value if you already have the value `id` from the hook `useParams()`?  You're just keeping two copies of the same thing.

app/subscriber/src/features/router/AppRouter.tsx
```diff
@@ -35,6 +36,7 @@ export const AppRouter: React.FC<IAppRouter> = () => {
         <Route path="welcome" element={<AccessRequest />} />
         <Route path="access/request" element={<AccessRequest />} />
         <Route path="/landing/:id" element={<Landing />} />
+        <Route path="/search/:id" element={<SearchPage />} />
```
Fosol, line 39: `:id` parameter is a very misleading parameter.  Which caused me confusion when reviewing your above code.  This also should be a query string variable as mentioned above.

Fosol (approved):

Fosol (approved):

PR #823, opened 2023-06-14
TNO-1386: (2/2 )Add media player to view

jdtoombs:
This PR includes the ability to click "PLAY MEDIA" in the search view when a piece of content has the appropriate file references. Doing so will open a media player on the right hand side of the screen, while also indicating that it is playing on the button that previously said PLAY MEDIA.

Bugs fixed 🐛 : 

- can now scroll on search results when contents expand pass screen view

sidemediaplayer

### DIFF
--- a/app/subscriber/src/features/search-page/player/Player.tsx
+++ b/app/subscriber/src/features/search-page/player/Player.tsx
@@ -16,8 +16,6 @@ export const Player: React.FC<IPlayerProps> = ({ content }) => {
   const { width } = useWindowSize();
   const path = content?.fileReferences ? content.fileReferences[0]?.path : '';
 
-  console.log(content);
-
   React.useEffect(() => {
     if (!!path)
       stream(path).then((result) => {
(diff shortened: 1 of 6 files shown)


app/subscriber/src/features/search-page/player/Player.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/** Component used to play media on the search results screen */
+export const Player: React.FC<IPlayerProps> = ({ content }) => {
+  const [avStream, setAVStream] = React.useState<IStream>();
+  const [, { stream }] = useContent();
+  const { width } = useWindowSize();
+  const path = content?.fileReferences ? content.fileReferences[0]?.path : '';
+
+  console.log(content);
```
Fosol, line 19: Remove

Fosol (changes requested):

app/subscriber/src/features/search-page/player/Player.tsx
jdtoombs: done

Fosol (approved):

PR #825, opened 2023-06-14
Frontpage bug

jdtoombs:
For some reason, this is 10 on my local; however, showing a `productId `of 11 on DEV.

image


libs/net/dal/Services/ContentService.cs
```diff
@@ -182,7 +182,7 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindFr
     {
         var productQueries = new List<Func<QueryContainerDescriptor<API.Areas.Services.Models.Content.ContentModel>, QueryContainer>>();
         var today = DateTime.Today.ToUniversalTime();
-        productQueries.Add(q => q.Raw(@"{""match"": {""productId"": 10}}"));
```
Fosol, line 185: Hardcoding things always leads to bugs long term

Fosol (approved):

PR #832, opened 2023-06-16
TNO-22 Create Scheduler Service

Fosol:
The new Scheduler Service provides a way to request reports and notifications based on a configured schedule.  The Scheduler Service is a Kafka Producer and can be extended to send any type of request if we need it to.  We could even schedule messages to show up in the UI.

> Update your local environments with `make renew`

## Summary

- New Scheduler Service
- New DB migration `1.0.29`
- Published `tno-core:0.0.263`


app/editor/src/features/admin/ingests/IngestForm.tsx
```diff
@@ -62,6 +62,7 @@ export const IngestForm: React.FC<IIngestProps> = (props) => {
   const reload = async () => {
     try {
       const data = await api.getIngest(ingestId);
+      setIngest({ ...data, schedules: [] }); // Required because react state and formik are horrible.
```
Fosol, line 65: This is stupid...

libs/net/models/Extensions/ConnectionModelExtensions.cs
```diff
@@ -82,12 +83,58 @@ public static string GetConfigurationValue(this API.Areas.Services.Models.DataLo
...
     }
+
+    /// <summary>
+    /// Convert number to specified type.
+    /// This method handles enum values.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="element"></param>
+    /// <returns></returns>
+    private static T? ConvertNumberTo<T>(JsonElement element)
```
Fosol, line 102: JSON doesn't understand different number types.  This method 'should' handle different number types better.

PR #833, opened 2023-06-16
feat: tno-1328, tno-1329 - Tone and Topic mappping

kylgarmor:
- Topic is mapped to an existing Topic if possible
- or a new Topic will be created
- Tone is mapped, but only the "Default" tone


libs/net/kafka/Models/TonePool.cs
```diff
@@ -0,0 +1,20 @@
+namespace TNO.Kafka.Models;
+
+public class TonePool
+{
+    #region Properties
+    public long Value { get; set; }
```
Fosol, line 6: Change to int

libs/net/kafka/Models/Topic.cs
```diff
@@ -0,0 +1,20 @@
...
+public class Topic
+{
+    #region Properties
+    public string TopicType { get; set; }
+    public string Name { get; set; }
+
+    #endregion
+
+    #region Constructors
+    public Topic() { }
```
Fosol, line 12: Resolve linting issues.

libs/net/kafka/Models/Topic.cs
```diff
@@ -0,0 +1,20 @@
+namespace TNO.Kafka.Models;
+
+public class Topic
+{
+    #region Properties
+    public string TopicType { get; set; }
```
Fosol, line 6: Change to enum type.

services/net/content/ContentManager.cs
```diff
@@ -446,4 +482,47 @@ private async Task<ContentModel> CopyFileWithSSHAsync(DataLocationModel dataLoca
...
+            Score = 0, // TODO: until we can dig score out of TNO 1.0
+            ContentId = 0 // for new content
+        };
+        var topic = topics.Where(s => s.Name == eodCategory && s.TopicType.ToString().Equals(eodCategoryGroup,StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
+        if (topic != null) {
+            topicModel.Id = topic.Id;
+            topicModel.Name = topic.Name;
+            topicModel.TopicType = topic.TopicType;
+        } else {
+            // KGM: will this create a new Topic, if one doesnt already exist?
```
Fosol, line 519: Yes the Content Service will need to create a topic if it doesn't exist.  However, it will have to disabled because we only want current topics enabled.

services/net/contentmigration/migrators/ClipMigrator.cs
```diff
@@ -73,6 +73,17 @@ public ClipMigrator(IApiService api, IOptionsSnapshot<MigratorOptions> migratorO
             content.UpdatedOn = newsItem.UpdatedOn != DateTime.MinValue ? newsItem.UpdatedOn.Value.ToUniversalTime() : null;
         }
 
+        if (newsItem.Tones != null) {
+            // TODO: replace the USER_RSN value on UserIdentifier with something that can be mapped by the Content Service to an MMIA user
+            // TODO: remove UserRSN filter once user can be mapped
+            content.TonePools = newsItem.Tones.Where(t => t.UserRSN == 0)
+                .Select(t => new Kafka.Models.TonePool { Value = t.ToneValue, UserIdentifier = t.UserRSN.ToString() });
```
Fosol, line 80: It'll have to be email or username, or some combination regrettably.

Fosol (changes requested):

services/net/contentmigration/migrators/ClipMigrator.cs
kylgarmor: I added a story to take care of this lookup/mapping on both sides
https://apps.itsm.gov.bc.ca/jira/browse/TNO-1436

Fosol (approved):

PR #834, opened 2023-06-19
Tno-1328

kylgarmor:
- add new Topics at the ContentController


services/net/content/ContentManager.cs
```diff
@@ -446,4 +482,47 @@ private async Task<ContentModel> CopyFileWithSSHAsync(DataLocationModel dataLoca
...
     }
     #endregion
+
+    #region Helper Methods
+    private static ContentTonePoolModel? GetTonePoolMapping(IEnumerable<API.Areas.Editor.Models.TonePool.TonePoolModel> tonePools, int toneValue, string? userIdentifier)
+    {
+        // use the "Default" TonePool
+        if (string.IsNullOrEmpty(userIdentifier)) {
+            // the "Default" tone should always be present
+            var tonePoolDefault = tonePools.Where(s => s.Name.Equals("Default")).First();
```
Fosol, line 492: Ideally this would be based on configuration rather than hardcoded.

services/net/contentmigration/sources/oracle/Entities/NewsItem.cs
```diff
@@ -329,4 +329,10 @@ public class NewsItem
     /// </summary>
     [Column("EOD_DATE")]
     public string? EodDateTime { get; set; }
+
+    /// <summary>
+    /// get/set
+    /// </summary>
+    public ICollection<UserTone>? Tones { get;} = new List<UserTone>();
```
Fosol, line 336: Normally relationships are not nullable, especially if you initialize them.

Fosol (approved):

services/net/content/ContentManager.cs
kylgarmor: fixed by adding a `ContentMigrationOptions` section for the `ContentManager` service.

services/net/contentmigration/sources/oracle/Entities/NewsItem.cs
kylgarmor: fixed

PR #836, opened 2023-06-19
TNO-1394 Fix Indexing Service

Fosol:
The Indexing Service was pushing content to the published index with the `status=publish` instead of `status=published`.

### DIFF
--- a/api/net/Areas/Subscriber/Controllers/ContentController.cs
+++ b/api/net/Areas/Subscriber/Controllers/ContentController.cs
@@ -11,14 +11,8 @@ using TNO.DAL.Services;
 using TNO.DAL.Config;
 using TNO.Entities.Models;
 using TNO.Core.Extensions;
-using TNO.Kafka;
-using TNO.API.Config;
 using System.Net.Mime;
-using TNO.API.Helpers;
 using System.Web;
-using System.Text.Json;
-using TNO.API.SignalR;
-using Microsoft.AspNetCore.SignalR;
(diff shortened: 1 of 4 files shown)


api/net/Areas/Subscriber/Controllers/ContentController.cs
```diff
@@ -11,14 +11,8 @@
 using TNO.DAL.Config;
 using TNO.Entities.Models;
 using TNO.Core.Extensions;
-using TNO.Kafka;
```
Fosol, line 14: @jdtoombs Cleaning up unused code in controller.

docker-compose.yml
```diff
@@ -19,6 +19,14 @@ services:
     container_name: tno-database
     build:
       context: ${DB_CONTEXT:-db/postgres/docker}
+    deploy:
```
Fosol, line 22: Adding resource limits to pods to reduce overhead and debug why API keeps crashing after it grows in memory usage.

services/net/indexing/IndexingManager.cs
```diff
@@ -289,6 +289,7 @@ private async Task PublishContentAsync(IndexRequestModel request, ContentModel c
         // Remove the transcript body if it hasn't been approved.
         var body = content.Body;
         if (!content.IsApproved && content.ContentType == ContentType.Snippet) content.Body = "";
+        content.Status = ContentStatus.Published;
```
Fosol, line 292: Here was the issue.

PR #837, opened 2023-06-19
TNO-1388: Add ability for subscriber to transcribe

jdtoombs:
This PR includes the functionality that allows subscribers to transcribe a video if it has not been transcribed already. Clearly my own instance of Azure is having some issues as seen by the text it has generated..

However, you can see that after the transcription is appended to the body the application will add a new section for the transcription as was discussed with Bobbi on Friday.

image

### DIFF
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -203,7 +203,7 @@ export const ViewContent: React.FC = () => {
               ])
             }
           >
-            Transcribe
+            <div className="text">Transcribe</div>
             <WorkOrderStatus workOrders={workOrders} type={WorkOrderTypeName.Transcription} />
           </Button>
         </Show>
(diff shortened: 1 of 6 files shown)


app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -153,7 +188,33 @@ export const ViewContent: React.FC = () => {
         >
           <p>{parse(content?.summary ?? '')}</p>
         </Show>
+        <Show visible={content?.contentType === ContentTypeName.Snippet}>
+          <Button
+            onClick={() => handleTranscribe()}
+            variant={ButtonVariant.action}
+            className="transcribe-button"
+            disabled={
+              (!!content?.fileReferences && !content?.fileReferences.length) ||
```
jdtoombs, line 197: Button disabled when there is no video attached to a snippet, or if the transcription has already been completed,

app/subscriber/src/features/content/view-content/utils/WorkOrderStatus.tsx
```diff
@@ -0,0 +1,63 @@
+import { FaCheckCircle, FaExclamationCircle, FaTimesCircle } from 'react-icons/fa';
```
jdtoombs, line 1: Should I just add this to tno-core?

jdtoombs: And other related items

Fosol: For now I don't think so.

Fosol (approved):

PR #840, opened 2023-06-20
feat: tno-1330 - Import Tags found in Summary

kylgarmor:
- existing tags will be mapped
- new tags will be added with a dummy name to prompt editors to fix them

### DIFF
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -37,7 +37,7 @@ export const ViewContent: React.FC = () => {
   const [, { getContent, stream }] = useContent();
   const [avStream, setAVStream] = React.useState<IStream>();
   const { width } = useWindowSize();
-  const [, { transcribe }] = useWorkOrders();
+  const [, { transcribe, findWorkOrders }] = useWorkOrders();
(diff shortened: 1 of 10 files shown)


services/net/content/Config/ContentOptions.cs
```diff
@@ -39,7 +39,7 @@ public class ContentOptions : ServiceOptions
     /// <summary>
     /// get/set - Settings around content coming from the Content Migration service
     /// </summary>
-    public ContentMigrationOptions MigrationOptions { get; set; } = new ContentMigrationOptions();
+    public ContentMigrationOptions? MigrationOptions { get; set; } = new ContentMigrationOptions();
```
Fosol, line 42: I removed the `?` because it's not needed, nor is it good practice when we can ensure the object is not null ever.

Fosol (changes requested):

app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -212,7 +215,7 @@ export const ViewContent: React.FC = () => {
         <hr />
         <h3>Transcription:</h3>
         <Row>
-          <p>{content?.body}</p>
+          <p>{parse(content?.body ?? '')}</p>
```
Fosol, line 218: Not certain why you're editing the Subscriber site?

PR #844, opened 2023-06-20
TNO-1411: Add page number to content view

jdtoombs:
Simple fix to add page number on the content view page.

image

### DIFF
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -152,7 +152,8 @@ export const ViewContent: React.FC = () => {
         </p>
         <p className="source-section">
           <Show visible={content?.contentType === ContentTypeName.PrintContent}>
-            {content?.source?.name} - {`${content?.section}: ${content?.page}`}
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -149,7 +149,7 @@ export const ViewContent: React.FC = () => {
         </p>
         <p className="source-section">
           <Show visible={content?.contentType === ContentTypeName.PrintContent}>
-            {content?.source?.name} - {content?.section}
+            {content?.source?.name} - {`${content?.section}: ${content?.page}`}
```
Fosol, line 152: Not all will have a section.  Which would result in an extra `:` in the output.

Fosol (changes requested):

app/subscriber/src/features/content/view-content/ViewContent.tsx
jdtoombs: Will adjust!

Fosol (approved):

PR #846, opened 2023-06-21
feat: #tno-1327 - capture migrated content Actions

kylgarmor:
- only map tags which are present in MMIA
- use configuration+enum to map Tag names

### DIFF
--- a/services/net/contentmigration/migrators/ContentMigrator.cs
+++ b/services/net/contentmigration/migrators/ContentMigrator.cs
@@ -1,4 +1,5 @@
 using System.Linq.Expressions;
+using System.Text.RegularExpressions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using TNO.API.Areas.Editor.Models.Lookup;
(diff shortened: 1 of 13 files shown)


services/net/contentmigration/migrators/ContentMigrator.cs
```diff
@@ -177,5 +179,54 @@ public ContentReferenceModel CreateContentReference(SourceModel source, string t
...
+    /// <param name="alert"></param>
+    /// <param name="commentary"></param>
+    /// <param name="commentaryTimeout"></param>
+    /// <returns></returns>
+    public IEnumerable<Kafka.Models.Action> GetActionMappings(bool frontPageStory, bool wapTopStory, bool alert, bool commentary, double? commentaryTimeout) {
+        List<Kafka.Models.Action> mappedActions = new();
+        string actionName;
+        ActionType actionType;
+
+        if (frontPageStory) {
```
Fosol, line 196: Looks like possible duplication that could be reduced.  Not important for this work however.

Fosol (approved):

PR #851, opened 2023-06-22
TNO-1394 Enhance Editor Papers

Fosol:
Made it easier to remove content from the selected/published list on the Papers page.

## Summary

- If a filter is applied (top stories, commentary, or homepage) it will display buttons to remove content from those
- A 'Publish All' button is displayed when one of the filters is applied so that all items on the page will be published

image

### DIFF
--- a/app/editor/src/features/content/papers/components/ReportActions.tsx
+++ b/app/editor/src/features/content/papers/components/ReportActions.tsx
@@ -4,6 +4,7 @@ import { getDefaultCommentaryExpiryValue } from 'features/content/form/utils';
 import * as React from 'react';
 import { toast } from 'react-toastify';
 import { useContent, useLookup } from 'store/hooks';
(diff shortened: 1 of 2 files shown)


app/editor/src/features/content/papers/components/ReportActions.tsx
```diff
@@ -92,34 +98,58 @@ export const ReportActions: React.FunctionComponent<IReportActionProps> = ({
...
+        onClick={() =>
+          handleAction(
+            ContentListActionName.Action,
+            ActionName.TopStory,
+            filter.topStory ? 'false' : 'true',
+          )
+        }
       >
-        Top Story
+        {filter.topStory ? `Remove from` : `Add to`} Top Story
```
Fosol, line 109: Switch from Add to Remove based on the filter.

PR #852, opened 2023-06-22
fix: #tno-1461 - fixed sql

kylgarmor:
- reduced the number of records being imported to NEWS_ITEMS
- reduced the number of records being imported to USERS_TONES


Fosol (approved):
I'm not a fan of including these files in source, but for now it's fine.

PR #853, opened 2023-06-22
TNO-1378: Multiple ministers can now be selected

jdtoombs:
In this PR:
- another console warning bug fix
- multiple ministers can now be selected
- when none are selected it will now return no content rather than all
- bold all ministers that are selected as they appear when viewing content

Little bit tricky with highlighting multiple ministers, let me know if anything needs clarification. Have it working as seen below:

myministers

### DIFF
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -3,7 +3,7 @@ import parse from 'html-react-parser';
 import React from 'react';
 import { useParams } from 'react-router-dom';
 import { toast } from 'react-toastify';
-import { useContent, useWorkOrders } from 'store/hooks';
+import { useApp, useContent, useWorkOrders } from 'store/hooks';
 import {
   Button,
   ButtonVariant,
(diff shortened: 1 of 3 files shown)


app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -61,20 +64,36 @@ export const ViewContent: React.FC = () => {
...
-      }
+          if (
+            content?.body &&
+            !content.body.includes(`<b>${minister}</b>`) &&
+            content.body.includes(minister)
+          ) {
+            tempBody = tempBody?.replace(regex, `<b>${minister}</b>`);
+          }
+
+          if (index === myMinisters.length - 1) {
```
jdtoombs, line 90: Have to use the temp variables as scope requires it (otherwise will overwrite with last replacement in the list), also only once we are done iterating the user's ministers is when we want to update the content

app/subscriber/src/features/my-minister/MyMinister.tsx
```diff
@@ -25,10 +26,12 @@ export const MyMinister: React.FC = () => {
           makeFilter({
             ...filter,
             startDate: '',
+            contentTypes: [],
             endDate: '',
           }),
         );
-        setHomeItems(data.items);
+        // don't want to keyword fetch when there is nothing to fetch
+        setHomeItems(userInfo?.preferences?.myMinisters.length > 0 ? data.items : []);
```
jdtoombs, line 34: This was causing a bug that was returning all content when the user had no ministers selected

Fosol (approved):

app/subscriber/src/features/my-minister/MyMinister.tsx
jdtoombs: (since it is tied to keyword and the keyword would be '')

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -1,75 +1,69 @@
...
+    }
+  }, [userInfo, ministers]);
+
   return (
     <styled.MyMinisterSettings>
       <p className="description">
         Choose the Minister you'd like to follow. Stories about your selected Minister will be
         available from a quick click in the sidebar menu.
       </p>
-      <FormikForm
```
jdtoombs, line 38: No need for Formik, was seeming to make it more complex and slow.

PR #854, opened 2023-06-22
TNO-1288: Adjust order of display of papers

jdtoombs:
In this PR:

- papers on the home page will now display based off the papers defined sort order 
- these can be updated at any time through the editor -> admin -> sources -> sort order


app/subscriber/src/features/home/Home.tsx
```diff
@@ -32,6 +32,7 @@ export const Home: React.FC = () => {
             contentTypes:
               filter.contentTypes.length > 0 ? filter.contentTypes : [ContentTypeName.PrintContent],
             startDate: filter.startDate ? filter.startDate : new Date().toDateString(),
+            sort: [{ id: 'sourceSort' }],
```
Fosol, line 35: This should be the property name to sort by, which would be `source.sortOrder`

Fosol (changes requested):

Fosol (approved):

PR #861, opened 2023-06-27
feat: #tno-1331 - Ingest files from TNO 1.0

kylgarmor:
- using the public endpoint to pull files over http
- files are copied locally and so are available to the content service
- updated the Ingest configuration to use local staging destination


libs/net/dal/Migrations/1.0.35/Up/PostUp/00-Connection.sql
```diff
@@ -0,0 +1,56 @@
...
+  , 'A locally mapped volume location for storing migrated image files.' -- description
+  , true -- is_enabled
+  , 0 -- connection_type - Local Volume
+  , '{ "path": "images-migrated" }' -- configuration
+  , false -- is_read_only
+  , 0 -- sort_order
+  , ''
+  , ''
+),  (
+  'Local Volume - Migrated Papers'
```
Fosol, line 35: I'd be surprised if there was any files associated with this content.

libs/net/dal/Migrations/1.0.35/Up/PostUp/00-Connection.sql
```diff
@@ -0,0 +1,56 @@
...
+  , 'A locally mapped volume location for storing migrated paper files.' -- description
+  , true -- is_enabled
+  , 0 -- connection_type - Local Volume
+  , '{"path": "papers-migrated"}' -- configuration
+  , false -- is_read_only
+  , 0 -- sort_order
+  , ''
+  , ''
+),  (
+  'Local Volume - Migrated Stories'
```
Fosol, line 45: I'd be surprised if there was any files associated with this content.

Fosol (approved):
Looks great

PR #862, opened 2023-06-28
fix: #tno-1473 - fixed config update

kylgarmor:
- created new endpoint for updating ingest configuration
- refactored existing endpoint to match name
- tweaked date format for storing ingest filter dates


Fosol (approved):
Thanks

PR #864, opened 2023-06-28
feat: #tno-14745 - using a union all to solve

kylgarmor:
- trying a UNION ALL query inside a CTE on the EntityBuilder


services/net/contentmigration/sources/oracle/Entities/NewsItem.cs
```diff
@@ -5,7 +5,7 @@ namespace TNO.Services.ContentMigration.Sources.Oracle;
 /// <summary>
 /// NewsItem class, provides an entity to store News Item records in the database.
 /// </summary>
-[Table("NEWS_ITEMS")]
+// [Table("NEWS_ITEMS")]
```
Fosol, line 8: Did you want to do this?

kylgarmor: yes, because we are no longer tied to the `NEWS_ITEMS` table.  The CTE query in the `NewsItemConfiguration.cs` takes care of it instead.  I have a slight worry about the query and the schema on the tables when this moves to dev/test, but it will be a quick fail if it does and I already have an idea to fall back on.

Fosol (approved):

PR #866, opened 2023-06-28
TNO-1439: Include source on internet types

jdtoombs:
Include source name on content with the content type of Internet, other types already include this. 

image


app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -74,7 +75,6 @@ export const ViewContent: React.FC = () => {
       myMinisters.length &&
         myMinisters.forEach((minister: string, index: number) => {
           const regex = new RegExp(minister ?? '', 'gi');
-          console.log(minister);
```
Fosol, line 77: Thanks for removing

Fosol (approved):

PR #872, opened 2023-06-29
fix: tno-1497 - re-mapped Categories to Labels

kylgarmor:
- Syndicated content has Categories
- They should be assigned to Labels instead of Tags

### DIFF
--- a/libs/net/kafka/Models/ContentBase.cs
+++ b/libs/net/kafka/Models/ContentBase.cs
@@ -120,6 +120,11 @@ public abstract class ContentBase
     public IEnumerable<Tag> Tags { get; set; } = Array.Empty<Tag>();
 
     /// <summary>
+    /// get/set - An array of labels to identify this content.
+    /// </summary>
+    public IEnumerable<ContentLabel> Labels { get; set; } = Array.Empty<ContentLabel>();
+
+    /// <summary>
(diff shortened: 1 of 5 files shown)


libs/net/kafka/Models/ContentBase.cs
```diff
@@ -119,6 +119,11 @@ public abstract class ContentBase
     /// </summary>
     public IEnumerable<Tag> Tags { get; set; } = Array.Empty<Tag>();
 
+    /// <summary>
+    /// get/set - An array of labels to identify this content.
+    /// </summary>
+    public IEnumerable<ContentLabel> Labels { get; set; } = Array.Empty<ContentLabel>();
```
Fosol, line 125: I think you're referencing an entity here.  It should be a model from the Kafka library

Fosol (changes requested):

libs/net/kafka/Models/ContentBase.cs
kylgarmor: no, it's referencing the Kafka model which is new.  I notice you had a TODO around changing the Kafka Model names be suffixed with Model wo when I created this one I took that step in advance
image

Fosol (approved):

PR #874, opened 2023-06-30
feat: #tno-1404 - time and effort now migrated

kylgarmor:
- created new models
- mapped value from TNO db "NUMBER2"
- fixed creation date time mapping

### DIFF
--- a/services/net/contentmigration/ContentMigrationAction.cs
+++ b/services/net/contentmigration/ContentMigrationAction.cs
@@ -11,6 +11,7 @@ using TNO.Services.Actions;
 using TNO.Services.ContentMigration.Config;
 using TNO.Services.ContentMigration.Sources.Oracle;
 using TNO.Services.ContentMigration.Migrators;
+using TNO.Services.ContentMigration.Extensions;
 
 namespace TNO.Services.ContentMigration;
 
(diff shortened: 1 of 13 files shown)


services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -130,19 +131,20 @@ public ContentMigrationAction(MigrationSourceContext sourceContext, ContentMigra
             {
                 var baseFilter = contentMigrator.GetBaseFilter();
                 IQueryable<NewsItem> items = GetFilteredNewsItems(_sourceContext.NewsItems, baseFilter, skip, count, manager.Ingest.LastRanOn, importDateStart, importDateEnd, creationDateOfLastImport);
-                var lastNewsItem = items.Last();
                 count = items.Count();
+                if (count == 0) break;
+
+                var lastNewsItem = items.Last();
```
Fosol, line 137: What is this used for?

services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -130,19 +131,20 @@ public ContentMigrationAction(MigrationSourceContext sourceContext, ContentMigra
             {
                 var baseFilter = contentMigrator.GetBaseFilter();
                 IQueryable<NewsItem> items = GetFilteredNewsItems(_sourceContext.NewsItems, baseFilter, skip, count, manager.Ingest.LastRanOn, importDateStart, importDateEnd, creationDateOfLastImport);
-                var lastNewsItem = items.Last();
                 count = items.Count();
+                if (count == 0) break;
```
Fosol, line 135: Why bother with this?  The foreach on 139 will exit immediately if there are no items.

services/net/contentmigration/Extensions/NewsItemExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+public static class DictionaryExtensions
+{
+    /// <summary>
+    /// Get the value from the dictionary for the specified 'key' and return it as an int.
+    /// </summary>
+    /// <param name="dict"></param>
+    /// <param name="key"></param>
+    /// <param name="defaultValue"></param>
+    /// <returns></returns>
+    public static int GetIntValue(this IDictionary<string, Microsoft.Extensions.Primitives.StringValues> dict, string key, int defaultValue = 0)
```
Fosol, line 17: I'm certain we already have this extension in the Core library.

services/net/contentmigration/migrators/ClipMigrator.cs
```diff
@@ -64,18 +65,18 @@ public ClipMigrator(IApiService api, IOptionsSnapshot<MigratorOptions> migratorO
             // FilePath = (ingest.DestinationConnection?.GetConfigurationValue("path")?.MakeRelativePath() ?? "")
             //     .CombineWith($"{ingest.Source?.Code}/{GetDateTimeForTimeZone(ingest):yyyy-MM-dd}/", reference.Uid),
             Language = "", // TODO: Need to extract this from the ingest, or determine it after transcription.
-            //Tags = item.Categories.Select(c => new TNO.Kafka.Models.Tag(c.Name, c.Label))
-
         };
 
+        var auditUser = lookups.Users.FirstOrDefault(u => u.Username == this.Options.DefaultUserNameForAudit);
```
Fosol, line 70: This could return null.  Throw a configuration exception to inform us something isn't configured correctly.

services/net/contentmigration/migrators/ClipMigrator.cs
```diff
@@ -64,18 +65,18 @@ public ClipMigrator(IApiService api, IOptionsSnapshot<MigratorOptions> migratorO
...
+
         // newsItem.string5 & newsItem.string5 seem to be the "Show/Program"
 
         if (newsItem.UpdatedOn != null)
         {
             content.UpdatedOn = newsItem.UpdatedOn != DateTime.MinValue ? newsItem.UpdatedOn.Value.ToUniversalTime() : null;
         }
 
-        if (newsItem.Tones != null)
+        if ((newsItem.Tones != null) && (newsItem.Tones.Any()))
```
Fosol, line 79: You can also do this `newsItem.Tones?.Any() == true`

Fosol (changes requested):

services/net/contentmigration/Extensions/NewsItemExtensions.cs
kylgarmor: bad copy paste...

services/net/contentmigration/ContentMigrationAction.cs
kylgarmor: There are two loops happening here: 
First loop is picking up a manageable chunk of records each time. Second loop iterates over each item in the "chunk". `skip` is used to manage how many records we skip in the fetch from the database as we have (in theory) already migrated those. The `skip` value is used as the value for the `OFFSET` when it gets down to the `SQL` call.

services/net/contentmigration/migrators/ClipMigrator.cs
kylgarmor: definitely will do.

services/net/contentmigration/migrators/ClipMigrator.cs
kylgarmor: I like it. Much more succinct.

Fosol (approved):

Fosol (changes requested):
The history is wrong

Fosol (approved):

PR #883, opened 2023-07-05
TNO-1485 Add report template admin

Fosol:
Providing admin pages to manage report templates.

## Summary

- Published tno-core:0.0.267
- Added way to view all report templates
- Added way to edit and delete report templates
- Added Cello and Kyle to code owners so they can receive PR notifications
- Fixed issue with Col and Row component not including inline styling

image


app/editor/src/features/admin/actions/styled/ActionList.tsx
```diff
@@ -2,6 +2,7 @@ import styled from 'styled-components';
 
 export const ActionList = styled.div`
   width: 100%;
+  height: 100%;
```
Fosol, line 5: @jdtoombs 100% is required if a child wants to use 100%.

libs/npm/core/src/components/flex/col/styled/Col.tsx
```diff
@@ -2,7 +2,9 @@ import styled from 'styled-components';
 
 import { IColProps } from '..';
 
-export const Col = styled.div<IColProps>`
+export const Col = styled.div.attrs<IColProps>(({ className }) => ({
```
Fosol, line 5: @jdtoombs check out this syntax that allows us to correctly apply inline attributes to a styled component.  Before adding this any attributes we included in our code were being ignored.

PR #884, opened 2023-07-06
TNO-1477 Add report content ordering

Fosol:
The backend now supports ordering content within a report.  We still do not have a UI designed for creating custom reports, but once we do, it should all work.  This update also refactored parts of the Report Admin Form to make it easier to maintain.

> Run the `make renew` command to update your local environments

## Summary

- Add DB migration 1.0.40
- Publish tno-core:0.0.268
- Fix bug where saving a report instance was failing to also save the content in the report
- Update Reporting Service
- Added new endpoints in API

image


app/editor/src/features/admin/reports/ReportForm.tsx
```diff
@@ -1,59 +1,31 @@
-import 'prismjs/themes/prism.css';
```
Fosol, line 1: Extracted each tab into a separate component to make it easier to maintain.

app/editor/src/features/admin/reports/ReportFormInstances.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * @returns Component.
+ */
+export const ReportFormInstance: React.FC = () => {
+  const { values } = useFormikContext<IReportModel>();
+  const [, { findInstancesForReportId }] = useReports();
+
+  const [instances, setInstances] = React.useState<IReportInstanceModel[]>([]);
+
+  React.useEffect(() => {
+    findInstancesForReportId(values.id).then((instances) => {
```
Fosol, line 19: This will fire each time user clicks on the instance tab.

libs/net/dal/Migrations/20230706124343_1.0.41.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace TNO.DAL.Migrations
+{
+    /// <inheritdoc />
+    public partial class _1041 : SeedMigration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
+            migrationBuilder.AddColumn<int>(
```
Fosol, line 15: Adding `ReportInstanceContent.SortOrder` to allow users to manually control the order.  A report that is dynamically generated will always set the value to `0`.

libs/net/dal/Services/Interfaces/IReportService.cs
```diff
@@ -5,6 +5,5 @@ namespace TNO.DAL.Services;
 public interface IReportService : IBaseService<Report, int>
 {
     IEnumerable<Report> FindAll();
-    Report? FindById(int id, bool includeInstances);
```
Fosol, line 8: Don't need to include instances anymore.

libs/net/dal/Services/ReportInstanceService.cs
```diff
@@ -25,20 +26,102 @@ public ReportInstanceService(TNOContext dbContext, ClaimsPrincipal principal, IS
     public override ReportInstance? FindById(long id)
     {
         return this.Context.ReportInstances
+            .Include(ri => ri.Owner)
```
Fosol, line 29: This will be important when users are manually adding content to a report.

libs/net/entities/ReportInstanceContent.cs
```diff
@@ -39,6 +39,12 @@ public class ReportInstanceContent : AuditColumns
     /// </summary>
     [Column("section_name")]
     public string? SectionName { get; set; } = "";
+
+    /// <summary>
+    /// get/set - The order the content is returned.
+    /// </summary>
+    [Column("sort_order")]
+    public int SortOrder { get; set; }
```
Fosol, line 47: Here is the new property that resulted in all the work related to this PR.

kylgarmor (approved):
Nothing bad jumping out at me here

PR #885, opened 2023-07-07
feat: #tno-1530 - change UID on content

kylgarmor:
- changed from using various different things in different ingests
- ingests will use some variation of a has on Source+Title+PublishDate
- tweaked filters on when to update contentreference if it is found
- if a contentreference is found it may be a duplicate based on the has as opposed to reimporting the same content

### DIFF
--- a/services/net/syndication/SyndicationAction.cs
+++ b/services/net/syndication/SyndicationAction.cs
@@ -16,6 +16,7 @@ using TNO.Services.Actions;
 using TNO.Services.Syndication.Config;
 using TNO.Services.Syndication.Xml;
 using HtmlAgilityPack;
+using TNO.Core.Extensions;
 
 namespace TNO.Services.Syndication;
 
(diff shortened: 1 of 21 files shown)


libs/net/dal/Migrations/20230707172534_1.0.42.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace TNO.DAL.Migrations
+{
+    /// <inheritdoc />
+    public partial class _1042 : SeedMigration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
+            migrationBuilder.AddColumn<string>(
```
Fosol, line 15: You need to update the `ContentConfiguration` and set the max length of the column.  This will require you to update your DB migration 1.0.42.  To do this you will need to do the following in the `/libs/net` folder.
`make db-rollback n=1.0.41`
`make db-remove`
`make db-add n=1.0.42`
`make db-update`

Fosol (changes requested):

services/net/syndication/SyndicationAction.cs
```diff
@@ -86,20 +86,24 @@ private async Task ImportFeedAsync(IIngestServiceActionManager manager, Syndicat
...
+                var reference = await this.FindContentReferenceAsync(manager.Ingest.Source?.Code, sourceContent.Uid);
                 if (reference == null)
                 {
-                    reference = await AddContentReferenceAsync(manager.Ingest, item);
+                    reference = await AddContentReferenceAsync(manager.Ingest, item, sourceContent);
+
                     await FetchContent(manager.Ingest, item, link);
                 }
-                else if ((reference.Status == (int)WorkflowStatus.InProgress && reference.UpdatedOn?.AddHours(1) < DateTime.UtcNow) ||
+                else if (reference.Status != (int)WorkflowStatus.Imported &&
```
Fosol, line 99: If I'm reading this correctly it would avoid all `Imported` references.  This would stop the logic of the second part of this statement related to content stuck in `Received`.  Not certain this is desirable.

kylgarmor: Yah this is the complicated condition I was talking about. The idea is to not touch content that is already in the Imported status.  But it has so many other conditions that I feel it's a bit tangled.  Might be more readable if it was broken into a series of else-if for each condition maybe?

Fosol: Up to you, but your change would break the existing behaviour.

Fosol (approved):

PR #886, opened 2023-07-07
TNO-1447: Search by alias

jdtoombs:
Tried many variations of an elastic query to match aliases, given that this ticket is a "3 pointer" I settled with being selective about our aliases for now. This will at least eliminate the issue we were having before.

- my minister filter will now search by the aliases rather than the name 

I would like to eventually have it loop through the comma separated list and query each piece of text as an exact search; however, it was not as intuitive as I was hoping it was.


libs/net/dal/Migrations/1.0.43/Up/PreUp/00-Minister.sql
```diff
@@ -0,0 +1,5 @@
+DO $$
+BEGIN
+ALTER TABLE public."minister" ADD COLUMN "aliases" TEXT;
```
Fosol, line 3: I don't understand why you are doing this.  The DB migration should handle this.

libs/net/dal/Migrations/20230707180726_1.0.43.cs
```diff
@@ -0,0 +1,189 @@
...
+namespace TNO.DAL.Migrations
+{
+    /// <inheritdoc />
+    public partial class _1043 : SeedMigration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
+            migrationBuilder.AlterColumn<long>(
```
Fosol, line 16: I don't understand why there are so many changes to the Minister table?

libs/net/entities/Minister.cs
```diff
@@ -13,6 +13,10 @@ public class Minister : BaseType<int>
     #region Constructors
     protected Minister() { }
 
+    /// <summary>
+    [Column("aliases")]
+    public string? Aliases { get; set; } = "";
```
Fosol, line 18: Don't make it nullable.  Set the default value in the `MinisterConfiguration` to be an empty string.

Fosol (changes requested):

libs/net/dal/Migrations/20230707180726_1.0.43.cs
jdtoombs: I'm not sure what happened here either, debugging it. I redid the migration after pulling in changes and it spat this out.

libs/net/dal/Migrations/1.0.43/Up/PreUp/00-Minister.sql
jdtoombs: Forgot to remove this!

libs/net/dal/Migrations/20230707180726_1.0.43.cs
Fosol: @CelloLiu @kylgarmor 

You need to correctly follow this process when removing invalid migrations that occur when another PR creates one.
Before getting the latest code and merging.
`make db-rollback n={prior version}`
`make db-remove`
Manually undo all changes to the `TNOContextModelSnapshot.cs` with git.
Get the latest code changes.
`make db-update`
`make db-add n={your new version}`
`make db-update`
Now test to make sure your rollback works
`make db-rollback n={prior version}`
`make db-update`

libs/net/dal/Configuration/MinisterConfiguration.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using TNO.Entities;
+
+public class MinisterConfiguration : BaseTypeConfiguration<Minister, int>
+{
+    public override void Configure(EntityTypeBuilder<Minister> builder)
+    {
+        builder.Property(m => m.Aliases).HasDefaultValueSql("''");
```
Fosol, line 11: Need to add a max length on strings too

Fosol (changes requested):
Just one small change still needed

Fosol (changes requested):
You'll need to update to `1.0.43`

jdtoombs:
@Fosol - updated just so you know

Fosol (approved):

PR #890, opened 2023-07-10
TNO-1483 Add chart template admin

Fosol:
I've added a very basic admin form to manage and create chart templates.  Chart templates use Razor syntax to convert content into JSON data.  The JSON data must match the Charts.js format.  This JSON data is then sent to our new Charts API which generates a graph image.

## Summary

- Add new admin forms in Editor app
- Add new API endpoints for chart templates
- Publish tno-core:0.0.271

image


app/editor/src/features/admin/actions/styled/ActionForm.tsx
```diff
@@ -8,6 +8,5 @@ export const ActionForm = styled(FormPage)`
 
   .back-button {
     align-self: start;
-    margin-bottom: 3%;
```
Fosol, line 11: Removed this from most places as it was a wasting vertical space.  @jdtoombs was this required for some reason?

PR #899, opened 2023-07-11
My Minister bug

jdtoombs:
Frustrating one as it is not occurring locally still... will have to see how this behaves in dev

### DIFF
--- a/app/subscriber/src/features/my-minister/MyMinister.tsx
+++ b/app/subscriber/src/features/my-minister/MyMinister.tsx
@@ -68,12 +68,13 @@ export const MyMinister: React.FC = () => {
 
   /** retrigger content fetch when change is applied */
   React.useEffect(() => {
+    if (!aliases.length) return;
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/my-minister/MyMinister.tsx
```diff
@@ -68,12 +68,13 @@ export const MyMinister: React.FC = () => {
 
   /** retrigger content fetch when change is applied */
   React.useEffect(() => {
```
Fosol, line 70: Isn't this the 3rd PR related to this?

Fosol (approved):

app/subscriber/src/features/my-minister/MyMinister.tsx
jdtoombs: Correct yes, annoyingly have to throw things against the wall since I can't recreate locally :(

PR #905, opened 2023-07-14
TNO-1507: UI for advanced search

jdtoombs:
Finally dug this out of my stash, was a bit of a nightmare.

This is the *intro* PR for the UI components before made functional with a new ticket this sprint. Leaving the old search bar in the meantime as well. A few questions for Bobbi as well based off the mockup and what happens when clicking certain elements.

Would also like to add some smoother animations; however, that can be down the road when the barebones is complete.

demoadvanced


Fosol (approved):
Looks nice

PR #913, opened 2023-07-18
feat: #1562 - add configurable delay for migration

kylgarmor:
- we can now add a delay on how "up to date" a migration can be kept
- use as a buffer between updates in TNO and migrations into MMIA


jdtoombs (approved):

jdtoombs:
Could be handy!

PR #915, opened 2023-07-19
feat: #tno-1489 - no re-alerting on migration

kylgarmor:
- after discussion with stakeholder (Scott) we can ignore the alert attribute
- it's enough that an Editor can trigger a new alert manually if they choose


jdtoombs (approved):

jdtoombs:
Keeps things simple 😃

PR #918, opened 2023-07-20
fix: #tno-1459 - channel was not being updated

kylgarmor:
- switch to using react-router-dom, inline with the content page
- also refactored loading state


app/editor/src/features/content/papers/Papers.tsx
```diff
@@ -120,21 +117,26 @@ export const Papers: React.FC<IPapersProps> = (props) => {
...
     async (filter: IPaperFilter & Partial<IContentListAdvancedFilter>) => {
       try {
-        setLoading(true);
-        const data = await findContent(
-          makeFilter({
-            ...filter,
-          }),
-        );
-        return new Page(data.page - 1, data.quantity, data?.items, data.total);
+        if (!isLoading) {
```
CelloLiu, line 120: If isLoading == true, the finally section will set it to false right away, is that what expected?

app/editor/package.json
```diff
@@ -122,7 +122,8 @@
     ],
     "rules": {
       "simple-import-sort/imports": "error",
-      "simple-import-sort/exports": "error"
+      "simple-import-sort/exports": "error",
+      "no-console": ["error", { "allow": ["error", "debug"] }]
```
Fosol, line 126: Thanks

app/editor/src/features/content/papers/Papers.tsx
```diff
@@ -120,21 +117,24 @@ export const Papers: React.FC<IPapersProps> = (props) => {
...
       } catch (error) {
         // TODO: Handle error
         throw error;
       } finally {
-        setLoading(false);
+        setIsLoading(false);
       }
     },
-    [findContent],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Fosol, line 136: What is being ignored in the dependencies?  We always add documentation to ensure we're not introducing a hidden bug.

app/editor/package.json
jdtoombs: Gonna be handy :)

PR #919, opened 2023-07-22
Tno-1565 - hide inactive ministers on subscriber site

kylgarmor:
- hidden now when no longer active
- minister is removed from preferences if set to inactive
- updated to latest tno-core
- sort my ministers list by `name`


app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -13,7 +13,9 @@ export const MyMinisterSettings: React.FC = () => {
   const [, store] = useAppStore();
   const api = useUsers();
 
-  const options = ministers.map((m) => new OptionItem(`${m.name} | ${m.description}`, m.name));
+  let activeMinisters = ministers
+    .filter((m) => m.isEnabled)
+    .sort((a, b) => a.name.localeCompare(b.name));
```
kylgarmor, line 19: I feel like maybe we should modify the MinisterModel so that it has a first name and last name field.  name is then set as the concat of those two on save.  This could lead to a couple of other improvements:
- sort by last name
- the last name could also be used as an automatic alias as well

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -31,30 +33,58 @@ export const MyMinisterSettings: React.FC = () => {
     if (userInfo?.preferences?.myMinisters) {
       setMyMinisters(userInfo?.preferences?.myMinisters);
     }
-  }, [userInfo, ministers]);
+
```
kylgarmor, line 36: this just checks to see if the current ministers they have saved in their preferences are still active and if any of the ministers have changed status, then they are automatically removed from their preferences withoyt the user having to take any action

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -31,30 +33,58 @@ export const MyMinisterSettings: React.FC = () => {
...
+  }, [activeMinisters, api, ministers, store, userInfo]);
 
   return (
     <styled.MyMinisterSettings>
       <p className="description">
-        Choose the Minister you'd like to follow. Stories about your selected Minister will be
+        Choose the Minister(s) you'd like to follow. Stories about your selected Minister(s) will be
         available from a quick click in the sidebar menu.
       </p>
+      <p>Please note that "Search Aliases" will also show up in your "My Minister" feed.</p>
```
kylgarmor, line 68: Added a hint around the new text under the minister name and position.

app/subscriber/src/features/settings/styled/MyMinisterSettings.tsx
```diff
@@ -5,16 +5,25 @@ export const MyMinisterSettings = styled(Col)`
...
-
-  .options {
-    margin-bottom: 10px;
   }
 
   .option-container {
+    margin-top: 2em;
     background-color: ${(props) => props.theme.css.lightGray};
   }
+  .chk-container {
```
kylgarmor, line 14: make sure that the aliases always stay under the minister name and position

app/editor/package.json
```diff
@@ -56,7 +56,7 @@
     "react-tooltip": "5.10.0",
     "redux-logger": "3.0.6",
     "styled-components": "5.3.9",
-    "tno-core": "0.0.276"
+    "tno-core": "0.0.279"
```
kylgarmor, line 59: Used the tno-core version from Jonny's PR since it fixes the missing `aliases` field on `IMinisterModel`

app/subscriber/src/features/settings/MyMinisterSettings.tsx
Fosol: Automatic alias isn't ideal regrettably, due to common names.  It's better to force the admin to choose the appropriate aliases to limit false positives.
Splitting first and last name could be useful for sorting, but isn't critical presently.

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -31,30 +33,58 @@ export const MyMinisterSettings: React.FC = () => {
...
-              onChange={(e) => {
-                if ((e.target as HTMLInputElement).checked) {
-                  setMyMinisters([...myMinisters, o.value]);
-                } else {
-                  // remove from prefrerences when unchecking
-                  setMyMinisters(myMinisters.filter((m: string) => m !== o.value));
-                }
-              }}
-            />
+            <div className="chk-container">
```
Fosol, line 72: Need to move the `key` to the outer parent.  Always test with the developer tools open in the browser to see new errors and warnings being introduced by changes.

app/subscriber/src/store/hooks/subscriber/interfaces/IMinisterModel.ts
```diff
@@ -1,7 +1,6 @@
 import { ISortableModel } from 'tno-core';
 
 export interface IMinisterModel extends ISortableModel<number> {
-  name: string;
```
Fosol, line 4: Thanks for cleaning up.  @jdtoombs note that inherited properties don't need to be duplicated.

libs/net/models/Areas/Subscriber/Models/Minister/MinisterModel.cs
```diff
@@ -17,6 +17,7 @@ public class MinisterModel : BaseTypeModel<int>
     public MinisterModel() { }
 
     public string? Aliases { get; set; } = "";
```
Fosol, line 19: Should not be nullable.

libs/net/models/Areas/Subscriber/Models/Minister/MinisterModel.cs
```diff
@@ -17,6 +17,7 @@ public class MinisterModel : BaseTypeModel<int>
...
+    public string? Position { get; set; } = "";
```
Fosol, line 20: Should not be nullable.

Fosol (changes requested):

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -31,30 +33,58 @@ export const MyMinisterSettings: React.FC = () => {
...
-                if ((e.target as HTMLInputElement).checked) {
-                  setMyMinisters([...myMinisters, o.value]);
-                } else {
-                  // remove from prefrerences when unchecking
-                  setMyMinisters(myMinisters.filter((m: string) => m !== o.value));
-                }
-              }}
-            />
+            <div className="chk-container" key={o.name}>
+
```
Fosol, line 73: Linting issue

Fosol (approved):

PR #925, opened 2023-07-25
fix: #tno-1524 - only update content status

kylgarmor:
Update the API endpoint and Content Service to only update the `Content.Status` to eliminate data corruption or concurrency issues.

### DIFF
--- a/api/net/Areas/Services/Controllers/ContentController.cs
+++ b/api/net/Areas/Services/Controllers/ContentController.cs
@@ -252,6 +252,23 @@ public class ContentController : ControllerBase
 
         return new JsonResult(new ContentModel(content));
     }
+    /// <summary>
+    /// Update content for the specified 'id'.
+    /// Will not trigger any re-index or audit trail update
+    /// </summary>
+    /// <param name="model"></param>
+    /// <returns></returns>
+    [HttpPut("{id}/status")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(ContentModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Content" })]
+    public IActionResult UpdateStatusAsync(ContentModel model)
+    {
+        var content = _contentService.UpdateStatusOnly((Content)model);
+
+        return new JsonResult(new ContentModel(content));
+    }
 
     /// <summary>
     /// Upload a file and link it to the specified content.
--- a/libs/net/dal/Services/ContentService.cs
+++ b/libs/net/dal/Services/ContentService.cs
@@ -223,7 +223,7 @@ public class ContentService : BaseService<Content, long>, IContentService
 
         if (filter.Sentiment.Any())
             filterQueries.Add(s => s.Nested(n => n.Path(p => p.TonePools).Query(y => y.Range(m => m.Field("tonePools.value").GreaterThanOrEquals(filter.Sentiment.First()).LessThanOrEquals(filter.Sentiment.Last())))));
-        
+
         if (filter.ProductIds.Any())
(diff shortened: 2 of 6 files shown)


kylgarmor:
@Fosol I wasnt sure whether this was the right approach here.  Would like some feedback if you have time and can point me in the right direction if I have gone off course.

api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -252,6 +252,24 @@ public async Task<IActionResult> UpdateAsync(ContentModel model, bool index = fa
...
+    /// </summary>
+    /// <param name="model"></param>
+    /// <param name="newStatus">New ContentStatus.</param>
+    /// <returns></returns>
+    [HttpPut("{id}/status")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(ContentModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Content" })]
+    public IActionResult UpdateStatusAsync(ContentModel model, ContentStatus newStatus)
```
Fosol, line 267: I would just include the content model.  Whatever status is in the model should be the new status.  But it can be either way.

libs/net/dal/Services/ContentService.cs
```diff
@@ -489,5 +489,19 @@ public IEnumerable<NotificationInstance> GetNotificationsFor(long contentId)
...
     }
+
+    /// <summary>
+    /// Update the specified 'entity' in the database.
+    /// </summary>
+    /// <param name="entity"></param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    /// TODO: Switch to not found exception throughout services.
+    public Content UpdateStatusOnly(long contentId, ContentStatus newStatus)
```
Fosol, line 500: The main issue here is you're not accounting for concurrency issues.  Is this intentional?  If so why?  In most cases we enforce optimistic concurrency.

api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -252,6 +252,24 @@ public async Task<IActionResult> UpdateAsync(ContentModel model, bool index = fa
...
+    {
+        var content = _contentService.UpdateStatusOnly(model.Id, newStatus);
```
Fosol, line 269: I would probably lean to not bothering with a dedicated function for this.  I would just make a request for the entity in the db and change the status, then update.  But there can be benefits to do it your way.

libs/net/dal/Services/ContentService.cs
```diff
@@ -501,5 +501,22 @@ public IEnumerable<NotificationInstance> GetNotificationsFor(long contentId)
...
             .Where(n => n.ContentId == contentId);
     }
+
+    /// <summary>
+    /// Update the specified 'entity' in the database.
+    /// </summary>
+    /// <param name="updated"></param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    /// TODO: Switch to not found exception throughout services.
```
Fosol, line 511: That would be great

libs/net/dal/Services/ContentService.cs
```diff
@@ -501,5 +501,22 @@ public IEnumerable<NotificationInstance> GetNotificationsFor(long contentId)
...
+    public Content UpdateStatusOnly(Content updated)
+    {
+        var original = FindById(updated.Id) ?? throw new InvalidOperationException("Entity does not exist");
+
+        this.Context.Entry(original.Status).CurrentValues.SetValues(updated.Status);
+        this.Context.ResetVersion(original);
```
Fosol, line 517: Is the desired behavior to ignore concurrency?  This will result in this operation to always 'win', essentially ignoring any changes that were made by other processes.  Which could result in loss of data.

kylgarmor: removed `ResetVersion()` call as per discussion

Fosol (approved):

PR #926, opened 2023-07-26
fix: #tno-1576 - hide aliases on my ministers page

kylgarmor:
- also make minister name bold

### DIFF
--- a/app/subscriber/src/features/settings/MyMinisterSettings.tsx
+++ b/app/subscriber/src/features/settings/MyMinisterSettings.tsx
@@ -15,7 +15,7 @@ export const MyMinisterSettings: React.FC = () => {
 
   let activeMinisters = ministers
     .filter((m) => m.isEnabled)
-    .sort((a, b) => a.name.localeCompare(b.name));
+    .sort((a, b) => a.sortOrder - b.sortOrder || a.name.localeCompare(b.name));
 
   const handleSubmit = async (values: number[]) => {
     try {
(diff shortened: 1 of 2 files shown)


app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -15,7 +15,7 @@ export const MyMinisterSettings: React.FC = () => {
 
   let activeMinisters = ministers
     .filter((m) => m.isEnabled)
-    .sort((a, b) => a.name.localeCompare(b.name));
+    .sort((a, b) => a.sortOrder - b.sortOrder);
```
Fosol, line 18: Wouldn't this only sort by the `sortOrder`?  Doesn't the other function also sort by the `name`?

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -84,7 +88,7 @@ export const MyMinisterSettings: React.FC = () => {
                   }
                 }}
               />
-              <div className="aliases">Search Aliases: {o.aliases}</div>
+              <div className="position">{o.position}</div>
```
Fosol, line 91: Was this requested?  I don't think we have the `position` data presently, which would result in this being blank.

app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -15,7 +15,7 @@ export const MyMinisterSettings: React.FC = () => {
 
   let activeMinisters = ministers
     .filter((m) => m.isEnabled)
-    .sort((a, b) => a.name.localeCompare(b.name));
+    .sort((a, b) => a.sortOrder - b.sortOrder || a.name.localeCompare(b.name));
```
Fosol, line 18: Won't this ignore `sortOrder` if `=-1 || 0` and then use the `name` instead?  The `sortOrder` should be the priority sort`.

app/subscriber/src/features/settings/MyMinisterSettings.tsx
kylgarmor: Yah we have `position` data now.  Carolynn has been entering it.

When you are back next week we can talk about the `description` field on the `ministers` table.  It doesn't have a purpose currently, so could arguably be hidden in the Editor UI.
Also, the Organization table/relationship you added before you left isn't quite that straight forward apparently.  Carolynn showed me that there are cases where a minister can have multiple parent orgs :(

Fosol: They don't have any of this functionality in the existing product, so we should limit how much we add for MVP.

app/subscriber/src/features/settings/MyMinisterSettings.tsx
kylgarmor: I honestly wasn't too sure how that worked either, but it does..
image

image

Fosol (approved):

PR #931, opened 2023-07-27
fix: #tno-1582 - aliases not showing up

kylgarmor:
- was only search for the value stored in preferences.myMinisters
- retrieve the minister record and then search for name AND aliases


jdtoombs (approved):
Thanks!

PR #932, opened 2023-07-27
TNO-1553,1552,1516:Create/manage folders, filter by sentiment

jdtoombs:
_**Known bug:**_ content not being returned by api on attached folders, Will create ticket for next sprint

Also a note for @Fosol. we are probably moving advanced search out of the sidebar so things aren't as squished.

**In this PR:**

1) can now filter out content based on their tone in the advanced search
sentiment

2) my folders section allows you to manage/update folders
foldermanage

3) add to folder button when viewing content now allows you to add content to existing folder, or a new folder
addtofolder


app/subscriber/src/features/home/utils/makeFilter.ts
```diff
@@ -23,6 +23,7 @@ export const makeFilter = (
     productIds: filter.productIds ?? undefined,
     publishedEndOn: filter.endDate ? filter.endDate : undefined,
     publishedStartOn: filter.startDate ? filter.startDate : undefined,
+    sentiment: filter.sentiment ?? undefined,
```
Fosol, line 26: Isn't filter an array or undefined?  This doesn't seem like a relevant operation.  Just use `filter.sentiment` I think.

app/subscriber/src/features/my-folders/styled/MyFolders.tsx
```diff
@@ -0,0 +1,79 @@
...
+    overflow: hidden;
+    .re-name {
+      height: 1.5em;
+      padding: 0;
+      &:focus {
+        box-shadow: none;
+      }
+    }
+    .group {
+      background-color: #f9f9f9 !important;
```
Fosol, line 37: Can you check if this is really the only way to do this?

libs/net/dal/Models/ContentFilter.cs
```diff
@@ -110,6 +110,12 @@ public class ContentFilter : PageFilter
     /// </summary>
     public string? Section { get; set; }
 
+    /// <summary>
+    /// get/set - The content sentitment.
+    /// </summary>
+    public long[] Sentiment { get; set; }
```
Fosol, line 116: Fix nullable issues.

Fosol (changes requested):

app/subscriber/src/features/my-folders/styled/MyFolders.tsx
jdtoombs: On it

jdtoombs:
Updated :) @Fosol

libs/net/dal/Models/ContentFilter.cs
```diff
@@ -173,6 +173,11 @@ public class ContentFilter : PageFilter
     /// </summary>
     public string[] Sort { get; set; } = Array.Empty<string>();
 
+    /// <summary>
+    /// get/set - The content sentitment.
+    /// </summary>
+    public long[] Sentiment { get; set; } = Array.Empty<long>();
```
Fosol, line 179: Is there any reason to use a `long`?  I think the value will never be more than -5 to +5.

jdtoombs: Copy and paste d'oh mistake, int now :)

Fosol (approved):

PR #935, opened 2023-07-31
feat: #tno-1579 - insert and update ministers

kylgarmor:


kylgarmor:
Please leave for Kyle to Merge once approved.  Need to do some manual data handling in the DEV environment which will not be neccesary in other environments

app/editor/.yarn/install-state.gz
kylgarmor, line 1: There's a build issue in the pipeline at the moment and I think it might be caused by this file being out of sync

CelloLiu: It seems we might not need this file?

CelloLiu (approved):

PR #938, opened 2023-08-01
fix: #tno-1588 - fixed search on name and alias

kylgarmor:
- search alias is generated dynamically
- based on first letter of first name plus last name seperated by period
- we can hide the aliases column in the UI - keep for synonyms?


libs/net/dal/Models/ContentFilter.cs
```diff
@@ -145,6 +145,9 @@ public class ContentFilter : PageFilter
     /// </summary>
     public string? Keyword { get; set; }
 
+    // get/set - Get/set the namess to search for (tied to ministers).
```
Fosol, line 148: Spelling `namess`.  What's the difference between `Names` and `Aliases`?

libs/net/dal/Services/ContentService.cs
```diff
@@ -265,6 +266,25 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
...
+                var nameSegments = name.Split(' ');
+                var firstName = nameSegments.First();
+                var firstInitial = firstName.Substring(0, 1);
+                var lastName = nameSegments.Last();
+                //(eby +(david | d.))
+                nameQueries.Add($"({lastName} +({firstName} | {firstInitial}.))");
+            }
+            filterQueries.Add(s =>
+                s.SimpleQueryString(qs => qs
+                .Fields(f => f.Field(p => p.Headline, 10).Field(p => p.Byline, 10).Field(p => p.Body).Field(p => p.Summary))
```
Fosol, line 284: What does the `10` do?

libs/net/dal/Services/ContentService.cs
```diff
@@ -265,6 +266,25 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
...
+        if (!string.IsNullOrWhiteSpace(filter.Names))
+        {
+            var names = filter.Names.Split(',').Select(a => a.Trim()).ToList();
+            List<string> nameQueries = new();
+            foreach (var name in names)
+            {
+                var nameSegments = name.Split(' ');
+                var firstName = nameSegments.First();
+                var firstInitial = firstName.Substring(0, 1);
+                var lastName = nameSegments.Last();
```
Fosol, line 278: This logic can fail under certain conditions.
`"firstname"` would be a valid but would result in `firstName == lastName`.

libs/net/dal/Services/ContentService.cs
```diff
@@ -265,6 +266,25 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
...
+                //(eby +(david | d.))
+                nameQueries.Add($"({lastName} +({firstName} | {firstInitial}.))");
+            }
+            filterQueries.Add(s =>
+                s.SimpleQueryString(qs => qs
+                .Fields(f => f.Field(p => p.Headline, 10).Field(p => p.Byline, 10).Field(p => p.Body).Field(p => p.Summary))
+                .Query(string.Join(" | ", nameQueries).ToLowerInvariant())));
```
Fosol, line 285: What's the point of using `ToLowerInvariant()`?

libs/net/dal/Models/ContentFilter.cs
kylgarmor: good catch on the `namess` typo.

Names requires a different type of search than just keyword and the name is now used to dynamically generate the alias on the fly.  Had a conversation with the business yesterday and the dynamic alias we generate is sufficient for now.  Having said that, in some future state we may re-use the alias field for configuring more synonyms in elasticsearch.

libs/net/dal/Services/ContentService.cs
kylgarmor: the numbers after the name are the `boost` value which gives matches in those fields a higher score which helps determine the ranking in the results

libs/net/dal/Services/ContentService.cs
kylgarmor: Are you referring to a scenario where the `name` of the individual is a single word like "Beyonce"?

libs/net/dal/Services/ContentService.cs
kylgarmor: An old habit I had from a previous life where we were consuming text in multiple languages.  Everything had to be invariant culture.  It's not relevant here.

libs/net/dal/Services/ContentService.cs
Fosol: yes

libs/net/dal/Services/ContentService.cs
kylgarmor: removed the `ToLower()` altogether and it seems to work fine

Fosol (approved):

PR #943, opened 2023-08-04
TNO-1486 Customize charts in Reports

Fosol:
The Reporting Service can now communicate with the Charts API to dynamically generate charts and include them in emails.

## Summary

- Updated reporting service
- Updated notification service
- Published `tno-core:0.0.288`
- Update VSCode settings to improve intellisense with C#.  Be aware that it only works for files that are open.
- Fixed bugs in folder/charts/reports admin forms
- Centralized some shared code between reports and charts


app/editor/src/features/admin/charts/ChartTemplateContext.tsx
Fosol, line 1: @jdtoombs This is what we should be doing when we have a bunch of related components that need to share state, but where we don't really want to have to also keep this state in redux.

Redux state is for application wide data
Context state is for related component data

libs/net/dal/Migrations/1.0.52/Up/PostUp/00-ChartTemplate.sql
```diff
@@ -22,7 +22,7 @@ INSERT INTO public."chart_template" (
 @using System.Linq
 @using TNO.Entities
 @{
-  var group = Settings.GroupBy switch
+  var group = SectionSettings.GroupBy switch
```
Fosol, line 25: This won't fix it for DBs that already have applied the migration, but we don't have any supported charts yet, so it can be resolved manually.

libs/npm/core/src/hooks/api-editor/interfaces/IChartRequestModel.ts
```diff
@@ -1,10 +1,6 @@
 import { IChartSettingsModel, IContentModel } from '.';
 
 export interface IChartRequestModel {
-  /** Width of the chart image. */
-  width?: number;
```
Fosol, line 5: Moved to the `settings`

PR #944, opened 2023-08-04
Fix JsonDocument extension method and elastic migration tool

Fosol:
I introduced a regression issue when I updated the extension methods for JsonDocument and JsonElement.  It would be wonderful if we could find time to add unit-tests for these things.

This bug was causing all elastic migrations to fail.


libs/net/core/Extensions/JsonElementExtensions.cs
```diff
@@ -73,7 +73,7 @@ public static string ToJson(this JsonElement element, JsonWriterOptions? options
     /// <returns></returns>
     public static T? GetElementValue<T>(this JsonElement element, string path = "", T? defaultValue = default, JsonSerializerOptions? options = null)
     {
-        var property = GetElement(element, path);
+        var property = path.Contains('.') ? GetElement(element, path) : element;
```
Fosol, line 76: We only need to traverse the path if it is more than one level deep.

PR #946, opened 2023-08-04
feat: #tno-1558 - connect to PROD oracle instance

kylgarmor:
- needed to introduce schema as an attribute for oracle
- allso added setup for running an oracle client for testing

### DIFF
--- a/services/net/contentmigration/sources/oracle/MigrationSourceContext.cs
+++ b/services/net/contentmigration/sources/oracle/MigrationSourceContext.cs
@@ -10,7 +10,7 @@ namespace TNO.Services.ContentMigration.Sources.Oracle;
 public class MigrationSourceContext : DbContext
 {
     private readonly IConfiguration configuration;
-    private string Schema => configuration.GetValue("Service:OracleConnection:DefaultSchema", "");
+    private string DefaultSchema => configuration.GetValue("Service:OracleConnection:DefaultSchema", "");
 
     #region Properties
     /// <summary>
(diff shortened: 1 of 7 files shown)


services/net/contentmigration/sources/oracle/MigrationSourceContext.cs
```diff
@@ -74,6 +79,19 @@ protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
     /// <param name="modelBuilder"></param>
     protected override void OnModelCreating(ModelBuilder modelBuilder)
     {
+        modelBuilder.HasDefaultSchema(Schema);
+
+        if (string.IsNullOrWhiteSpace(Schema))
```
Fosol, line 84: Change to a simpler easier to maintain solution by using string interpolation.  Having multiple sql statements to maintain will always result in regression errors.

`.ToSqlQuery($"WITH ATN AS (SELECT * from {(!String.IsNullOrEmpty(Schema) ? $"{Schema}." : "")}.NEWS_ITEMS UNION ALL SELECT * from  {(!String.IsNullOrEmpty(Schema) ? $"{Schema}." : "")}.HNEWS_ITEMS) SELECT * FROM ATN")

Fosol (changes requested):

services/net/contentmigration/sources/oracle/MigrationSourceContext.cs
Fosol: Or cleaner would be
`var schema = (!String.IsNullOrEmpty(Schema) ? $"{Schema}." : "";
`.ToSqlQuery($"WITH ATN AS (SELECT * from {schema}.NEWS_ITEMS UNION ALL SELECT * from  {schema}.HNEWS_ITEMS) SELECT * FROM ATN"`

kylgarmor: option 2 is better for sure.

Fosol (approved):

PR #947, opened 2023-08-05
feat: #tno-1558 - connect to PROD oracle instance

kylgarmor:
- added logging on main query
- fixed sort issue and use of lastModified instead of published

### DIFF
--- a/services/net/contentmigration/ContentMigrationAction.cs
+++ b/services/net/contentmigration/ContentMigrationAction.cs
@@ -169,6 +169,7 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
         catch (Exception)
         {
             // migration has never been run before.
+            this.Logger.LogInformation("No creationDateOfLastImport found for ingest '{name}'", manager.Ingest.Name);
         }
         double migrationTimeOffsetInHours = 0;
         try
@@ -177,7 +178,7 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
         }
         catch (Exception)
         {
-            // migration has never been run before.
+            this.Logger.LogInformation("No migrationTimeOffsetInHours found for ingest '{name}'", manager.Ingest.Name);
         }
 
         if (manager.Ingest.SourceConnection != null)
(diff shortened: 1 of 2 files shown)


services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -153,17 +161,35 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
...
-        DateTime? creationDateOfLastImport = !string.IsNullOrEmpty(manager.Ingest.GetConfigurationValue("creationDateOfLastImport")) ? manager.Ingest.GetConfigurationValue<DateTime>("creationDateOfLastImport") : null;
-        double migrationTimeOffsetInHours = !string.IsNullOrEmpty(manager.Ingest.GetConfigurationValue("migrationTimeOffsetInHours")) ? manager.Ingest.GetConfigurationValue<double>("migrationTimeOffsetInHours") : 0;
+        DateTime? creationDateOfLastImport = null;
+        try
+        {
+            creationDateOfLastImport = manager.Ingest.GetConfigurationValue<DateTime>("creationDateOfLastImport");
+        }
+        catch (Exception)
+        {
+            // migration has never been run before.
```
Fosol, line 171: Ignoring exceptions is always risky behaviour.  You should almost always log.

services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -153,17 +161,35 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
...
+        }
+        double migrationTimeOffsetInHours = 0;
+        try
+        {
+            migrationTimeOffsetInHours = manager.Ingest.GetConfigurationValue<double>("migrationTimeOffsetInHours");
+        }
+        catch (Exception)
+        {
+            // migration has never been run before.
```
Fosol, line 180: Ignoring exceptions is always risky behaviour.  You should almost always log.

Fosol (approved):

services/net/contentmigration/ContentMigrationAction.cs
kylgarmor: sorted

services/net/contentmigration/ContentMigrationAction.cs
kylgarmor: sorted

PR #958, opened 2023-08-11
My Folders - Bugs

jdtoombs:
- fix tooltip appearing behind table
- fix concurrency issue
- clear folder name when making a new one


app/subscriber/src/features/my-folders/MyFolders.tsx
```diff
@@ -34,15 +34,15 @@ export const MyFolders = () => {
...
     });
   };
 
   const handleSave = () => {
     if (!!active) {
-      updateFolder(active).then(() => {
+      updateFolder(active).then((data) => {
         toast.success(`${active.name} updated successfully`);
-        setMyFolders([...myFolders.filter((folder) => folder.id !== active.id), active]);
+        setMyFolders([data]);
```
Fosol, line 45: Appears like this might be problematic.  This would result in loosing any other folders currently in state.  Is this what you want?

app/subscriber/src/features/my-folders/MyFolders.tsx
```diff
@@ -34,13 +34,13 @@ export const MyFolders = () => {
     }).then((data) => {
       toast.success(`${data.name} created successfully`);
       setNewFolderName('');
-      setMyFolders([...myFolders, data]);
+      setMyFolders([data]);
```
Fosol, line 37: This will cause a similar issue.  When you add a folder it will remove all other folders that were originally in the array.

app/subscriber/src/features/my-folders/MyFolders.tsx
```diff
@@ -40,7 +40,7 @@ export const MyFolders = () => {
 
   const handleSave = () => {
     if (!!active) {
-      updateFolder(active).then(() => {
+      updateFolder(active).then((data) => {
```
Fosol, line 43: `data` is never used.  `active` will not have correct `version`.  Edit this folder twice in a row will result in a failure.

Fosol (changes requested):

app/subscriber/src/features/my-folders/MyFolders.tsx
```diff
@@ -40,8 +40,8 @@ export const MyFolders = () => {
 
   const handleSave = () => {
     if (!!active) {
-      updateFolder(active).then(() => {
-        toast.success(`${active.name} updated successfully`);
+      updateFolder(active).then((data) => {
+        toast.success(`${data.name} updated successfully`);
         setMyFolders([...myFolders.filter((folder) => folder.id !== active.id), active]);
```
Fosol, line 45: This is the line that needs `data`.  I have a meeting for the next hour, but I can go over this with you later if you want?

Fosol (changes requested):

Fosol (approved):

PR #960, opened 2023-08-11
TNO-1177 Fix Image Service

Fosol:
With the latest changes to support Globe & Mail frontpage images, it broke the other newspaper front page images.  This update will fix this.

Requires manually updating the front page image ingest configuration values.


app/editor/src/features/admin/ingests/configurations/FrontPage.tsx
```diff
@@ -14,6 +14,13 @@ export const FrontPage: React.FC = (props) => {
         label="Path to Files"
         name="configuration.path"
         value={values.configuration.path}
+        tooltip="Path to files can include '<date>' which will be replaced with today's date"
+      />
+      <FormikText
```
Fosol, line 19: Need a way to dynamically use the date for the folder path.

PR #966, opened 2023-08-14
fix: #tno-1662 - replace pipe with p tags

kylgarmor:
TNO has pre-replace all paragraph markers with a pipe '|' character.  This fix, undoes this on the way into MMIA

### DIFF
--- a/services/net/contentmigration/migrators/PaperMigrator.cs
+++ b/services/net/contentmigration/migrators/PaperMigrator.cs
@@ -138,11 +138,7 @@ public class PaperMigrator : ContentMigrator<ContentMigrationOptions>, IContentM
             if (index !== -1)
             {
                 // found at least one linebreak marker
-                StringBuilder newBody = new StringBuilder();
-                newBody.Append("<p>");
-                newBody = text.Replace(marker, "</p><p>");
-                newBody.Append("</p>");
-                body = newBody.ToString();
+                body = $"<p>{text.Replace(marker, "</p><p>")}</p>";
             }
             else
             {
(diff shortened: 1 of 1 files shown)


services/net/contentmigration/migrators/PaperMigrator.cs
```diff
@@ -115,6 +121,42 @@ public override SourceContent CreateSourceContent(LookupModel lookups, SourceMod
...
+    /// <param name="summary"></param>
+    /// <returns>formatted body string</returns>
+    private string GetNewsItemBody(string text, string summary)
+    {
+        string body = string.Empty;
+
+        if (!string.IsNullOrEmpty(text))
+        {
+            string const marker = '|';
+            int index = text.IndexOf(marker);
```
Fosol, line 137: Don't change but this seems like something that would be nice to use configuration for.  If there are other special characters showing up, we can modify to handle with configuration instead of a code change.

services/net/contentmigration/migrators/PaperMigrator.cs
```diff
@@ -115,6 +121,42 @@ public override SourceContent CreateSourceContent(LookupModel lookups, SourceMod
...
+            if (index !== -1)
+            {
+                // found at least one linebreak marker
+                StringBuilder newBody = new StringBuilder();
+                newBody.Append("<p>");
+                newBody = text.Replace(marker, "</p><p>");
```
Fosol, line 143: Won't this result in an extract `<p></p>` every time?  Why can't we replace with `\n`?

Fosol: I think this is clearer `$"<p>{text.Replace(marker, "</p><p>")}</p>"`

Fosol (approved):

PR #979, opened 2023-08-17
fix: #tno-1682 - updated filters

kylgarmor:
- split out the filter for Print vs Online
- PrintContent now limited to records with any of these Types - { "Newspaper", "Regional" }
- Story now limited to records with any of these Types - { "CP News", "Internet" }

### DIFF
--- a/services/net/contentmigration/ContentMigrationAction.cs
+++ b/services/net/contentmigration/ContentMigrationAction.cs
@@ -200,7 +200,7 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
         {
             try
             {
-                var baseFilter = contentMigrator.GetBaseFilter(manager.Ingest.IngestType!.ContentType);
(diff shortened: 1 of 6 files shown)


services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -200,7 +200,7 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
         {
             try
             {
-                var baseFilter = contentMigrator.GetBaseFilter();
+                var baseFilter = contentMigrator.GetBaseFilter(manager.Ingest.IngestType!.ContentType);
```
Fosol, line 203: Using a `!` is dangerous as it's ignoring the lint warning.  Handle the nullable issue.

Fosol (changes requested):

Fosol (approved):

PR #983, opened 2023-08-18
feat: #tno-1625 - created new kustomize config

kylgarmor:
- added new HISTORIC container, existing will be used for RECENT
- added SupportedImportMigrationTypes config
- updated memory requiremenmts in OpenShift
- added config for audit user name


app/editor/src/features/admin/ingests/configurations/DbMigration.tsx
```diff
@@ -89,6 +92,22 @@ export const DbMigration: React.FC = (props) => {
...
         </Col>
       </Row>
+      <Row>
+        <Col flex="1 1 1">
+          <FormikRadioGroup
+            label="Migration Type"
+            name="ImportMigrationType"
+            value={!!values.configuration.importMigrationType ? 'Recent' : undefined}
+            onChange={(val) => {
+              isImportStartDateRequired = val.target.value == 'Recent';
```
Fosol, line 102: Fixing linting issues.  If you install some plugins with vscode you should see these before the PR.

Fosol (changes requested):
DB migration `1.0.57` is already taken.  Make sure you don't incorrectly handle resetting your environment.
`make db-rollback n=1.0.56`  Rollback your version
`make db-remove` Delete the files related to your 1.0.57
`git fetch upstream dev` Get the latest code
`git rebase upstream/dev` Rebase your commit at the end
`make db-update` Update your local environment to the latest
`make db-add n=1.0.58` Create a new migration for your changes
`make db-update` Update the database with your migration

Now test your migration rollback and update again.

app/editor/src/features/admin/ingests/configurations/DbMigration.tsx
kylgarmor: fixed.

openshift/kustomize/services/contentmigration-historic/base/config-map.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
Fosol, line 1: Couldn't you have a different overlay instead of another config?

Fosol (approved):

openshift/kustomize/services/contentmigration-historic/base/config-map.yaml
kylgarmor: Funny you should say that because I was thinking the same thing, but when I talked to Kulpreet about this he seemed to think this was easier.

PR #986, opened 2023-08-21
TNO-1667: Return all content on search page

jdtoombs:
Leaving in draft as I am blocking myself with some NPM changes in my other PR:

- this will set the quantity on the search page to the elastic search max of 10000


app/subscriber/src/features/search-page/SearchPage.tsx
```diff
@@ -107,6 +107,7 @@ export const SearchPage: React.FC = () => {
           makeFilter({
             ...filter,
             contentTypes: [],
+            pageSize: 10000,
```
Fosol, line 110: I'm not a fan of this solution, but for now it'll have to do.
Essentially we're asking Elasticsearch for the largest number of hits possible every time.  Ideally we'd only ask for a realistic number of stories and page through results.

jdtoombs: @Fosol this one ready for review now as well - although it is a "for now" solution as discussed

Fosol (approved):

PR #987, opened 2023-08-21
AV Overview: User feedback on save

jdtoombs:
A number of bugs I have this Sprint seemed to be resolved, throwing this in as a quick thing I found when using the AV overview...

### DIFF
--- a/app/editor/src/features/reports/av-overview/AVOverview.tsx
+++ b/app/editor/src/features/reports/av-overview/AVOverview.tsx
@@ -76,7 +76,9 @@ export const AVOverview: React.FC = () => {
       <FormikForm
         initialValues={instance}
         onSubmit={(values, { setSubmitting }) => {
-          handleSubmit(values);
+          handleSubmit(values)
+            .catch((err) => toast.warning(() => 'Failed to save.'))
+            .finally(() => toast.success(() => 'Evening overview saved.'));
           setSubmitting(false);
         }}
       >


app/editor/src/features/reports/av-overview/AVOverview.tsx
```diff
@@ -76,7 +76,9 @@ export const AVOverview: React.FC = () => {
       <FormikForm
         initialValues={instance}
         onSubmit={(values, { setSubmitting }) => {
-          handleSubmit(values);
+          handleSubmit(values)
+            .catch((err) => toast.warning(() => 'Failed to save.'))
```
Fosol, line 80: 1) Why a warning when it's an error?
2) Why doesn't the default toast work when submitting?  It should already be toasting the error message?

app/editor/src/features/reports/av-overview/AVOverview.tsx
```diff
@@ -76,7 +76,9 @@ export const AVOverview: React.FC = () => {
...
+            .finally(() => toast.success(() => 'Evening overview saved.'));
```
Fosol, line 81: This is incorrect.  Think about what `finally` means

Fosol (changes requested):

app/editor/src/features/reports/av-overview/AVOverview.tsx
jdtoombs: @Fosol - closin this since I see you fixed it

PR #994, opened 2023-08-22
TNO-1408 Fix MOV keyframe

Fosol:
Browsers have a client side feature that allows us to control the default keyframe loaded in a video file.  It appears that in the case of the default MOV file created by editors they are perhaps incorrectly creating the first frame, or they are not including a default image.  With controlling the keyframe it is essentially moving the head forward a little with the hope that an image can be displayed.  I have tested the example MOV file in Chrome and it works.

image


kylgarmor (approved):
nice find!

PR #995, opened 2023-08-22
feat: #Tno-1526 - Handle Updating Edited Content

kylgarmor:
- added metadata field to ContentReference


libs/net/models/Areas/Services/Models/ContentReference/ContentReferenceModel.cs
```diff
@@ -60,14 +66,16 @@ public class ContentReferenceModel : AuditColumnsModel
...
+    /// <param name="options"></param>
+    public ContentReferenceModel(Entities.ContentReference entity, JsonSerializerOptions options) : base(entity)
     {
         this.Source = entity.Source;
         this.Uid = entity.Uid;
         this.Status = (int)entity.Status;
         this.Topic = entity.Topic;
         this.Offset = entity.Offset;
         this.Partition = entity.Partition;
+        this.Metadata = JsonSerializer.Deserialize<Dictionary<string, object>>(entity.Metadata, options) ?? new Dictionary<string, object>();
```
Fosol, line 78: You should setup your linter so you catch these issues when coding.

services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -295,13 +297,46 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
...
+                {
+                    if (!DateTime.TryParse(reference.Metadata[UpdatedOnMetadataKey].ToString(), out originalLastUpdateDate))
+                        originalLastUpdateDate = DateTime.MinValue;
+                }
+
+                if (sourceContent.UpdatedOn > originalLastUpdateDate)
+                {
+                    isUpdatedSourceContent = true;
+                    reference.Status = (int)WorkflowStatus.Received;
+                    // What about the worst case scenario: one Editor changes it in MMIA and another Editor changes it in TNO?
```
Fosol, line 323: This can occur through our automated processes.  Content that is ingested is published separately.  Which results in the `UpdatedOn` being changed.

kylgarmor: this comparison is between the originalLastUpdateDate from TNO (which we will store with the item upon ingest) and the item currently being ingest from TNO (which we have picked up as an existing reference)

services/net/content/ContentManager.cs
```diff
@@ -223,10 +224,33 @@ private async Task HandleMessageAsync(ConsumeResult<string, SourceContent> resul
     {
         this.Logger.LogDebug("Importing Content from Topic: {topic}, Uid: {key}", result.Topic, result.Message.Key);
         var model = result.Message.Value;
+        bool updateSourceContent = false;
+        long? existingContentId = null;
+
+        var reference = await this.Api.FindContentReferenceAsync(model.Source, model.Uid);
+
+        if (reference != null && reference.Metadata.ContainsKey(ContentReferenceMetaDataKeys.MetadataKeyIngestSource)) {
```
Fosol, line 232: Not suggesting you change this, but it's hard to understand what's going on.

Fosol (approved):

PR #1003, opened 2023-08-25
TNO-454 Fix Morning Report Template

Fosol:
Updated the Custom Report Template so that it would work with the Morning Report.  Updated the report admin page to provide clearer and additional configuration options (including a Table of Contents).  Enhanced Filter admin to generate Elasticsearch query.

NOTE - The Morning Report still has a lot more work regrettably.  There are some challenges related to how the existing TNO solution generates these reports that doesn't align easily with the new MMIA architecture.  I need to review these details to come up with a solution.

## Summary

- Add DB Migration 1.0.62
- Publish tno-core:0.0.300
- Update Reporting Service
- Enhanced Filter Admin
- Enhanced Report Admin
- Update Custom Report Template
- Added new report section "Table of Contents"

## Report Admin

image

## Filter Admin

image


api/net/Areas/Admin/Controllers/ReportController.cs
```diff
@@ -133,7 +133,7 @@ public IActionResult FindInstancesForReportId(int reportId, int? ownerId)
     [SwaggerOperation(Tags = new[] { "Report" })]
     public IActionResult Add(ReportModel model)
     {
-        var result = _reportService.AddAndSave(model.ToEntity(_serializerOptions, false));
+        var result = _reportService.AddAndSave(model.ToEntity(_serializerOptions));
```
Fosol, line 136: Defaults to `false` now.

app/editor/src/features/admin/filters/constants/searchInOptions.ts
```diff
@@ -0,0 +1,22 @@
+export const searchInOptions = (callback: (value: string | number | undefined) => void) => [
+  {
+    id: 'all',
+    label: 'All',
+    onClick: () => callback('all'),
```
Fosol, line 5: I'm not a fan of the implementation of the `ToggleGroup` requiring event hooks to each button instead of the group.

app/editor/src/features/admin/filters/utils/generateQuery.ts
```diff
@@ -0,0 +1,151 @@
+import moment from 'moment';
+import { IFilterSettingsModel } from 'tno-core';
+
+/**
+ * Generates an Elasticsearch query based on specified 'query'.
+ * @param settings Form values that will be used to configure the elasticsearch query.
+ * @param query Original query object.
+ * @returns Elasticsearch query JSON.
+ */
+export const generateQuery = (settings: IFilterSettingsModel, query: any = {}) => {
```
Fosol, line 10: @jdtoombs This is an example of generate an Elasticsearch query from an advanced search form.  We should place this type of code in our shared library so that we can maintain it in a single place.

app/editor/src/features/admin/reports/ReportFormSections.tsx
```diff
@@ -44,7 +45,7 @@ export const ReportFormSections = () => {
       [...values.sections, defaultReportSection(values.id)].map((section, index) => {
         return {
           ...section,
-          name: `section-${index + 1}`,
+          name: `section-${index + 1}-${now.getTime()}`,
```
Fosol, line 48: I don't like this solution however it appears to be necessary due to how Postgres and EF update the sections which require a unique name.  The issue occurs when sections are moved around which would result in them receiving a new name based on their position.

app/subscriber/src/components/toggle-group/ToggleGroup.tsx
Fosol, line 1: Need to remove duplicate components that already exist in tno-core.

libs/net/dal/Migrations/1.0.62/Up/PostUp/00-ReportTemplate.sql
```diff
@@ -0,0 +1,177 @@
+DO $$
+BEGIN
+
+UPDATE public."report_template"
+SET
+  "subject" = '@using TNO.TemplateEngine
+@Settings.Subject.Text @(Settings.Subject.ShowTodaysDate ? $" - {ReportExtensions.GetTodaysDate():dd-MMM-yyyy}" : "")'
+  , "body" = '@inherits RazorEngineCore.RazorEngineTemplateBase<TNO.TemplateEngine.Models.Reports.ReportTemplateModel>
```
Fosol, line 8: This template currently handles all the configured options.

libs/net/entities/ReportSectionType.cs
```diff
@@ -0,0 +1,22 @@
+namespace TNO.Entities;
+
+/// <summary>
+/// Provides report section types.
+/// </summary>
+public enum ReportSectionType
```
Fosol, line 6: Each section has a type which controls the way we present the data.

libs/net/models/Areas/Editor/Models/Report/ReportModel.cs
```diff
@@ -80,7 +80,7 @@ public Entities.Report ToEntity(JsonSerializerOptions options)
         entity.Settings = JsonDocument.Parse(JsonSerializer.Serialize(this.Settings, options));
         entity.Sections.ForEach(s =>
         {
-            var section = this.Sections.FirstOrDefault(us => us.Id == s.Id || us.Name == s.Name) ?? throw new InvalidOperationException("Unable to find matching section");
+            var section = this.Sections.FirstOrDefault(us => us.Name == s.Name) ?? throw new InvalidOperationException("Unable to find matching section");
```
Fosol, line 83: Fixed bug where saving multiple new sections would use the first  sections settings for all of the below.

libs/npm/core/src/components/form/radio/RadioGroup.tsx
```diff
@@ -138,7 +138,7 @@ export const RadioGroup = <OT extends string | number | IOptionItem | HTMLOption
               } else if (typeof option === 'number') {
                 const value = option as number;
                 return (
-                  <span key={value}>
+                  <styled.Span spaceUnderRadio={!!spaceUnderRadio} key={value}>
```
Fosol, line 141: Fixed styling inconsistencies

PR #1008, opened 2023-08-26
TNO-1699 Add Sentiment to Filter

Fosol:
The Filter Admin now supports Sentiment.  This took a lot longer than planned.  I ran into issues when attempting to update a number of the npm packages.

> @jdtoombs Take a look at the way I am generating the Elasticsearch query.  We should reuse this code in the Subscriber app.

## Summary

- Published `tno-core:0.0.304`
- Added sentiment to the filter admin which generates an Elasticsearch query
- Created common SentimentSlider component and placed it in the tno-core library
- Replaced Subscriber Advanced sentiment slider with new component

## Filter Admin

image

## Subscriber Advanced Search

image


app/editor/package.json
```diff
@@ -42,7 +42,7 @@
     "react-drag-drop-files": "2.3.9",
     "react-draggable": "4.4.5",
     "react-error-boundary": "3.1.4",
-    "react-icons": "4.8.0",
+    "react-icons": "4.10.1",
```
Fosol, line 45: Updated to get the latest icons, specifically the X

app/subscriber/src/components/sidebar/advanced-search/components/SentimentSection.tsx
```diff
@@ -21,35 +17,20 @@ export const SentimentSection: React.FC<ISentimentSectionProps> = ({
...
-      <Row className="expanded sentiment-range">
-        <Slider
-          marks={marks}
-          range
-          min={-5}
-          max={5}
-          value={advancedSearch.sentiment}
-          onChange={(e) =>
+      <Col justifyItems="stretch">
+        <SentimentSlider
```
Fosol, line 23: We now have a shared component

libs/npm/core/src/components/form/timeinput/TimeInput.tsx
```diff
@@ -5,7 +5,8 @@ import { Error } from '../../form';
 import { Show } from '../../show';
 import * as styled from './styled';
 
-export interface ITimeInputProps extends InputHTMLAttributes<HTMLInputElement> {
+export interface ITimeInputProps
```
Fosol, line 8: There are typescript issues with this when upgrading packages.  I've left it in since we don't use these properties anyway.  The MaskInput is from a dead project too...

Fosol:
I'll merge later tonight so that I can continue with the next two filter updates.

jdtoombs (approved):
Looks good and will do!

PR #1010, opened 2023-08-28
TNO-1700 Add actions to filters

Fosol:
We can now dynamically add actions to a report filter through the admin form.  Created separate utility functions to generate the Elasticsearch query.

## Filter Admin Form

image


app/editor/src/features/admin/filters/FilterFormQuery.tsx
```diff
@@ -157,73 +163,112 @@ export const FilterFormQuery: React.FC = () => {
...
-          }}
-        />
-        <SentimentSlider
-          label="Sentiment"
-          value={values.settings.sentiment ?? []}
-          onChange={(value) => {
-            updateQuery('sentiment', value);
-          }}
-        />
+        <Row>
```
Fosol, line 166: Tightened up the form so that each row has two input fields.

app/editor/src/features/admin/filters/utils/generateMultiMatch.ts
```diff
@@ -5,22 +5,19 @@
  * @returns An Elasticsearch query for either a multi_match or a match.
  */
 export const generateMultiMatch = (fields: string | string[], values: any) => {
+  if (values === undefined || values === null) return undefined;
```
Fosol, line 8: All the functions return `undefined` when there isn't anything to return.  This makes more sense and is easier to implement.

app/editor/src/features/admin/filters/utils/generateQuery.ts
```diff
@@ -23,14 +24,15 @@ export const generateQuery = (settings: IFilterSettingsModel, query: any = {}) =
...
-        ],
+          generatePublishedOnQuery(settings),
+          generateTerms('sourceId', settings.sourceIds),
+          generateTerms('productId', settings.productIds),
+          generateTerms('seriesId', settings.seriesIds),
+          generateTerms('contributorId', settings.contributorIds),
+          generateRangeForArrayField('tonePools.value', settings.sentiment),
+          ...generateQueryForActions(settings.actions ?? []),
+          generateTextQuery(settings),
+        ].filter((v) => v !== undefined),
```
Fosol, line 35: Remove any of the `undefined` filters as they are not needed.

app/editor/src/features/admin/filters/utils/generateQueryForAction.ts
```diff
@@ -0,0 +1,38 @@
+export const generateQueryForAction = (
```
Fosol, line 1: I tried to make this as generic and reusable as possible, but it's still essentially a two path solution.  It supports a single wildcard search, or a specific value search.

PR #1011, opened 2023-08-28
TNO-1701 Add content type filter

Fosol:
We can now dynamically add tags, edition, section, and page values to the Elasticsearch filter.

## Summary

- Add tags filter
- Add edition filter
- Add section filter
- Add page filter

## Filter Admin

image

### DIFF
--- a/app/editor/src/features/admin/filters/FilterFormQuery.tsx
+++ b/app/editor/src/features/admin/filters/FilterFormQuery.tsx
@@ -22,8 +22,8 @@ import {
   ToggleGroup,
 } from 'tno-core';
 
-import { searchInOptions } from './constants/searchInOptions';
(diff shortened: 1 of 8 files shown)


app/editor/src/features/admin/filters/FilterFormQuery.tsx
```diff
@@ -163,6 +168,41 @@ export const FilterFormQuery: React.FC = () => {
...
           </Row>
         </Row>
+        <Row>
+          <Col flex="1">
+            <FormikText
+              name="edition"
+              label="Edition"
+              value={values.settings.edition ?? ''}
+              onChange={(e) => {
+                const value = e.target.value.length ? e.target.value : undefined;
```
Fosol, line 178: Empty string values are set as undefined so that they are not included in the query.

PR #1012, opened 2023-08-28
fix: #tno-1711 - renamed service in yaml

kylgarmor:
- now matches pattern


kylgarmor:
not suire if this is entirely neccesary, but it bothered me...

Fosol (approved):

PR #1015, opened 2023-08-28
fix: #tno-1712 - added lookup for contenttype

kylgarmor:
created a quick lookup for contenttype based on file extension.  Based on the source data this should fix the error being seen in the logs

### DIFF
--- a/services/net/contentmigration/ContentMigrationAction.cs
+++ b/services/net/contentmigration/ContentMigrationAction.cs
@@ -318,9 +318,18 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
                         if (string.IsNullOrEmpty(newsItem.ContentType)) {
                             string ext = Path.GetExtension(newsItem.FilePath);
                             switch (ext.ToUpper()){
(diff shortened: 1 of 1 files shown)


services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -315,10 +315,35 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
...
+                                case ".PDF":
+                                    newsItem.ContentType = "application/pdf";
+                                    break;
+                                case ".TXT":
+                                    newsItem.ContentType = "text/plain";
+                                    break;
+                            }
+                        }
+
+                        if (string.IsNullOrEmpty(newsItem.ContentType)) {
```
Fosol, line 339: What other file types show up in the the DB?

Fosol (approved):

services/net/contentmigration/ContentMigrationAction.cs
kylgarmor: found two more after expanding my search criteria.

PR #1017, opened 2023-08-29
TNO-1698 Exclude content from report

Fosol:
Reports now have a way to exclude content found in the prior report instance.  Specifically, it now ensures Elasticsearch will return other content if there is any.

## Summary

- Requests to Elasticsearch will now include a `not_match` to exclude content
- Report admin can now send a test email without generating a report instance
- Report admin can now delete and resent a prior instance
- Updated Reporting Service
- Fixed styling on Modal component
- Publish `tno-core:0.0.306`

## Report Admin

image


libs/npm/core/src/hooks/modal/useModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+   */
+  const show = (props: IModalProps) => {
+    setModalProps({ ...props, isShowing: true });
+  };
+
+  return {
+    isShowing: modalProps.isShowing,
+    toggle,
+    show,
+    Modal: <Modal {...modalProps} hide={toggle} />,
```
Fosol, line 33: Not sure I love this, but I was trying to come up with a way to support multiple modals from one hook statement.  This still doesn't do exactly what I wanted, but it does provide a way to dynamically change the values of the modal.

PR #1024, opened 2023-08-30
TNO-1698 Fix report generation

Fosol:
Fixing issue where sending a test email was generating a report instance.

## Summary

- Publish tno-core:0.0.307
- Updated Col and Row components to support `ref`
- Fixed API endpoint for testing reports


api/net/Areas/Admin/Controllers/ReportController.cs
```diff
@@ -193,7 +193,8 @@ public async Task<IActionResult> SendToAsync(int id, string to)
         {
             RequestorId = user.Id,
             To = to,
-            UpdateCache = true
+            UpdateCache = true,
+            GenerateInstance = false
```
Fosol, line 197: Now it's in the correct endpoint.

app/editor/src/store/hooks/admin/useReports.ts
```diff
@@ -49,13 +49,16 @@ export const useReports = (): [IAdminState & { initialized: boolean }, IReportCo
...
       getReport: async (id: number) => {
         const response = await dispatch<IReportModel>('get-report', () => api.getReport(id));
-        store.storeReports((reports) =>
-          reports.map((ds) => {
-            if (ds.id === response.data.id) return response.data;
-            return ds;
-          }),
-        );
-        return response.data;
+        if (response.status === 200) {
```
Fosol, line 52: We should be doing this more often.  When there is a failure our response has invalid data.

PR #1025, opened 2023-08-30
Fixing Elasticsearch sort bug

Fosol:
A regression issue was introduced when sorting was added to Elasticsearch queries.  This PR resolves that issue.

### DIFF
--- a/libs/net/elastic/Models/HitModel.cs
+++ b/libs/net/elastic/Models/HitModel.cs
@@ -16,7 +16,7 @@ public class HitModel<T>
     public string Id { get; set; } = "";
 
     [JsonPropertyName("_score")]
-    public float Score { get; set; }
+    public float? Score { get; set; }
 
     [JsonPropertyName("_source")]
     public T Source { get; set; } = default!;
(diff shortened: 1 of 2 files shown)


libs/net/elastic/Models/HitModel.cs
```diff
@@ -16,7 +16,7 @@ public class HitModel<T>
     public string Id { get; set; } = "";
 
     [JsonPropertyName("_score")]
-    public float Score { get; set; }
+    public float? Score { get; set; }
```
Fosol, line 19: When you sort the score is returned as null.

https://stackoverflow.com/questions/41301691/getting-score-null-in-elastic-search

I don't think this is ideal behaviour.

libs/net/elastic/TNOElasticClient.cs
```diff
@@ -73,9 +73,9 @@ private static ConnectionSettings GetConnectionSettings(ElasticOptions options)
         /// <param name="query"></param>
         /// <returns></returns>
         public async Task<SearchResultModel<T>> SearchAsync<T>(
-            string index, JsonDocument query, string sortBy) where T : class
+            string index, JsonDocument query, string? sortBy = null) where T : class
```
Fosol, line 76: Make this optional and handle it correctly.

libs/net/elastic/Models/HitModel.cs
Fosol: @CelloLiu testing this would have discovered the issue immediately.

PR #1026, opened 2023-08-31
TNO-1705: Advanced search bugs

jdtoombs:
This ticket ended up expanding a bit more as I have been running into issues, I have it down to 1 now.

My current issue as you will see in the GIF below is that when searching for content with files attached I get an "all shards failed" message. 

Interestingly enough this only started to happen once I had content with a file attached, before it was correctly returning no items without an error (previously when no content with a file existed).

adv-filt

This PR includes the functionality for filtering by

- top story
- front page
- if a file includes an image
- disable/enable bolding

### DIFF
--- a/app/subscriber/src/components/sidebar/advanced-search/AdvancedSearch.tsx
+++ b/app/subscriber/src/components/sidebar/advanced-search/AdvancedSearch.tsx
@@ -67,6 +67,7 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ expanded, setEx
   const advancedFilter = React.useMemo(
     () =>
       makeFilter({
+        actions: advancedSearch?.topStory ? ['Top Story'] : [],
         headline:
           advancedSearch?.searchTerm && advancedSearch.searchInField?.headline
             ? advancedSearch.searchTerm
@@ -87,8 +88,9 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ expanded, setEx
           Object.values(advancedSearch.searchInField).every((v) => v === false)
             ? advancedSearch.searchTerm
             : '',
+        productIds: advancedSearch?.frontPage ? [11] : [],
         startDate: advancedSearch?.startDate,
-        sourceIds: advancedSearch?.frontPage ? [11] : advancedSearch?.sourceIds,
+        sourceIds: advancedSearch?.sourceIds,
         sentiment: advancedSearch?.sentiment,
         endDate: advancedSearch?.endDate,
         topStory: advancedSearch?.topStory,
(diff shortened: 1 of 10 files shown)


app/subscriber/src/components/sidebar/advanced-search/AdvancedSearch.tsx
```diff
@@ -87,8 +88,9 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ expanded, setEx
           Object.values(advancedSearch.searchInField).every((v) => v === false)
             ? advancedSearch.searchTerm
             : '',
+        productIds: advancedSearch?.frontPage ? [11] : [],
```
Fosol, line 91: Hardcoding primary keys will almost always lead to problems.  Extract them to an environmental config.

app/subscriber/src/components/sidebar/advanced-search/components/MoreOptions.tsx
```diff
@@ -25,26 +25,26 @@ export const MoreOptions: React.FC<IMoreOptionsProps> = ({
           your search using these settings.
         </p>
         <Col>
-          {/* TODO: What is the below? */}
-          <Checkbox label="featured on the MMIA home page" />
```
Fosol, line 29: This is published content.  This checkbox controls which index we search.

app/subscriber/src/components/sidebar/advanced-search/AdvancedSearch.tsx
```diff
@@ -87,8 +88,9 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ expanded, setEx
           Object.values(advancedSearch.searchInField).every((v) => v === false)
             ? advancedSearch.searchTerm
             : '',
+        productIds: advancedSearch?.frontPage ? [Number(process.env.FRONT_PAGE_ID)] : [],
```
jdtoombs, line 91: @Fosol - is this what you meant?

Fosol: You specific implementation won't work. Env values are not available to clients, only to the server.  For now a simple solution is to create a config JSON file somewhere with constants that we can include in the build of the image.  However, this has issues when the image is used with a different dataset.

Fosol: You can add a JSON file as constants, and then that file can be overridden by our build process if required.

app/subscriber/src/components/sidebar/advanced-search/constants/constants.json
Fosol, line 1: Move this to the public folder.  Then the pattern for replacing JSON files will be the same as the one we use for keycloak.

Fosol (changes requested):

Fosol (approved):

PR #1030, opened 2023-09-06
fix: #tno-1734 - tweaked bash script portion

kylgarmor:
seemingly the user context was being lost inside the call to rm which is odd


Fosol (approved):
Awesome!

PR #1033, opened 2023-09-06
TNO-1615 Add master pipeline

Fosol:
Added a pipeline and trigger for the `master` branch to build and deploy to the TEST environment.

Still need to do more testing on this to ensure I haven't broken the `dev` branch pipeline to the DEV environment.

### DIFF
--- a/openshift/kustomize/tekton/base/pipelines/buildah-all.yaml
+++ b/openshift/kustomize/tekton/base/pipelines/buildah-all.yaml
@@ -14,11 +14,15 @@ metadata:
     managed-by: kustomize
 spec:
   params:
-    - name: GIT_REF
+    - name: GIT_SOURCE
       description: >-
         The git revision reference to build.
       type: string
       default: dev
+    - name: GIT_REF
+      description: >-
+        The git revision reference to build.
+      type: string
 
     - name: CONTEXT
       description: >-
@@ -86,6 +90,8 @@ spec:
       runAfter:
         - wait
       params:
+        - name: GIT_SOURCE
+          value: $(params.GIT_SOURCE)
         - name: GIT_REF
           value: $(params.GIT_REF)
         - name: CONTEXT
--- a/openshift/kustomize/tekton/base/tasks/git-conditions.yaml
+++ b/openshift/kustomize/tekton/base/tasks/git-conditions.yaml
@@ -19,9 +19,12 @@ metadata:
     managed-by: kustomize
 spec:
   params:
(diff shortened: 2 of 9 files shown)


openshift/kustomize/tekton/base/pipelines/buildah-all.yaml
```diff
@@ -14,11 +14,15 @@ metadata:
     managed-by: kustomize
 spec:
   params:
-    - name: GIT_REF
+    - name: GIT_SOURCE
```
Fosol, line 17: There was a template bug between GIT_SOURCE and GIT_REF.  I have resolved this issue and now both the `master` and `dev` branch should work and trigger the pipeline.

openshift/kustomize/tekton/base/tasks/git-conditions.yaml
```diff
@@ -48,7 +51,10 @@ spec:
 
         # Update the git repo with the latest commit.
         git config --global --add safe.directory '*'
-        git checkout $(params.GIT_REF)
+        git fetch origin master
+        git fetch origin dev
+        git checkout $(params.GIT_SOURCE)
+        # TODO: Make the branch equal to the GIT_REF
```
Fosol, line 57: We're not using the GIT_REF currently in this step, but it would be nice to be able to.  The way it is currently implemented could result in pulling and using the most recent commit rather than the one that fired the webhook trigger.

openshift/kustomize/tekton/base/triggers/dev/ingress.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: git-webhook
+  name: git-webhook-dev
```
kylgarmor, line 5: curious about the use of the word/token `dev` here.  Are we tying the use of this too closely to the `dev` branch or environment?

kylgarmor (approved):
ok, to merge, but just made one comment around the use of the word/token `dev`

openshift/kustomize/tekton/base/triggers/dev/ingress.yaml
Fosol: It's a simple way to know which branch is using this webhook.
Originally I wanted to use a single webhook from Github, but I couldn't figure out how to make it trigger different pipelines.  I've always run into some kind of limitation either in Tekton or in my knowledge.

PR #1036, opened 2023-09-08
TNO-1707: Link to subscriber

jdtoombs:
In this PR:
- allow editors to view published content on the subscriber site
- only shows if content is published
- link changes depending on environment

I am curious what is the best way to store these external subscriber links and if a JSON is the best approach.

externallink

### DIFF
--- a/app/editor/public/links.json
+++ b/app/editor/public/links.json
@@ -1,5 +1,0 @@
-{
-  "devSubscriber": "https://mmia-dev.apps.silver.devops.gov.bc.ca",
-  "testSubscriber": "https://mmia-test.apps.silver.devops.gov.bc.ca",
-  "localSubscriber": "http://localhost:40081"
-}
--- a/app/editor/public/links.json
+++ b/app/editor/public/links.json
@@ -1,0 +1,5 @@
+{
+  "devSubscriber": "https://mmia-dev.apps.silver.devops.gov.bc.ca",
+  "testSubscriber": "https://mmia-test.apps.silver.devops.gov.bc.ca",
+  "localSubscriber": "http://localhost:40081"
+}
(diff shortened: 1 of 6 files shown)


app/editor/public/links.json
```diff
@@ -0,0 +1,5 @@
+{
```
jdtoombs, line 1: Not sure if this is the best approach

Fosol: This won't work because of how images are built and deployed.  I'll dig into how we're getting this value in other places.  I can't recall if we have something similar for the web app.

jdtoombs: Thanks!

Fosol: Ah I forget more than I remember regrettably.  I've a solution for this and your prior PR that used a constant.  You can fix both together.

The API has an Admin page for Settings.  You can configure a Key=Value pair for anything you need.  Then you can access the value from the Redux store in the lookups I believe.

image

Fosol (changes requested):
Change this and the prior PR that is using the `FrontPageId`

Fosol:
It also means you'll need to throw a toast error if the value isn't there so that we know it's not been configured yet

app/editor/public/links.json
jdtoombs: Perfect, thank you!

jdtoombs:
@Fosol, I have updated it for this PR. Need to do some more work for my previous PR as the hooks and such are not set up for the subscriber - may create a separate ticket for that

app/editor/src/features/content/form/components/tool-bar/form/PublishedSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+  /** Form values */
+  values: IContentForm;
+}
+
+export const PublishedSection: React.FC<IPublishedSectionProps> = ({ values }) => {
+  const [subscriberUrl, setSubscriberUrl] = React.useState<string>();
+  const [{ settings }, api] = useSettings();
+
+  React.useEffect(() => {
+    api.findAllSettings().then((data) => {
```
Fosol, line 20: You shouldn't need to make this separate request.  I believe the lookups already include these values.  Please confirm.

Fosol (changes requested):

Fosol (approved):

Fosol (changes requested):

Fosol (approved):

PR #1038, opened 2023-09-09
TNO-504 Configure Kafka partitions

Fosol:
Kafka will now be configured with default replication of 3 and partitions of 6.  However, I will need to recycle all the pods to pick up the new default configuration.  Tonight I will do it since I need to stop everything from running until it's done.


openshift/kustomize/kafka/broker/base/config-map.yaml
```diff
@@ -27,4 +27,6 @@ data:
   KAFKA_JMX_HOSTNAME: kafka-headless
   KAFKA_HEAP_OPTS: "-Xmx8G -Xms6G"
   KAFKA_JVM_PERFORMANCE_OPTS: "-server -XX:MetaspaceSize=96m  -XX:G1HeapRegionSize=16M -XX:MinMetaspaceFreeRatio=50 -XX:MaxMetaspaceFreeRatio=80 -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+ExplicitGCInvokesConcurrent -Djava.awt.headless=true"
+  KAFKA_NUM_PARTITIONS: "6"
+  KAFKA_DEFAULT_REPLICATION_FACTOR: "3"
```
Fosol, line 31: Must keep replication to be equal or less than the number of brokers.  We are running 4 presently.

db/kafka/README.md
```diff
@@ -36,3 +36,39 @@ SSH into the Kafka broker pod.
...
+
+Update the topic replication.
+
+Create a JSON file `replicas.json` and copy to configure. This requires that the topic have the specified number of partitions (i.e. 6).
+
+```json
+{
+  "version": 1,
+  "partitions": [
+    { "topic": "BCNG", "partition": 0, "replicas": [2, 3, 4] },
```
Fosol, line 54: I don't understand the spread of numbers, but I copied from an example.

PR #1049, opened 2023-09-12
TNO-1752 Add Content.PostedOn

Fosol:
We now capture the date and time when an Editor posts content to subscribers.  This provides us a way to create reports that only include newly posted content since the prior report instance was run.  This allows for scenarios where an old story (a story with an older published date) has been added after a report has been run, but it should also be included in the next report.  

Note, that if a filter has a `PublishedOn` range condition this will take priority over this new feature to include any newly posted content.

## Summary

- Add DB migration 1.0.65
- Publish tno-core:0.0.314
- Add elastic migration 1.0.8
- Added new property `Content.PostedOn`
- Updated models for content
- Update Report Service to filter content to only allow for newly posted based on the prior instance run
- Fixed bug related to updating the status of content when indexing, and the related SignalR message to inform the owner of the content

## Report Admin

Users can select the "Include only new content posted since previous run" to include any newly posted content regardless of the published date.

image


api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -206,6 +206,7 @@ public async Task<IActionResult> AddAsync(ContentModel model)
         var user = _userService.FindByUsername(username) ?? throw new NotAuthorizedException("User does not exist");
         var newContent = (Content)model;
         newContent.OwnerId = user.Id;
+        newContent.PostedOn = newContent.Status == ContentStatus.Publish || newContent.Status == ContentStatus.Published ? DateTime.UtcNow : null;
```
Fosol, line 209: Set the `PostedOn` if the content is being published.

libs/net/dal/Migrations/1.0.65/Up/PostUp/00-Content.sql
```diff
@@ -0,0 +1,8 @@
+DO $$
+BEGIN
+
+UPDATE "content"
+SET "posted_on" = c."published_on"
```
Fosol, line 5: Set the default value to whatever the published date is.

services/net/reporting/ReportingManager.cs
```diff
@@ -477,6 +445,38 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
...
+    /// Send out an email for the specified report.
+    /// Generate a report instance for this email.
+    /// Send an email merge to CHES.
+    /// This will send out a separate email to each context provided.
+    /// </summary>
+    /// <param name="request"></param>
+    /// <param name="instance"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentException"></exception>
+    private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Services.Models.AVOverview.AVOverviewInstanceModel instance)
```
Fosol, line 458: Moved it down to organize.

tools/elastic/migration/Migrations/1.0.8/up/post/00-update-unpublished_content.json
```diff
@@ -0,0 +1,13 @@
+{
+  "action": "updateMapping",
+  "settings": {
+    "indexName": "unpublished_content"
+  },
+  "data": {
```
Fosol, line 6: I think ideally we would do a full reindex.  I may still choose to do this tomorrow morning.

api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -271,10 +272,20 @@ public async Task<IActionResult> UpdateAsync(ContentModel model, bool index = fa
     [ProducesResponseType(typeof(ContentModel), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Content" })]
-    public IActionResult UpdateStatusAsync(ContentModel model)
+    public async Task<IActionResult> UpdateStatusAsync(ContentModel model)
```
Fosol, line 275: @kylgarmor I fixed several regression issues introduced in a prior PR.
1) Async function naming
2) SignalR message when status has been updated
3) Update status in database

libs/net/dal/Services/ContentService.cs
```diff
@@ -554,7 +554,7 @@ public Content UpdateStatusOnly(Content updated)
         original.Version = updated.Version;
         this.Context.ResetVersion(original);
 
-        return base.Update(original);
+        return base.UpdateAndSave(original);
```
Fosol, line 557: @kylgarmor We have to methods, one updates the data in the DB, the other does not.  `Update` does not save to the DB.  It provides a way to do multiple actions with one transaction.  Use `UpdateAndSave` to save to DB.  I should have caught this originally :(

libs/net/dal/Services/ReportService.cs
```diff
@@ -338,15 +354,18 @@ public override Report Update(Report entity)
     /// <exception cref="Exception"></exception>
     public async Task<Dictionary<string, Elastic.Models.SearchResultModel<API.Areas.Services.Models.Content.ContentModel>>> FindContentWithElasticsearchAsync(string index, Report report)
     {
-        var results = new Dictionary<string, Elastic.Models.SearchResultModel<API.Areas.Services.Models.Content.ContentModel>>();
+        var searchResults = new Dictionary<string, Elastic.Models.SearchResultModel<API.Areas.Services.Models.Content.ContentModel>>();
```
kylgarmor, line 357: good name change.  Much clearer.

PR #1050, opened 2023-09-13
feat: #tno-1771 - add new migrator

kylgarmor:
- added new Ingest records
- updated logic around only looking at TNO records marked as published


services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -149,13 +190,20 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
     /// <param name="cancellationToken"></param>
     /// <returns></returns>
     /// <exception cref="ConfigurationException"></exception>
+    /// <exception cref="ArgumentNullException"></exception>
     public override async Task PerformActionAsync<T>(IIngestServiceActionManager manager, string? name = null, T? data = null, CancellationToken cancellationToken = default) where T : class
     {
-        string importMigrationType = manager.Ingest.GetConfigurationValue("importMigrationType");
-        if (string.IsNullOrEmpty(importMigrationType))
+        string importMigrationTypeRaw = manager.Ingest.GetConfigurationValue("importMigrationType");
```
Fosol, line 196: I think we have this working for enum types

libs/net/models/Extensions/IngestModelExtensions.cs
```diff
@@ -48,14 +49,17 @@ public static string GetConfigurationValue(this IngestModel ingest, string key,
...
-            return element.ValueKind switch
-            {
-                JsonValueKind.String => (T)Convert.ChangeType($"{element.GetString()}".Trim(), typeof(T)),
-                JsonValueKind.Null or JsonValueKind.Undefined => default,
-                JsonValueKind.Number => (T)Convert.ChangeType($"{element.GetInt32()}", typeof(T)),
-                JsonValueKind.True or JsonValueKind.False => (T)Convert.ChangeType($"{element.GetBoolean()}", typeof(T)),
-                _ => (T)Convert.ChangeType($"{element}", typeof(T)),
-            };
+            if (typeof(T).GetTypeInfo().IsEnum)
+                return (T)Enum.Parse(typeof(T), $"{element.GetString()}".Trim());
```
Fosol, line 53: This will only work if the value being parsed is a string.  You should also check the element.ValueKind.

kylgarmor: added check for element.ValueKind

Fosol (approved):

PR #1054, opened 2023-09-14
feat: #tno-1773 - modified PerformActionAsync

kylgarmor:
- now returns an enum to signify a more specific result


libs/net/services/Actions/ServiceAction.cs
```diff
@@ -4,6 +4,16 @@
 
 namespace TNO.Services.Actions;
 
+/// <summary>
+/// ServiceActionResult enum, encapsulates how a service action completed.
+/// </summary>
+public enum ServiceActionResult
```
Fosol, line 10: Move this to a separate file.  Generally only have multiple objects in a single file when it's the only file that access that object.

Fosol (changes requested):
I didn't see the other ingest services returning a skip.  Is there a reason?

Fosol (approved):

PR #1063, opened 2023-09-15
Implement Route-based code splitting

jdtoombs:
@Fosol - you can let me know what you think.

This is based off of https://web.dev/code-splitting-suspense/

Locally this appears to only take off milliseconds according to my lighthouse report but maybe it will have more of an affect in DEV/TEST. The annoying things was that React's `lazy` only supports default exports so I had to go through and change them. The other drawback is the imports are kind of ugly on the routers but maybe that is not an issue at the moment

This also reduces the amount of unused JS appearing in these reports, but not by as much as I would like.

### DIFF
--- a/app/editor/src/components/layout/DefaultLayout.tsx
+++ b/app/editor/src/components/layout/DefaultLayout.tsx
@@ -33,7 +33,7 @@ interface ILayoutProps extends React.HTMLAttributes<HTMLDivElement> {
  * @param param0 Component properties.
  * @returns DefaultLayout component.
  */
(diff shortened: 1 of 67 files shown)


Fosol (approved):
Makes me nervous.  Everything I read makes the statement that using default exports is a bad practice.

jdtoombs:
> Makes me nervous. Everything I read makes the statement that using default exports is a bad practice.

Me too, just throwing everything at the wall to see if it sticks at the moment 😕

PR #1064, opened 2023-09-15
Reduce memory usage - redux

Fosol:
I've reduce the amount of data we're storing in redux memory on our Content List View and Papers View pages.

image

### DIFF
--- a/app/editor/src/features/content/papers/Papers.tsx
+++ b/app/editor/src/features/content/papers/Papers.tsx
@@ -2,13 +2,12 @@ import React, { lazy } from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
 import { toast } from 'react-toastify';
 import { HubMethodName, useApiHub, useChannel, useContent } from 'store/hooks';
-import { useContentStore } from 'store/slices';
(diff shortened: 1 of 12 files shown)


app/editor/src/store/slices/content/interfaces/IContentSearchResult.ts
```diff
@@ -0,0 +1,26 @@
+import { ContentStatusName, ContentTypeName } from 'tno-core';
+
+export interface IContentSearchResult {
```
Fosol, line 3: We now will only store a minimum of data.

app/editor/src/features/content/papers/Papers.tsx
```diff
@@ -90,17 +89,21 @@ const Papers: React.FC<IPapersProps> = (props) => {
   }, [onContentUpdated, hub]);
 
   const handleClickUse = React.useCallback(
-    (content: IContentModel) => {
-      updateStatus(content)
+    (content: IContentSearchResult) => {
+      getContent(content.id)
```
Fosol, line 93: I don't like this solution.  But it temporarily resolves the issue for updating the status of the selected content.

PR #1066, opened 2023-09-18
Remove tooltip from list view

jdtoombs:
Performance is current abysmal with react-tooltip, looking at reimplementing and making sure we are up to par with their most recent documentation. In the meantime since it is taking a little bit, I have removed the tooltips from the list view (maybe can reintroduce later if need be and performance is fixed). 

This bit of work will be two parts.


Fosol (changes requested):
Remove it from the papers page too

Fosol (approved):

PR #1077, opened 2023-09-21
TNO-1724 Update My Reports

Fosol:
Using an icon to indicate if the report contains media analytics (charts).  It's a basic check if the report has a section where the settings has `showCharts=true`.  This can be a false positive, but for now I think it'll be okay.

image

### DIFF
--- a/app/subscriber/src/features/my-reports/admin/ReportAdminSettings.tsx
+++ b/app/subscriber/src/features/my-reports/admin/ReportAdminSettings.tsx
@@ -44,7 +44,12 @@ export const ReportAdminSettings: React.FC = () => {
   return (
     <Col gap="1rem">
       <Box title="Identify your report" icon={<FaFileInvoice />}>
-        <FormikText name="name" label="Report Name:" required />
+        <FormikText
+          name="name"
+          label="Report Name:"
+          required
+          placeholder="Enter unique report name"
+        />
         <FormikTextArea name="description" label="Description:" />
       </Box>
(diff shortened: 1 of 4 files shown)


app/subscriber/src/features/my-reports/admin/ReportAdminSettings.tsx
```diff
@@ -44,7 +44,12 @@ export const ReportAdminSettings: React.FC = () => {
   return (
     <Col gap="1rem">
       <Box title="Identify your report" icon={<FaFileInvoice />}>
-        <FormikText name="name" label="Report Name:" required />
+        <FormikText
+          name="name"
+          label="Report Name:"
+          required
+          placeholder="Enter unique report name"
```
Fosol, line 51: The error that occurs when a report has the same name isn't great.  But this may reduce that type of issue.

app/subscriber/src/features/my-reports/hooks/useColumns.tsx
```diff
@@ -30,11 +30,19 @@ export const useColumns = (
       name: 'name',
       width: 2,
       cell: (cell) => (
-        <CellEllipsis>
-          <Link to={`/reports/${cell.original.id}/edit`} title="Edit">
-            {cell.original.name}
-          </Link>
-        </CellEllipsis>
+        <Row gap="0.5rem" alignItems="center">
+          <Show visible={cell.original.sections.some((section) => section.settings.showCharts)}>
```
Fosol, line 34: Wish we could create a component that handles or better.

PR #1086, opened 2023-09-26
TNO-1787 Fix memory issue

Fosol:
When using multiple tabs memory consumption would result in crashing the browser.  This update removes communication between the tabs within the browser.  Removing this does regrettably result in loss of some functionality related to the active row and the current search results.  I will look into solving these smaller syncing issues once I can confirm the memory issue is addressed.

## Summary

- Fix memory issue
- Publish tno-core:0.0.323
- Update Editor app
- Each click on Content List View will open a new tab each time
- The Content Form Prev/Next buttons are linked to the search results that spawned the new tab.  It doesn't handle a change in the search results.
- The Content List View does not know which tab/item is active.


api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -224,15 +224,7 @@ public async Task<IActionResult> UpdateAsync(ContentModel model, bool index = fa
...
-        // Send notification to user who requested the content.
-        if (content.OwnerId.HasValue)
-        {
-            var owner = content.Owner ?? _userService.FindById(content.OwnerId.Value);
-            if (!String.IsNullOrWhiteSpace(owner?.Username))
-                await _kafkaMessenger.SendMessageAsync(_kafkaHubOptions.HubTopic, new KafkaHubMessage(HubEvent.SendUser, owner.Username, new InvocationMessage("Content", new[] { new ContentMessageModel(content) })));
-            else
-                await _kafkaMessenger.SendMessageAsync(_kafkaHubOptions.HubTopic, new KafkaHubMessage(HubEvent.SendAll, new InvocationMessage("Content", new[] { new ContentMessageModel(content) })));
-        }
+        await _kafkaMessenger.SendMessageAsync(_kafkaHubOptions.HubTopic, new KafkaHubMessage(HubEvent.SendAll, new KafkaInvocationMessage(MessageTarget.ContentAdded, new[] { new ContentMessageModel(content) })));
```
Fosol, line 227: Always send updates to everyone

app/editor/src/components/layout/DefaultLayout.tsx
```diff
@@ -106,24 +104,6 @@ const DefaultLayout: React.FC<ILayoutProps> = ({ name, children, ...rest }) => {
     return hub.listen(HubMethodName.WorkOrder, onWorkOrderMessage);
   }, [onWorkOrderMessage, hub]);
 
-  const onContentMessage = React.useCallback((content: IContentMessageModel) => {
```
Fosol, line 109: We don't need this toast showing up all the time.

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -90,22 +91,6 @@ const ContentForm: React.FC<IContentFormProps> = ({
   const { combined, formType } = useCombinedView(initContentType);
   const { setShowValidationToast } = useTabValidationToasts();
 
-  const channel = useChannel<any>({
```
Fosol, line 93: I believe this was causing the memory issues.

app/editor/src/features/content/hooks/useTab.tsx
```diff
@@ -1,23 +1,24 @@
...
+}
 
+// TODO: When `useOnTab=true` they get out of sync easily.
+export const useTab = (options: ITabOptions = { useOneTab: false }) => {
   const [tab, setTab] = React.useState<Window | null>(_window);
 
   return React.useCallback(
     (id: number, path: string = '/contents') => {
-      if (!tab || tab.closed) {
+      if (!tab || tab.closed || !options.useOneTab) {
```
Fosol, line 15: Defaults to a new tab for each click.  I'll need to figure out how we can do this another way later.

app/editor/src/store/slices/content/contentSlice.ts
```diff
@@ -72,6 +72,7 @@ export const contentSlice = createSlice({
             items: action.payload.items.map((content) => castContentToSearchResult(content)),
           }
         : action.payload;
+      saveToLocalStorage('content', state.content?.items ?? []);
```
Fosol, line 75: We now store the search results in local storage so that each tab has access to them.

libs/npm/core/scripts/publish.sh
```diff
@@ -29,4 +29,6 @@ elif [ "$version" == "p" ]; then
 fi
 npm version $version
 
-yarn npm publish
\ No newline at end of file
+yarn npm publish
+
+echo "version: $version"
```
Fosol, line 34: I'm always wondering what version was published.

PR #1087, opened 2023-09-26
fix: #tno-1550 - moved options for easier control

kylgarmor:
- re-used the ReportingOptions Config model in Notiofications
- TODO - configure shared ConfigMap for Reporting and Notification svc

### DIFF
--- a/services/net/notification/Config/NotificationOptions.cs
+++ b/services/net/notification/Config/NotificationOptions.cs
@@ -19,24 +19,5 @@ public class NotificationOptions : ServiceOptions
     /// </summary>
     public int AlertId { get; set; }
 
-    /// <summary>
-    /// get/set - The MMIA URL.
-    /// </summary>
-    public Uri? MmiaUrl { get; set; }
-
-    /// <summary>
-    /// get/set - The view content URL.
-    /// </summary>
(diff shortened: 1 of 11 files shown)


services/net/notification/Config/NotificationOptions.cs
```diff
@@ -19,24 +19,5 @@ public class NotificationOptions : ServiceOptions
     /// </summary>
     public int AlertId { get; set; }
```
Fosol, line 21: You could add a Reporting section in this config if you didn't want to change the dependency injection configuration.

Fosol (approved):

services/net/notification/Config/NotificationOptions.cs
kylgarmor: I did think of that, but my thought was that then the ENV style path to the variables would be different to the way it's set up in the Reporting Service. 

For example in the Reporting Service the path would be `Reporting__AddToReportUrl` but if we nested it in the `NotificationOptions` It would become `Service__ReportingOptions__AddToReportUrl`. If we can keep them the same, the possibly the DeploymentConfigs will be easier to keep in line.

PR #1089, opened 2023-09-26
TNO-1787 Fix multi-tab

Fosol:
We now have a little control over how multiple tabs work.  I've added a tri-state button icon in the header that controls how opening content will behave. 

1. Open in a new tab is the default behaviour.  This can lead to a lot of new tabs
2. Open in same tab.  This is essentially a two page view.  Links will override whatever is in the other tab.
3. Open in same window.  This is good for smaller screens that don't want to use tabs.

## Summary

- Publish tno-core:0.0.324
- Updated Editor app
- Provided way to open content in a new tab, one tab, or on same page.
- Ctrl+Click will open content in a new tab
- User's choice is stored in localstorage
- Enhanced FlexboxTable header to be able to use `React.ReactNode`
- Content Form can hide the navigation bar

## Content List View

image

## Papers

image


app/editor/src/components/layout/DefaultLayout.tsx
```diff
@@ -115,7 +122,9 @@ const DefaultLayout: React.FC<ILayoutProps> = ({ name, children, ...rest }) => {
         </Header>
         <div className="main-window">
           <LayoutErrorBoundary>
-            <NavBar />
+            <Show visible={showNav}>
```
Fosol, line 125: A simple way to hide the navigation bar on the Content Form tab

app/editor/src/components/tab-control/TabControl.tsx
```diff
@@ -0,0 +1,35 @@
+import { FaExternalLinkAlt, FaExternalLinkSquareAlt, FaRegClone } from 'react-icons/fa';
+import { useApp } from 'store/hooks';
+import { Show } from 'tno-core';
+
+import { NavigateOptions } from './constants';
+
+export const TabControl: React.FC = () => {
```
Fosol, line 7: Simple component to control the tab behaviour.

app/editor/src/store/hooks/app/useApp.ts
```diff
@@ -70,6 +74,12 @@ export const useApp = (): [IAppState, IAppController] => {
 
   const hasClaim = keycloak.hasClaim();
 
+  React.useEffect(() => {
```
Fosol, line 77: Initial load it will fetch from localstorage.

app/editor/src/components/tab-control/TabControl.tsx
kylgarmor: I dont quite feel like that the header of the row is the right place for this component, but I can see how having it there puts it front and center to the user.  Bobbi might have some say on it when she sees it in play.

kylgarmor (approved):

app/editor/src/components/tab-control/TabControl.tsx
Fosol: I 100% agree.  Just couldn't think of a better location without the whole toolbar resulting in a wrap...

PR #1091, opened 2023-09-26
TNO-1728, 1747: Press gallery list view

jdtoombs:
Currently debugging an error I am having locally, it may be not an issue as I am getting the same one in the admin filter sections when trying to preview the filter. All this works in DEV but not local for me at the moment.

Also still need to publish package after all is working.

In this PR:
- subscribers can add content to folders from list view
- grabs the contributors labelled as press and extracts their aliases 
- searches based on aliases or names


app/subscriber/package.json
```diff
@@ -55,7 +55,7 @@
     "react-tooltip": "5.10.0",
     "redux-logger": "3.0.6",
     "styled-components": "5.3.9",
-    "tno-core": "0.0.324"
+    "tno-core": "./tno-core-0.0.324.tgz"
```
Fosol, line 58: You'll want to publish a new package version

app/subscriber/src/features/press-gallery/PressGallery.tsx
```diff
@@ -1,17 +1,87 @@
...
-/** TODO: Didn't realize I had two different stories - this page will be done in the next ticket*/
 export const PressGallery: React.FC = () => {
   const navigate = useNavigate();
+  const [{ filterAdvanced }, { findContentWithElasticsearch }] = useContent();
+  const [, api] = useContributors();
+  const [pressContributors, setPressContributors] = React.useState<IContributorModel[]>([]);
+  const [results, setResults] = React.useState<any>([]);
   const [selected, setSelected] = React.useState<IContentModel[]>([]);
 
+  const query = (queryText: string, filterAdvanced: IContentListAdvancedFilter) => {
```
Fosol, line 22: We have shared methods in tno-core that we should use so that we have a single place to maintain.

libs/net/models/Areas/Subscriber/Models/Content/ContributorModel.cs
```diff
@@ -7,12 +7,47 @@ namespace TNO.API.Areas.Subscriber.Models.Content;
 /// </summary>
 public class ContributorModel : BaseTypeModel<int>
 {
-    #region Properties
+   #region Properties
+    /// <summary>
+    /// get/set - The primary key of the contributor.
+    /// </summary>
+    public int Id { get; set; }
```
Fosol, line 14: Fix issues

libs/net/models/Areas/Subscriber/Models/Contributor/ContributorModel.cs
```diff
@@ -0,0 +1,109 @@
+using TNO.API.Models;
+
+namespace TNO.API.Areas.Subscriber.Models.Contributor;
+
+/// <summary>
+/// ContributorModel class, provides a model that represents an contributor.
+/// </summary>
+public class ContributorModel : AuditColumnsModel
```
Fosol, line 8: You won't be editing these, so there is no need for the audit columns

Fosol (changes requested):

libs/npm/core/src/utils/generateSimpleQueryString.ts
```diff
@@ -4,13 +4,17 @@
...
+  values: any,
+  defaultOperator?: string,
+) => {
   if (values === undefined || values === null) return undefined;
   return {
     simple_query_string: {
       query: values,
       fields,
-      default_operator: 'and',
+      default_operator: !!defaultOperator ? defaultOperator : 'and',
```
Fosol, line 17: might be clearer if you just did `defaultOperator ?? 'and'`

jdtoombs: much clearer, that was a late night line of code :D

Fosol (approved):

PR #1097, opened 2023-09-27
fix: #tno-1830 - update controller and editor

kylgarmor:
- changed controller to use a boolean instead of index name


Fosol (approved):
You should update the Subscriber app `tno-core` package version too.

jdtoombs (approved):
LGTM!

PR #1101, opened 2023-09-28
TNO-1785 Fix m4a transcription

Fosol:
Added `m4a` format to auto convert to `mp3` before sending to Azure Cognitive Services.

### DIFF
--- a/services/net/transcription/appsettings.json
+++ b/services/net/transcription/appsettings.json
@@ -19,7 +19,7 @@
     "VolumePath": "/data",
     "AzureRegion": "canadacentral",
     "AzureCognitiveServicesKey": "",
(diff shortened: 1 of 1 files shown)


services/net/transcription/appsettings.json
```diff
@@ -19,7 +19,7 @@
     "VolumePath": "/data",
     "AzureRegion": "canadacentral",
     "AzureCognitiveServicesKey": "",
-    "ConvertToAudio": ["mp4", "mov"]
+    "ConvertToAudio": ["mp4", "mov", "m4a"]
```
Fosol, line 22: FFmpeg was able to convert `m4a` to `mp3` and transcribe.

PR #1104, opened 2023-09-28
fix: #tno-1828 - fix missing properties

kylgarmor:
- when updating content, some previously set properties where being lost

### DIFF
--- a/services/net/content/ContentManager.cs
+++ b/services/net/content/ContentManager.cs
@@ -227,9 +227,10 @@ public class ContentManager : ServiceManager<ContentOptions>
         bool updateSourceContent = false;
         long? existingContentId = null;
 
-        // Only add if doesn't already exist.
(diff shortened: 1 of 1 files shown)


services/net/content/ContentManager.cs
```diff
@@ -374,8 +375,18 @@ private async Task HandleMessageAsync(ConsumeResult<string, SourceContent> resul
             }
 
             if (updateSourceContent && (existingContentId != null)) {
-                // before updating, reinstate the Id value
-                content.Id = existingContentId.Value;
+                // before saving, reinstate some values from the original content object
+                if (originalContent != null) {
```
Fosol, line 379: Are these the only properties that need to be reset?

Fosol (approved):

services/net/content/ContentManager.cs
kylgarmor: all of the other properties are being newed up from the content that came in from Kafka

PR #1105, opened 2023-09-29
TNO-1823 Enhance Ingest services

Fosol:
Updated base services library which enables all services to send an email when they have failures.  Currently only the File Monitory Service, Syndication Service and the Content Migration Service have been configured to send an email when an Ingest configuration has reached its max failure limit.

## Summary

- Add email ability to all services on failure
- Updated Infrastructure as Code

### DIFF
--- a/libs/net/models/Areas/Subscriber/Models/Content/ContributorModel.cs
+++ b/libs/net/models/Areas/Subscriber/Models/Content/ContributorModel.cs
@@ -7,11 +7,7 @@ namespace TNO.API.Areas.Subscriber.Models.Content;
 /// </summary>
 public class ContributorModel : BaseTypeModel<int>
 {
-   #region Properties
-    /// <summary>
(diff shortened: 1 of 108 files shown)


libs/net/models/Areas/Subscriber/Models/Content/ContributorModel.cs
```diff
@@ -7,11 +7,7 @@ namespace TNO.API.Areas.Subscriber.Models.Content;
 /// </summary>
 public class ContributorModel : BaseTypeModel<int>
 {
-   #region Properties
-    /// <summary>
-    /// get/set - The primary key of the contributor.
-    /// </summary>
-    public int Id { get; set; }
```
Fosol, line 14: @jdtoombs Not sure how this snuck in with your PR.  I remember there was linting errors and I also posted a message on it.  Don't overwrite inherited properties or it causes issues.

I've removed them.

PR #1110, opened 2023-09-30
feat: #tno-1733 - fixed transcript

kylgarmor:
- swapped out string sanitize method for something better


Fosol (approved):
I think this will always have one extra <p></p> :D

PR #1111, opened 2023-10-03
TNO-1823 Fix ingest services

Fosol:
Ingest service error alerts highlighted two issues.

- Attempting to add a second authorization header
- Processing DTD in XML

### DIFF
--- a/libs/net/core/Http/OpenIdConnectRequestClient.cs
+++ b/libs/net/core/Http/OpenIdConnectRequestClient.cs
@@ -1,10 +1,11 @@
+using System.IdentityModel.Tokens.Jwt;
+using System.Net;
+using System.Net.Http.Headers;
+using System.Text.Json;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using TNO.Core.Exceptions;
 using TNO.Core.Http.Configuration;
-using System.IdentityModel.Tokens.Jwt;
-using System.Net.Http.Headers;
-using System.Text.Json;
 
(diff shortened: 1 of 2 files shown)


libs/net/core/Http/OpenIdConnectRequestClient.cs
```diff
@@ -244,7 +245,13 @@ public override async Task<HttpResponseMessage> SendAsync(string url, HttpMethod
 
             var token = await RequestAccessToken();
 
-            if (!String.IsNullOrWhiteSpace(token)) headers.Add("Authorization", token.ToString());
+            if (!String.IsNullOrWhiteSpace(token))
+            {
+                if (!headers.TryGetValues("Authorization", out _))
```
Fosol, line 250: Replace or add depending on whether there already is an authorization header.

services/net/syndication/SyndicationAction.cs
```diff
@@ -398,12 +398,13 @@ private async Task<SyndicationFeed> GetFeedAsync(Uri url, IIngestServiceActionMa
...
     /// <param name="data"></param>
     /// <returns></returns>
-    private SyndicationFeed GetCustomFeed(string data)
+    static SyndicationFeed GetCustomFeed(string data)
     {
         var settings = new XmlReaderSettings()
         {
             IgnoreComments = false,
             IgnoreWhitespace = true,
+            DtdProcessing = DtdProcessing.Parse,
```
Fosol, line 407: We should be able to process these XML documents now.

PR #1112, opened 2023-10-03
TNO-1739 Fix file paths

Fosol:
Currently files are uploaded to a directly that named by the `Source.Code`.  If the code has a space it causes issues.  Code has been updated to replace a space with an underscore.


libs/net/dal/Models/ContentFileReference.cs
```diff
@@ -258,7 +258,7 @@ public ContentFileReference(FileReference fileReference, System.IO.FileInfo file
     public static string GenerateFilePath(Content content, IFormFile file)
     {
         var fileName = GenerateFileName(content, file);
-        return System.IO.Path.Combine(content.OtherSource, fileName);
+        return System.IO.Path.Combine(content.OtherSource.Replace(' ', '_'), fileName);
```
Fosol, line 261: simple solution

PR #1117, opened 2023-10-04
feat: #tno-1730 - added basic filter import/export

kylgarmor:
This feature was not really developed for general use, it's lifespan is more for moving filters from dev -> test -> production.  Once we are live, it could probably be disabled. Which begs the question, do we want to make this a feature than can be enabled/disabled in the app itself?

image


app/editor/src/features/admin/filters/FilterFormImportExport.tsx
```diff
@@ -0,0 +1,114 @@
...
+      var importedFilter = parseExportedFilter(
+        rawExportedFilter,
+        actions,
+        contributors,
+        series,
+        sources,
+        products,
+      );
+
+      setFieldValue('name', importedFilter.name ?? '');
```
Fosol, line 44: Instead of three calls just do one.

`setValues({...importedFilter})`

app/editor/src/features/admin/filters/utils/filterImportExport.ts
```diff
@@ -0,0 +1,208 @@
...
+      .filter((x) => currentSettings.contributorIds.some((y: number) => y === x.id))
+      .map((x) => x.name);
+    exportSettings = { ...exportSettings, contributorNames: contributorNames };
+  }
+  exportFilter.settings = exportSettings;
+
+  return exportFilter;
+};
+
+export const parseExportedFilter = (
```
Fosol, line 114: Separate functions into their own files.  It's difficult to find things if they're all buried in one long file.

Fosol (changes requested):

app/editor/src/features/admin/filters/FilterFormImportExport.tsx
kylgarmor: cant find that method...?

Fosol (approved):

PR #1125, opened 2023-10-06
Tno 1775 scroll to the top commentary and frontpages click

fbarreta:
Turns out that when the screen is smaller (mobile) and frontpages or commentaries are clicked (and both are in the bottom), it navigates but does no scroll to the top.
So I came with a less intrusive way of scrolling, just in case of those two click situations.


app/subscriber/src/features/utils/navigateAndScroll.ts
```diff
@@ -0,0 +1,7 @@
+export const navigateAndScroll = (navigate: any, path: string) => {
```
Fosol, line 1: Change this to a hook `useNavigateAndScroll` so that we don't need to pass the `navigate` function.

Also try to never use `any`.  We only use it when we have to.  Once you convert this to a hook you won't need this parameter anyway.

app/subscriber/src/features/utils/navigateAndScroll.ts
```diff
@@ -0,0 +1,7 @@
...
+  navigate(path);
```
Fosol, line 2: Pull in the `navigate` function with the `useNavigate()`

app/subscriber/src/features/utils/navigateAndScroll.ts
```diff
@@ -0,0 +1,7 @@
...
+  const cont = document.getElementsByClassName('contents-container')[0];
```
Fosol, line 3: Should we use a `ref` instead of DOM search?

Fosol (changes requested):

app/subscriber/src/features/utils/navigateAndScroll.ts
jdtoombs: ref may improve maintainability so we don't rely on `className` - not a big deal though

karmour:
I think I was mistakenly added as a Reviewer on this project, but don't have any connection to it. Please remove me so that I'm not slowing you down.

app/subscriber/src/store/hooks/subscriber/useNavigateAndScroll.ts
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { useNavigate } from 'react-router-dom';
+
+export const useNavigateAndScroll = () => {
+  const navigate = useNavigate();
+
+  const controller = React.useMemo(
```
Fosol, line 7: Since this hook only returns a single function you don't need to return an object.  Do the following.

```
return React.useCallback((path: string) => {
navigate(path);
ref.current?.scrollTo(0, 0);
}, [navigate]);
```

app/subscriber/src/features/landing/Landing.tsx
```diff
@@ -32,6 +32,7 @@ export const Landing: React.FC = () => {
   const [activeItem, setActiveItem] = React.useState<string>(SidebarMenuItems.home.label);
   const [{ userInfo }] = useApp();
   const navigate = useNavigate();
+  const contentFrame = React.useRef();
```
Fosol: You will need to attach the ref to the component being scrolled to I believe

app/subscriber/src/features/front-pages/FrontPages.tsx
```diff
@@ -39,7 +39,7 @@ export const FrontPages = () => {
             alt={s.id}
             className="front-page"
             src={s.url}
-            onClick={() => navigate(`/view/${s.id}`)}
+            onClick={() => navigateAndScroll.goTo(`/view/${s.id}`)}
```
Fosol, line 42: You will need to include the ref to the component being scrolled to.

Fosol (changes requested):

app/subscriber/src/features/utils/navigateAndScroll.ts
fbarreta: The frame that need to be scrolled is into the Landing file, for that the hierarchy needs to be built up.
Since it's just a scroll into the DOM in two specific situations, I was talking to @jdtoombs about it, and we are ok, since it would involver bigger changes.

app/subscriber/src/features/landing/Landing.tsx
fbarreta: It was sent by mistake after some tries. The file will remain unchanged.

app/subscriber/src/features/utils/navigateAndScroll.ts
Fosol: Ah so it's way too far up to get a good reference on it.  One of the headaches with React.

fbarreta: Unfortunately ... yes ... :(

Fosol (approved):

PR #1133, opened 2023-10-10
TNO-1861 Add frontpages session

fbarreta:
Just added the menu option to load the frontpages by it's own.

Some changes to be aware off:

- Filter option in "_libs/net/dal/Services/ContentService.cs_" was change to 10 (Front page images) the previous was 11 (Corporate Calendar). It's it safe to keep it hardcoded and it will reflect the same in production ?
- Later work will be captured in an upcoming ticket since @bbjornholt have the idea of having this session as a photo gallery.


libs/net/dal/Services/ContentService.cs
```diff
@@ -179,7 +179,7 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindFr
     {
         var productQueries = new List<Func<QueryContainerDescriptor<API.Areas.Services.Models.Content.ContentModel>, QueryContainer>>();
         var today = DateTime.Today.ToUniversalTime();
-        productQueries.Add(q => q.Raw(@"{""match"": {""productId"": 11}}"));
+        productQueries.Add(q => q.Raw(@"{""match"": {""productId"": 10}}"));
```
Fosol, line 182: We can't hardcode this.  It shouldn't have been hardcoded in the first place.
Additionally, we don't want searches to use NEST going forward.
We have a way to configure setting values that are essentially constants for each environment.
Talk to @jdtoombs about using an Elasticsearch query from the web app, and how to use Settings.

Fosol (changes requested):

libs/net/dal/Services/ContentService.cs
fbarreta: That's nice. I'll work on this.

libs/net/dal/Migrations/1.8.80/Up/PostUp/00-Settings.sql
Fosol, line 1: This isn't an ideal solution due to the configurable nature of filters and their names.  Remove the DB migration.  Add a toast error to the page to inform the user that a setting needs to be configured.  We have other examples of this.

app/subscriber/src/features/todays-front-pages/TodaysFrontPages.tsx
```diff
@@ -0,0 +1,96 @@
...
+      try {
+        const res: unknown = await findContentWithElasticsearch(filter, false);
+        setResults(res);
+      } catch {}
+    },
+    [findContentWithElasticsearch],
+  );
+
+  React.useEffect(() => {
+    const id = settings.find((s) => s.name === 'FrontpageFilter')?.value;
```
Fosol, line 36: Add a `toast.error("FrontpageFilter setting needs to be configured.")`

Fosol (changes requested):
It's looking very close.  Great work.  Just a couple minor changes.
- Remove the DB migration.
- Add a toast to inform user that we need to configure the setting.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
Fosol, line 1: Remove this file

Fosol (changes requested):

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
Fosol, line 1: Your branch history is off.  This file should not be part of your PR.

Fosol (changes requested):

Fosol (approved):

Fosol (changes requested):
The build shows some linting issues.

Fosol (approved):

PR #1137, opened 2023-10-11
TNO-1835: Fix evening overview clip options

jdtoombs:
Not sure exactly what time it did this; however, the timezone was wrong so it would jump forward a day on the content filter resulting in no options when doing the evening report. If you tested it before 5ish it would work but not otherwise.

image

### DIFF
--- a/app/editor/src/features/reports/av-overview/OverviewGrid.tsx
+++ b/app/editor/src/features/reports/av-overview/OverviewGrid.tsx
@@ -43,7 +43,7 @@ export const OverviewGrid: React.FC<IOverviewGridProps> = ({ editable = true, in
    */
   const shouldFetch = React.useMemo(() => {
     return !values.sections[index].startTime.includes('_') && !!values.sections[index].startTime;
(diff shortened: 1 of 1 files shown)


app/editor/src/features/reports/av-overview/OverviewGrid.tsx
```diff
@@ -38,28 +38,36 @@ export const OverviewGrid: React.FC<IOverviewGridProps> = ({ editable = true, in
...
+    return !values.sections[index].startTime.includes('_') && !!values.sections[index].startTime;
+  }, [index, values.sections]);
+
   /** fetch pieces of content that are related to the series to display as options for associated clips, search for clips published after the start time if it is specified - otherwise filter based on that day.*/
   React.useEffect(() => {
-    findContent({
-      seriesId: values.sections[index].seriesId,
-      publishedStartOn: !!values.sections[index].startTime
-        ? moment()
-            .utcOffset(0)
```
jdtoombs, line 47: culprit

Fosol (approved):

PR #1138, opened 2023-10-11
Subscriber site double scrollbar fix

jdtoombs:
Since our menu has been growing vertically it was causing some funky behaviour with scroll bars. The sidebar height is now more dynamic depending on the users device.

sidebarheight

### DIFF
--- a/app/subscriber/src/components/sidebar/styled/CustomSidebar.tsx
+++ b/app/subscriber/src/components/sidebar/styled/CustomSidebar.tsx
@@ -30,7 +30,7 @@ export const CustomSidebar = styled(Sidebar)`
         return props.theme.css.testBackgroundColor;
       else return props.theme.css.productionBackgroundColor;
     }};
-    padding: 0.65rem;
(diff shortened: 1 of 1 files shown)


app/subscriber/src/components/sidebar/styled/CustomSidebar.tsx
```diff
@@ -67,14 +67,18 @@ export const CustomSidebar = styled(Sidebar)`
       padding-left: 6px !important;
       padding-right: 6px !important;
     }
+    @media (max-height: 770px) {
+      /* important tag needed as we are fighting the library */
```
Fosol, line 71: What library are we fighting with?

Fosol (approved):

app/subscriber/src/components/sidebar/styled/CustomSidebar.tsx
jdtoombs: react-pro-sidebar

PR #1139, opened 2023-10-12
TNO-1531 Add collections to folders

Fosol:
The Editor app now has a way to administrate Folders.  A Filter can be applied and run to find content to be added to a Folder.

A Folder Collection Service will now Consume all `index` messages in Kafka to determine if the content should be added to any Folders.  This means Folders can automatically be filled with content that matches a specified Filter.

## Summary

- Added collections to folders
- Added DB migration 1.0.73
- Published tno-core:0.0.340
- Added Folder Collection Service
- Added Content list component
- Added Overlay component
- Added Elasticsearch types.  This provides Typescript strong types for our elasticsearch queries.
- Added Strongly typed Folder and Filter `Settings`.

## Folder Admin List

image

## Folder Admin Details

image

## Folder Admin Collection

image

## Folder Admin Content

image


api/net/Areas/Services/Controllers/ConnectionController.cs
```diff
@@ -52,12 +53,11 @@ public ConnectionController(IConnectionService serviceConnection, IOptions<JsonS
...
     [Produces(MediaTypeNames.Application.Json)]
     [ProducesResponseType(typeof(ConnectionModel), (int)HttpStatusCode.OK)]
-    [ProducesResponseType((int)HttpStatusCode.NoContent)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Connection" })]
     public IActionResult FindById(int id)
     {
-        var result = _serviceConnection.FindById(id);
-        if (result == null) return new NoContentResult();
+        var result = _serviceConnection.FindById(id) ?? throw new NoContentException();
```
Fosol, line 60: Switched to exception to return a BadRequest to keep it aligned with all other implementations.

api/net/Areas/Subscriber/Controllers/FilterController.cs
```diff
@@ -4,7 +4,7 @@
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
 using Swashbuckle.AspNetCore.Annotations;
-using TNO.API.Areas.Admin.Models.Filter;
+using TNO.API.Areas.Subscriber.Models.Filter;
```
Fosol, line 7: @jdtoombs Fixed this to use its own models

api/net/Areas/Subscriber/Controllers/FolderController.cs
```diff
@@ -4,7 +4,7 @@
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
 using Swashbuckle.AspNetCore.Annotations;
-using TNO.API.Areas.Admin.Models.Folder;
+using TNO.API.Areas.Subscriber.Models.Folder;
```
Fosol, line 7: @jdtoombs Fixed this to use its own models

app/editor/src/features/admin/filters/FilterFormPreview.tsx
```diff
@@ -21,7 +22,7 @@ export const FilterFormPreview: React.FC = () => {
   const [results, setResults] = React.useState<any>([]);
 
   const fetchResults = React.useCallback(
-    async (filter: unknown) => {
+    async (filter: MsearchMultisearchBody) => {
```
Fosol, line 25: We now have strongly typed Elasticsearch models.

app/editor/src/features/admin/filters/FilterFormQuery.tsx
```diff
@@ -151,32 +152,23 @@ export const FilterFormQuery: React.FC = () => {
...
-        <Row>
+        <Row gap="1rem">
           <label>Search for Keywords in: </label>
-          <Checkbox
-            id="chk_inHeadline"
-            checked={values.settings.inHeadline ?? false}
-            onChange={(e) => {
-              updateQuery('inHeadline', e.target.checked);
-            }}
+          <FormikCheckbox
```
Fosol, line 157: @CelloLiu We have FormikCheckbox and Checkbox.  Use them appropriately.  Also both support their own `label` property.

api/net/Areas/Services/Controllers/UserController.cs
```diff
@@ -48,14 +49,11 @@ public UserController(IUserService service)
...
-    [ProducesResponseType((int)HttpStatusCode.NoContent)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "User" })]
     public IActionResult FindById(int id)
     {
-        var result = _service.FindById(id);
-        if (result == null)
-            return NoContent();
-
+        var result = _service.FindById(id) ?? throw new NoContentException();
```
kylgarmor, line 56: why throw an exception here? Seems like a valid scenario that we might not have a user with a specific id.  Feels like we are pushing the responsibility back to the caller to ensure they only send valid ids?

app/editor/src/components/link/Link.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import * as styled from './styled';
+
+export interface ILinkProps extends React.HTMLAttributes<HTMLDivElement> {
+  to: string;
+  options?: NavigateOptions;
+}
+
+/**
+ * Provides a link that can open a new tab when ctrl+click is pressed.
```
kylgarmor, line 11: nice!

app/editor/src/features/admin/filters/FilterFormQuery.tsx
```diff
@@ -71,7 +71,7 @@ export const FilterFormQuery: React.FC = () => {
 
   return (
     <>
-      <Col>
+      <Col gap="0.5rem">
```
kylgarmor, line 74: couple of inline styles here, should they be moved to style classes?

app/subscriber/src/features/press-gallery/PressGallery.tsx
```diff
@@ -24,27 +26,20 @@ export const PressGallery: React.FC = () => {
...
-      inByline: true,
-      inHeadline: true,
-      inStory: true,
-      searchUnpublished: false,
-      defaultOperator: 'or',
-    };
-  }, [filterAdvanced.startDate]);
-
-  const [pressSettings, setPressSettings] = React.useState<IFilterSettingsModel>(defaultSettings);
+  const [pressSettings, setPressSettings] = React.useState<IFilterSettingsModel>(
```
kylgarmor, line 29: might be nit picking here, but I actually prefer the name `defaultSettings` here.  If I was looking at this component, I would immediately make a connection as the "these are the default settings for the component" as opposed to `pressSettings` which could be "settings for pressing something"?  I try to follow the pattern of that `defaultSettings` name in components so that I know where I can find the things used "by default"

libs/npm/core/src/components/content/DroppableContentContainer.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
kylgarmor, line 1: will this replace the one that Jonny implemented for all drag and drop content?

libs/npm/core/src/components/sentiment/Sentiment.tsx
```diff
@@ -0,0 +1,35 @@
...
+  /** The title attribute */
+  title?: string;
+}
+
+/**
+ * Component that does a quick check on the toning value in order to determine which icon to present
+ * @param tone the numeric value used to determine the icon to present
+ * @returns An svg image that represents the tone
+ */
+export const Sentiment: React.FC<ISentimentProps> = ({ value, title }) => {
```
kylgarmor, line 18: nice!

libs/npm/core/src/hooks/api-editor/admin/useApiAdminActions.ts
```diff
@@ -20,15 +20,15 @@ export const useApiAdminActions = (
 
   return React.useRef({
     findAllActions: () => {
-      return api.get<IActionModel[], AxiosResponse<IActionModel[]>, any>(`/admin/actions`);
+      return api.get<never, AxiosResponse<IActionModel[]>, any>(`/admin/actions`);
```
kylgarmor, line 23: what is the significance of `never` here?

kylgarmor (approved):

api/net/Areas/Services/Controllers/UserController.cs
Fosol: I generally prefer to return a 204, however it does result in more work on the client.  Which means by throwing a 400 we are actually giving the client less work since it already handles errors.  It also doesn't matter whether the caller gets a 204 or a 400, it's still their responsibility to request valid content.

app/editor/src/features/admin/filters/FilterFormQuery.tsx
Fosol: Not for these type of minor styling changes.  It's actually more work to setup random classes just for a basic gap.  Now, if we had some standardized css layout that everything inherited then I would entirely agree.  However, we've never received this, nor has it been implemented.

app/subscriber/src/features/press-gallery/PressGallery.tsx
Fosol: I didn't rename the state variables.  I only replaced the const `defaultSettings` with a function `createFilterSettings`.

libs/npm/core/src/components/content/DroppableContentContainer.tsx
Fosol: The goal is to have a shared component that is reused to eliminate the amount of work it is to add the functionality everywhere it is needed.  We'll switch over to using the new component slowly.

libs/npm/core/src/hooks/api-editor/admin/useApiAdminActions.ts
Fosol: Essentially we `never` pass a value in a get body.  I find the Axios generic types annoying.

PR #1144, opened 2023-10-13
fix: #tno-1877 - fixed name field length

kylgarmor:
- updated helper method to truncate excessively long names

### DIFF
--- a/app/editor/src/features/admin/reports/constants/defaultReportSchedule.ts
+++ b/app/editor/src/features/admin/reports/constants/defaultReportSchedule.ts
@@ -25,5 +25,9 @@ export const defaultReportSchedule = (
   dayOfMonth: 0,
 });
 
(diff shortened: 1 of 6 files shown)


libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -3651,7 +3651,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<string>("Name")
                         .IsRequired()
                         .HasMaxLength(50)
-                        .HasColumnType("character varying(50)")
+                        .HasColumnType("character varying(100)")
```
Fosol, line 3654: You need to edit the ScheduleConfiguration or this will get overwritten.

Fosol (changes requested):

Fosol (approved):

PR #1148, opened 2023-10-14
fix: #tno-1877 - fixed name field length

kylgarmor:
- updated helper method to truncate excessively long names
- found another place the long name was being used

### DIFF
--- a/libs/net/dal/Migrations/20231014003235_1.0.75.cs
+++ b/libs/net/dal/Migrations/20231014003235_1.0.75.cs
@@ -1,0 +1,63 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+using TNO.DAL;
+
+#nullable disable
+
+namespace TNO.DAL.Migrations
+{
+    /// <inheritdoc />
+    public partial class _1075 : SeedMigration
+    {
+        /// <inheritdoc />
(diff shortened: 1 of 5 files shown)


kylgarmor:
very confused how this worked when I tested it locally before.  Also figured out that the sql was just overkill so got rid of it.

libs/net/dal/Migrations/20231014003235_1.0.75.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace TNO.DAL.Migrations
+{
+    /// <inheritdoc />
+    public partial class _1075 : SeedMigration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
+            migrationBuilder.AlterColumn<string>(
```
Fosol, line 15: This isn't good as it highlights your prior migration was incorrect.  I don't think it will break anything thankfully because the change will not cause a conflict.  

If you make a change after generating a migration you need to rollback first and remove the migration.  Then create a new migration.

`make db-rollback n=1.0.73`
`make db-remove`
`make db-add n=1.0.74`
`make db-update`

Then make sure your rollback works.
`make db-rollback n=1.0.73`
`make db-update`

Fosol (approved):

PR #1159, opened 2023-10-18
TNO-1826 Add unzip to Image Service

Fosol:
The Image Service can now handle ZIP files, and will also delete the original files based on configuration.

## Ingest Configuration

image

### DIFF
--- a/app/editor/src/features/admin/ingests/configurations/FrontPage.tsx
+++ b/app/editor/src/features/admin/ingests/configurations/FrontPage.tsx
@@ -67,6 +67,7 @@ export const FrontPage: React.FC = (props) => {
           />
         </Col>
       </Row>
+      <FormikCheckbox label="Delete original files" name="configuration.deleteOriginal" />
       <p>
         Front page images are often delivered the evening before the publish date. Use a regular
         expression to parse the file name to determine the published on date. Use regex group names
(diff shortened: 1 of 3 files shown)


app/editor/src/features/admin/ingests/configurations/FrontPage.tsx
```diff
@@ -67,6 +67,7 @@ export const FrontPage: React.FC = (props) => {
           />
         </Col>
       </Row>
+      <FormikCheckbox label="Delete original files" name="configuration.deleteOriginal" />
```
Fosol, line 70: A way to control whether the original files are deleted.  This is required because the SFTP location slowly files up and it takes longer each time to parse the files.

libs/net/dal/Migrations/Initial/Up/PostUp/02-Ingest.sql
```diff
@@ -483,9 +483,9 @@ INSERT INTO public.ingest (
   , (SELECT id FROM public.source WHERE code = 'POST') -- source_id
   , 'POST' -- topic
   , frontPageId -- product_id
-  , '{ "path": "binaryroot/<date>",
+  , '{ "path": "infomart/papers",
```
Fosol, line 486: This won't be applied, but it captures what they should be going forward.

services/net/image/ImageAction.cs
```diff
@@ -112,20 +114,39 @@ public ImageAction(IApiService api, IOptions<ImageOptions> options, ILogger<Imag
...
             {
-                var newReference = CreateContentReference(manager.Ingest, file.Name);
+                // Downloading the file can result in multiple Image Services performing the same task.
+                // However, we need to unzip the images before creating a content reference.
+                var filePath = await DownloadFileAsync(client, manager.Ingest, remotePath.CombineWith(file.Name), deleteOriginal);
+
+                // If a zip file it needs to be unzipped first.
+                if (Path.GetExtension(file.Name).Equals(".zip", StringComparison.OrdinalIgnoreCase))
+                {
+                    UnzipFile(filePath, outputPath, true);
```
Fosol, line 139: We delete the zip file after unzipping.

PR #1160, opened 2023-10-18
Tno 1802 report form validation

fbarreta:
Added notification if the there's a same name report.
Added validation using yup, highlight tabs with validation errors.

image
Validation messages.

image
Tab warning.


app/editor/src/features/admin/reports/ReportFormDetails.tsx
```diff
@@ -54,7 +54,6 @@ export const ReportFormDetails: React.FC = () => {
         <FormikText
           name="name"
           label="Name"
-          required
```
Fosol, line 57: How do we highlight a required input without this?

app/subscriber/src/features/my-reports/admin/ReportAdmin.tsx
```diff
@@ -63,38 +65,50 @@ export const ReportAdmin: React.FC<IReportAdminProps> = ({ path: defaultPath = '
...
-      } else {
-        setReport({
-          ...report,
-          sections: report.sections.map((section, index) => ({
-            ...section,
-            expand: values.sections[index].expand,
-          })),
-          hideEmptySections: getHideEmpty(report.sections),
+      try {
+        const sameNameReport = await findReports({
```
Fosol, line 69: When someone attempts to save a report with the same name it should return a message you can parse to determine if a report already exists with the name.  

Doing it this way requires an additional request to the API before submitting.

Can you see how it would work with the original error response and parsing the error?

app/editor/src/features/admin/reports/ReportFormDetails.tsx
fbarreta: Yup handles it with the Schema, and it gives it better visual output, and messaging. (I'll add a screenshot)

Fosol: What about the initial visual output?  The user won't know which fields are required until they attempt to save?

app/subscriber/src/features/my-reports/admin/ReportAdmin.tsx
fbarreta: I haven't came into a final conclusion yet, but with a few minutes talking with Jonny, I couldn't intercept the global toast error, and the error I can capture at the component level is more generic, without the Duplicate Key Info. (That was my very initial goal ... but I'm really struggling with it)

app/editor/src/features/admin/reports/ReportFormDetails.tsx
fbarreta: Yes, they will. Works exactly like the usual validation.

Fosol (approved):

PR #1172, opened 2023-10-20
TNO-1845 Fix Scheduler Service

Fosol:
Updated the Scheduler Service to use the correct timezone.


services/net/scheduler/SchedulerManager.cs
```diff
@@ -101,14 +101,17 @@ private bool VerifySchedule(EventScheduleModel scheduledEvent)
     {
         var schedule = scheduledEvent.Schedule ?? throw new InvalidOperationException($"Schedule is missing from event {scheduledEvent.Id}:{scheduledEvent.Name}");
 
-        var now = DateTime.Now;
+        var now = DateTime.Now.ToTimeZone(this.Options.TimeZone);
```
Fosol, line 104: The timezone will ensure it runs based on the configured service settings.

PR #1179, opened 2023-10-24
TNO-1802 Function to calculate next report send

fbarreta:
Preview next report send based on the schedule.

image


Fosol (approved):
Looks great.  Thanks for adding documentation too!

PR #1180, opened 2023-10-24
TNO-1650 Fix Event Handler Service

Fosol:
Fixed issue where Event Handler Service would throw concurrency error due to the race condition.

Added other team members to contributors so they receive PR emails.

Updated resources for Index Service to improve performance.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,7 +6,7 @@
 
 # review when someone opens a pull request.
 
-- @Fosol @jdtoombs @kylgarmor @CelloLiu
+- @Fosol @jdtoombs @kylgarmor @CelloLiu @fbarreta @mihailistov
 
 # Order is important; the last matching pattern takes the most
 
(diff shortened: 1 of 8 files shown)


.github/CODEOWNERS
```diff
@@ -6,7 +6,7 @@
 
 # review when someone opens a pull request.
 
-- @Fosol @jdtoombs @kylgarmor @CelloLiu
+- @Fosol @jdtoombs @kylgarmor @CelloLiu @fbarreta @mihailistov
```
Fosol, line 9: @fbarreta @mihailistov you should now receive emails when a PR is created.  These emails go to your github account.

services/net/event-handler/EventHandlerManager.cs
```diff
@@ -257,6 +257,8 @@ private async Task ProcessEventHandlerAsync(ConsumeResult<string, EventScheduleR
                 await this.Api.RemoveContentFromFolder(eventSchedule.FolderId.Value);
                 this.Logger.LogInformation("Event schedule cleaned folder.  Key: {key}, Event ID: {eventId}", result.Message.Key, request.EventScheduleId);
 
+                // Need to fetch the latest because it could have been updated recently.
+                eventSchedule = await this.Api.GetEventScheduleAsync(request.EventScheduleId) ?? throw new NoContentException("Event schedule no longer exists");
```
Fosol, line 261: Due to the speed of Kafka we need to do this or we run into concurrency issues.

PR #1185, opened 2023-10-25
Tno 1813 add byline on report

fbarreta:
Add "Show Byline" settings, and change Custom Report template to check this config.


libs/net/dal/Migrations/1.0.81/Up/PostUp/00-Report_Template.sql
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+}
+
+<p style="font-size: 9pt; margin-top: 0.5rem;">
+  Terms of Use - This summary is a service provided by Government Communications and Public Engagement and is only intended for original addressee. All content is the copyrighted property of a third party creator of the material.
+  Copying, retransmitting, archiving, redistributing, selling, licensing, or emailing the material to any third party or any employee of the Province who is not authorized to access the material is prohibited.
+</p>
+    '
+WHERE "name" = 'Custom Report';
```
Fosol, line 201: @kylgarmor can you review to ensure we don't lose you latest changes?

libs/net/models/Settings/ReportSectionsSettingsModel.cs
```diff
@@ -7,6 +7,7 @@ public class ReportSectionsSettingsModel
 {
     #region Properties
     public bool UsePageBreaks { get; set; }
+    public bool HideEmpty { get; set; }
```
Fosol, line 10: I don't think this is required.  The `hideEmpty` checkbox on the top form merely is a way to quickly turn on or off the `section.hideEmpty` checkboxes.

libs/net/models/Settings/ReportContentSettingsModel.cs
```diff
@@ -25,6 +26,7 @@ public ReportContentSettingsModel(Dictionary<string, object> settings, JsonSeria
         this.ExcludeReports = settings.GetDictionaryJsonValue("excludeReports", Array.Empty<int>(), options)!;
         this.ShowLinkToStory = settings.GetDictionaryJsonValue("showLinkToStory", false, options)!;
         this.HighlightKeywords = settings.GetDictionaryJsonValue("highlightKeywords", false, options)!;
+        this.IncludeStory = settings.GetDictionaryJsonValue("includeStory", false, options)!;
```
Fosol, line 29: What is this option for?

libs/net/models/Settings/ReportSectionSettingsModel.cs
```diff
@@ -51,6 +53,7 @@ public ReportSectionSettingsModel(JsonDocument settings, JsonSerializerOptions o
         this.HideEmpty = settings.GetElementValue("hideEmpty", false, options)!;
         this.GroupBy = settings.GetElementValue("groupBy", "", options)!;
         this.SortBy = settings.GetElementValue("sortBy", "", options)!;
+        this.ShowContent = settings.GetElementValue("showContent", false, options);
```
Fosol, line 56: What is this option for?

libs/net/models/Settings/ReportSectionsSettingsModel.cs
fbarreta: I really don't know. I just added to be able to compile (and test) the template I had (I had errors on preview on these options). So I left to ask about it today.
If the template doesn't require:
IncludeStory 
ShowContent 
HideEmpty 
I'll just revert this files.
@kylgarmor any thoughts on this ?

libs/net/models/Settings/ReportContentSettingsModel.cs
fbarreta: Same as first comment above.

libs/net/models/Settings/ReportSectionSettingsModel.cs
fbarreta: Same as first comment above.

libs/net/models/Settings/ReportSectionsSettingsModel.cs
Fosol: Each section has an option for `hideEmpty` that controls how the template behaves.  But the overall sections settings doesn't require it.  The UI form has a checkbox there to flip the checkbox on all the sections.

libs/net/models/Settings/ReportContentSettingsModel.cs
Fosol: Each section has a setting `showFullStory` which sounds like it is related.  I don't see why we need this property here.

libs/net/models/Settings/ReportSectionSettingsModel.cs
Fosol: I don't understand what `ShowContent` is compared to `ShowFullStory`?

libs/net/models/Settings/ReportSectionsSettingsModel.cs
fbarreta: Every time I tried to preview the report without this fields, I have this error:
image
It was just a workaround to make sure my real changes would work.
I can (and probably will) remove all this additional fields, but I'll need some clarification, why I'm having these errors.

fbarreta: I was only using an old template to test it.
Unnecessary fields removed ... good to go !!!

Fosol (changes requested):
I've published `tno-core:0.0.349` for you.  You'll need to update both the editor and subscriber apps.  Then `yarn install` in both.

libs/net/dal/Migrations/1.0.81/Up/PostUp/00-Report_Template.sql
```diff
@@ -0,0 +1,203 @@
...
+                  }
+                </div>
+                <h3>@content.Headline</h3>
+                <div>@content.Source?.Name</div>
+                <div>@content.PublishedOn?.AddHours(utcOffset).ToString("dddd, MMMM d, yyyy")</div>
+                @if (!string.IsNullOrWhiteSpace(content.Page))
+                {
+                    <div>Page @content.Page</div>
+                }
+                @if (!string.IsNullOrWhiteSpace(content.Byline) && Settings.Headline.ShowByline)
```
kylgarmor, line 138: This is the only line changed in the template since my change.  The key point is this new check
`&& Settings.Headline.ShowByline`

kylgarmor (approved):
looks good

libs/net/dal/Migrations/1.0.81/Up/PostUp/00-Report_Template.sql
fbarreta: Exactly ! tks !

kylgarmor (approved):

Fosol (approved):

PR #1186, opened 2023-10-25
TNO-1612: Bug fix

jdtoombs:
Was previously trying to fetch instance with `reportId` instead of `instanceId`.

One question down below:

### DIFF
--- a/app/subscriber/src/features/my-reports/hooks/useColumns.tsx
+++ b/app/subscriber/src/features/my-reports/hooks/useColumns.tsx
@@ -51,11 +51,6 @@ export const useColumns = (
       width: '100px',
       cell: (cell) => (
         <Row gap="0.5rem">
-          <FaEye
-            className="btn-link"
-            onClick={() => navigate(`/report/instances/${cell.original.id}/view`)}
-            title="Instances"
-          />
           <FaGear
             className="btn-link"
             onClick={() => navigate(`/reports/${cell.original.id}`)}
--- a/app/subscriber/src/features/my-reports/hooks/useColumns.tsx
+++ b/app/subscriber/src/features/my-reports/hooks/useColumns.tsx
@@ -1,4 +1,4 @@
-import { FaEdit, FaEye, FaFileAlt, FaTrash } from 'react-icons/fa';
+import { FaEdit, FaFileAlt, FaTrash } from 'react-icons/fa';
 import { FaChartPie, FaGear } from 'react-icons/fa6';
 import { Link } from 'react-router-dom';
 import { useNavigate } from 'react-router-dom';
--- a/app/subscriber/src/features/my-reports/utils/isAutoSend.ts
+++ b/app/subscriber/src/features/my-reports/utils/isAutoSend.ts
@@ -1,7 +1,7 @@
 import { IReportModel } from 'tno-core';
 
 export const isAutoSend = (report: IReportModel) => {
-  if (report.events === undefined) return;
+  if (report.events === undefined) return false;
   return (
     report.isEnabled &&
     report.events &&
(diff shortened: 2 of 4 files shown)


app/subscriber/src/features/my-reports/hooks/useColumns.tsx
```diff
@@ -51,11 +51,16 @@ export const useColumns = (
...
-            className="btn-link"
-            onClick={() => navigate(`/report/instances/${cell.original.id}/view`)}
-            title="Instances"
-          />
+          <Show visible={!!cell.original.instances.length}>
+            <FaEye
+              className="btn-link"
+              onClick={() => {
+                const instances = cell.original.instances;
+                navigate(`/report/instances/${instances[instances.length - 1].id}/view`);
```
jdtoombs, line 59: Does this make sense to get the latest report instance?

Fosol: I would remove this column for now.  It causes too much confusion.  The report instance page you were building was only meant as a page to view reports that have been sent out by email.

app/subscriber/src/features/my-reports/utils/isAutoSend.ts
```diff
@@ -1,6 +1,7 @@
 import { IReportModel } from 'tno-core';
 
 export const isAutoSend = (report: IReportModel) => {
+  if (report.events === undefined) return;
```
Fosol, line 4: Why return `undefined`?

app/subscriber/src/features/my-reports/utils/isAutoSendDisabled.ts
```diff
@@ -1,6 +1,7 @@
 import { IReportModel } from 'tno-core';
 
 export const isAutoSendDisabled = (report: IReportModel) => {
+  if (report.events === undefined) return;
```
Fosol, line 4: Why return `undefined`?

libs/net/dal/Services/ReportService.cs
```diff
@@ -123,6 +123,7 @@ public IEnumerable<Report> FindMyReports(int userId)
             .Include(f => f.Owner)
             .Include(r => r.Template).ThenInclude(t => t!.ChartTemplates)
             .Include(r => r.Sections)
+            .Include(r => r.Instances)
```
Fosol, line 126: Please don't do this.  Think about how much data will be returned each time.

Fosol (changes requested):

libs/net/dal/Services/ReportService.cs
Fosol: Many reports run every day.  This means in a year this would return 365 extra items.  Of which none will ever be viewed again.

app/subscriber/src/features/my-reports/utils/isAutoSend.ts
jdtoombs: Not sure if this was a weird one off in my local, but I was getting an error because report.events was undefined when the below was trying to execute

app/subscriber/src/features/my-reports/hooks/useColumns.tsx
jdtoombs: Column removed

app/subscriber/src/features/my-reports/utils/isAutoSend.ts
Fosol: Then return false so that the function `isAutoSend` is a good function.

app/subscriber/src/features/my-reports/utils/isAutoSendDisabled.ts
Fosol: Then return false so that the function `isAutoSendDisabled` is a good function.

Fosol (changes requested):

Fosol (approved):

PR #1189, opened 2023-10-25
feat: #tno-1940 - tweaks to Dockerfiles

kylgarmor:
- switched from use of sdk image to aspnet
- commented out unneccessary stuff from api Dockerfile


Fosol (approved):
Very nice

PR #1190, opened 2023-10-26
[Now Review Ready ✅] TNO-1904: Add events filter

jdtoombs:
In this PR:

- persist date so when the user views a piece of content and navigates back, the date filter is on the appropriate date
- add events filter to the home screen to show contents tagged with the events product id
- rework home page to use elastic queries like the rest of the application

tno-1904

### DIFF
--- a/app/subscriber/src/components/date-filter/DateFilter.tsx
+++ b/app/subscriber/src/components/date-filter/DateFilter.tsx
@@ -11,7 +11,9 @@ export interface IDateFilterProps {}
 export const DateFilter: React.FC<IDateFilterProps> = () => {
   const [{ filterAdvanced }, { storeFilterAdvanced }] = useContent();
   /** default to today's date or the filters previously saved date if present */
-  const [date, setDate] = React.useState<Date>(new Date(filterAdvanced.startDate ?? ''));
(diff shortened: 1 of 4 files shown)


app/subscriber/src/features/content/list-view/interfaces/IContentListFilter.ts
```diff
@@ -7,6 +7,7 @@ export interface IContentListFilter {
   boldKeywords?: boolean;
   contentTypes: ContentTypeName[];
   excludeSourceIds?: number[];
+  excludeProductIds?: number[];
```
Fosol, line 10: Remember we are moving away from this to use Elasticsearch queries instead.

jdtoombs: Thought of this as I was falling asleep last night, changing this!

jdtoombs: Adjusted to use elastic queries, also fixed a bug with the date persisting when navigating back (much similar to how it would lose state on other filters) @Fosol

Fosol (approved):

PR #1194, opened 2023-10-26
Fix login

Fosol:
Resolve the login page issue caused by SignalR changes.

### DIFF
--- a/app/editor/src/store/hooks/signalr/useApiHub.ts
+++ b/app/editor/src/store/hooks/signalr/useApiHub.ts
@@ -9,7 +9,7 @@ import { useAppStore } from 'store/slices';
 import { MessageTargetName, Settings, useKeycloakWrapper } from 'tno-core';
 
 const url = Settings.ApiPath + '/hub';
-let hub: HubConnection;
+let hub: HubConnection | null = null;
 
 export interface IHubController {
   /** The state of the hub connection. */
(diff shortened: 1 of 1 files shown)


app/editor/src/store/hooks/signalr/useApiHub.ts
```diff
@@ -9,7 +9,7 @@ import { useAppStore } from 'store/slices';
 import { MessageTargetName, Settings, useKeycloakWrapper } from 'tno-core';
 
 const url = Settings.ApiPath + '/hub';
-let hub: HubConnection;
+let hub: HubConnection | null = null;
```
Fosol, line 12: It begins as null.

app/editor/src/store/hooks/signalr/useApiHub.ts
```diff
@@ -52,22 +52,22 @@ export const useApiHub = (): IHubController => {
         .withAutomaticReconnect()
         .build();
       hub.onclose(() => {
-        changeHubState(hub.state);
+        changeHubState(hub!.state);
```
Fosol, line 55: These will never be null.

app/editor/src/store/hooks/signalr/useApiHub.ts
```diff
@@ -80,26 +80,26 @@ export const useApiHub = (): IHubController => {
   return {
     state: hubState,
     send: async (target: string, args: any[]) => {
-      if (hub.state === HubConnectionState.Connected) await hub.send(target, ...args);
+      if (hub?.state === HubConnectionState.Connected) await hub.send(target, ...args);
```
Fosol, line 83: Handle nullable

PR #1195, opened 2023-10-26
feat: #tno-1864 - changed caching

kylgarmor:
- changed mechanism for determining when to get from or add to cache.
 - now based on the hash of the template plus key
 - key is now based on the caller saying isPreview
 - preview and final reports are considered different
 - changed report preview UI so that it calls a "warm up" preview
 - TODO - make calls to elastic in parallel

### DIFF
--- a/api/net/Areas/Admin/Controllers/ReportController.cs
+++ b/api/net/Areas/Admin/Controllers/ReportController.cs
@@ -193,7 +193,9 @@ public class ReportController : ControllerBase
         {
             RequestorId = user.Id,
             To = to,
-            UpdateCache = true,
+            // no longer utilized, the caching mechanism
+            // determines whether to update or not
+            // UpdateCache = true,
             GenerateInstance = false
         };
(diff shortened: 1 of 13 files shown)


api/net/Areas/Admin/Controllers/ReportController.cs
```diff
@@ -193,7 +193,9 @@ public async Task<IActionResult> SendToAsync(int id, string to)
         {
             RequestorId = user.Id,
             To = to,
-            UpdateCache = true,
+            // no longer utilized, the caching mechanism
+            // determines whether to update or not
+            // UpdateCache = true,
```
Fosol, line 198: Do we need this to perhaps force an update under some condition?

Fosol (approved):

api/net/Areas/Admin/Controllers/ReportController.cs
kylgarmor: I was going to remove it, but noticed we use the property in some other cases via the message broker.  It does smell a bit funny to me...

Arguably we should switch to some sort of better solution that uses distributed caching rather than what is effectively a per service instance in memory cache. That of course is a bigger deal entirely.

PR #1196, opened 2023-10-26
TNO-1909: error message when trying to add myself as subscriber to Evening Report

CelloLiu:
TNO-1909: error message when trying to add myself as subscriber to Evening Report


libs/net/dal/Services/AVOverviewTemplateService.cs
```diff
@@ -83,6 +83,8 @@ public override AVOverviewTemplate Update(AVOverviewTemplate entity)
                 originalSubscriber.IsSubscribed = s.IsSubscribed;
             }
         });
+        entity.SubscribersManyToMany.Clear();
```
Fosol, line 86: This will result in defragmentation in the database due to deleting and readding records.
The code above deals with adding/updating/removing subscribers.

Fosol (changes requested):

libs/net/dal/Services/AVOverviewTemplateService.cs
CelloLiu: In this case it won't result in deleting and readding records actually, and it will only update the record(s) which we have touched. I can do a quick demo about it if needed, thanks.

CelloLiu: Updated, thanks Kyle for helping look into this together.

libs/net/dal/Services/AVOverviewTemplateService.cs
```diff
@@ -91,7 +91,6 @@ public override AVOverviewTemplate Update(AVOverviewTemplate entity)
         });
         entity.Sections.ForEach(section =>
         {
-            section.Template = entity;
```
Fosol, line 94: Have you tested adding and updating sections?  This was most likely needed when creating a new section an linking it to an existing template.

CelloLiu: Yes and it seems to be working fine so far.

Fosol (approved):

PR #1200, opened 2023-10-27
Tno 1814 highlight search terms

fbarreta:
Changes made:

- Template update
- New Extension method to highlight Key words on the section filter
- Report Instance FindById method updated to make the filter property accessible on subscriber side
- Just a few UI updates

image
Editor

image
Subscriber


Fosol (approved):
Great solution!

PR #1203, opened 2023-10-30
TNO-1855: Move advanced search

jdtoombs:
- move adv search from sidebar
- add text area when in adv search
- missing functionality pieces will be coming in the next PR
- dynamically adjust height if expanded

adv-srch

### DIFF
--- a/app/subscriber/src/components/layout/DefaultLayout.tsx
+++ b/app/subscriber/src/components/layout/DefaultLayout.tsx
@@ -24,7 +24,6 @@ export interface ILayoutProps extends React.HTMLAttributes<HTMLDivElement> {
  */
 export const DefaultLayout: React.FC<ILayoutProps> = ({ children, ...rest }) => {
   const keycloak = useKeycloakWrapper();
-  const [advancedSearch, setAdvancedSearch] = React.useState(false);
   const { setToken } = React.useContext(SummonContext);
   useToastError();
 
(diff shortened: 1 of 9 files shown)


app/subscriber/src/css/_variables.module.scss
```diff
@@ -1,88 +1,81 @@
 @import './_colors';
 @import './_fonts';
-
```
jdtoombs, line 3: trying to sort out my linting on other pc...

app/subscriber/src/components/layout/DefaultLayout.tsx
```diff
@@ -52,10 +51,10 @@ export const DefaultLayout: React.FC<ILayoutProps> = ({ children, ...rest }) =>
 
   // check if width is mobile and if so collapse the sidebar or when window is being resized
   React.useEffect(() => {
-    if (width && width < 768 && !collapsed && !advancedSearch) {
+    if (width && width < 768 && !collapsed) {
```
Fosol, line 54: Why can't we use css media query?

jdtoombs: adjusting!

jdtoombs: oh wait, it's because I need to call the hook to collapse the sidebar dynamically when the screen reaches a certain size so I needed to check width here instead of the scss

Fosol (approved):

PR #1205, opened 2023-10-31
fix: #tno-1956 - stopped content service from stripping

kylgarmor:
- tweaked paragraph sanitizing in filemonitor and syndication

### DIFF
--- a/services/net/contentmigration/migrators/PaperMigrator.cs
+++ b/services/net/contentmigration/migrators/PaperMigrator.cs
@@ -1,4 +1,5 @@
-﻿using System.Net;
+﻿using System.CommandLine.Parsing;
+using System.Net;
 using LinqKit;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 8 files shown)


services/net/syndication/SyndicationAction.cs
```diff
@@ -259,6 +260,37 @@ private SourceContent CreateSourceContent(IngestModel ingest, SyndicationItem it
         };
     }
 
+    /// <summary>
+    /// swap line breaks with paragraph tags
+    /// </summary>
+    /// <param name="text"></param>
+    /// <returns>formatted body string</returns>
+    protected string SanitizeLineBreaks(string? text)
```
Fosol, line 268: Can we move this to a shared library so that we have consistent implementation on different Ingests?

services/net/contentmigration/migrators/PaperMigrator.cs
```diff
@@ -49,8 +50,8 @@ public override SourceContent CreateSourceContent(LookupModel lookups, SourceMod
         Logger.LogDebug("NewItem.RSN: {rsn}, PublishedDateTime: {publishedDateTime}, ToUtc: {publishedDateTimeInUtc}", newsItem.RSN, publishedOnInDefaultTimeZone, publishedOnInUtc);
 
         var newsItemTitle = newsItem.GetTitle();
-        var sanitizedSummary = SanitizeLineBreaks(newsItem.Summary);
-        var sanitizedBody = SanitizeLineBreaks(newsItem.Text);
+        var sanitizedSummary = TNO.Core.Extensions.StringExtensions.ConvertTextToParagraphs(newsItem.Summary, @"\r\n?|\n|\|");
```
Fosol, line 53: One thing I like about extension methods is how they can be called.
`var sanitizedSummary = newsItem.Summary.ConvertTextToParagraph(@"\r\n?|\n|\|");`

Fosol (approved):

services/net/contentmigration/migrators/PaperMigrator.cs
kylgarmor: funnily enough I always forget about using extensions for simple types...

PR #1207, opened 2023-10-31
Mac compatibility

mihailistov:


Fosol (commented):
Who's willing to download this PR and stand up the whole environment fresh on their PC?  If you do it in another folder it shouldn't impact your currently installed setup.

api/node/Dockerfile
```diff
@@ -4,8 +4,11 @@ WORKDIR /usr/app
 COPY .yarn ./.yarn
 COPY package*.json yarn.lock .yarnrc.yml ./
 
+RUN apt-get update
+RUN apt-get install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
```
jdtoombs, line 8: @Fosol - I like your idea to make a Dockerfile.mac or something of the likes - I seem to be getting an error from this above line on Windows

jdtoombs: Seems to be the only thing failing though

db/kafka/broker/Dockerfile
```diff
@@ -1 +1,3 @@
-FROM confluentinc/cp-kafka:7.1.1
+# FROM confluentinc/cp-kafka:7.1.1
+# for M1/M2 Macs:
+FROM confluentinc/cp-kafka:7.2.2.arm64
```
Fosol, line 3: Won't this break all the PCs?

libs/net/dal/Migrations/Initial/Up/PostUp/02-Ingest.sql
```diff
@@ -463,7 +463,7 @@ INSERT INTO public.ingest (
   , 'GLOBE' -- topic
   , frontPageId -- product_id
   , '{ "path": "gandm/papers",
-      "fileName": "sv-GLB-<date>-.+?\.jpg",
+      "fileName": "sv-GLB-<date>-.+?.jpg",
```
Fosol, line 466: Why are you making these changes?  This is a regex expression and needs the leading escape character.

mihailistov: Mac thing I guess... get an error about the `\` char when running script

Fosol: It can't be removed.  You also shouldn't be running this script.  The DB migration tool should be running this script.

mihailistov: yeah i wasn't running it directly, it was failing regardless, can't recall on exactly which step, i'll revert the change 👍

api/net/Program.cs
```diff
@@ -154,6 +154,7 @@
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen(options =>
 {
+    options.ResolveConflictingActions(apiDescriptions => apiDescriptions.First());
```
Fosol, line 157: I'm curious what this does?

Fosol (approved):

api/net/Program.cs
mihailistov: sorry just saw your comment now -- i was getting an error: `Actions require unique method/path combination for Swagger 2.0` when running api swagger, this solves that error

PR #1209, opened 2023-10-31
TNO-1693: Date related Elasticsearch queries must handle time zone calculation

CelloLiu:
TNO-1693: Date related Elasticsearch queries must handle time zone calculation

### DIFF
--- a/app/editor/src/features/admin/filters/FilterSettingsForm.tsx
+++ b/app/editor/src/features/admin/filters/FilterSettingsForm.tsx
@@ -65,10 +65,11 @@ export const FilterSettingsForm: React.FC<IFilterSEttingsFormProps> = ({
       var values = { ...settings };
       values[key] = value;
       if (key === 'dateOffset') {
-        values = { ...settings, startDate: undefined, endDate: undefined };
+        values = { ...values, startDate: undefined, endDate: undefined };
       } else if (key === 'startDate' || key === 'endDate') {
-        values = { ...settings, dateOffset: undefined };
+        values = { ...values, dateOffset: undefined };
       }
+      values = { ...values, validDateOffset: key === 'dateOffset' && value === 3 };
       if (defaultPath) setFieldValue(defaultPath, values);
       else setValues(values);
       onChange?.(values);
--- a/app/editor/src/features/admin/filters/FilterSettingsForm.tsx
+++ b/app/editor/src/features/admin/filters/FilterSettingsForm.tsx
@@ -69,7 +69,6 @@ export const FilterSettingsForm: React.FC<IFilterSEttingsFormProps> = ({
       } else if (key === 'startDate' || key === 'endDate') {
         values = { ...values, dateOffset: undefined };
       }
-      values = { ...values, validDateOffset: key === 'dateOffset' && value === 3 };
       if (defaultPath) setFieldValue(defaultPath, values);
       else setValues(values);
       onChange?.(values);
--- a/app/editor/src/features/content/hooks/useElasticsearch.ts
+++ b/app/editor/src/features/content/hooks/useElasticsearch.ts
@@ -31,7 +31,11 @@ export const useElasticsearch = () => {
         startDate: filter.startDate ? moment(filter.startDate).toISOString() : undefined,
         endDate: filter.endDate ? moment(filter.endDate).toISOString() : undefined,
         dateOffset:
-          filter.startDate || filter.endDate ? undefined : filter.timeFrame ? +filter.timeFrame : 0,
+          filter.startDate || filter.endDate || filter.timeFrame === 3
+            ? undefined
+            : filter.timeFrame
+            ? +filter.timeFrame
+            : 0,
         actions: getActionFilters(filter, actions),
         seriesIds: [],
         contributorIds: [],
--- a/libs/npm/core/src/hooks/api-editor/interfaces/IFilterSettingsModel.ts
+++ b/libs/npm/core/src/hooks/api-editor/interfaces/IFilterSettingsModel.ts
@@ -32,4 +32,5 @@ export interface IFilterSettingsModel {
   tags?: string[];
   sentiment?: number[];
   sort?: ISortField[];
+  validDateOffset?: boolean;
 }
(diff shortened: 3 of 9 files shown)


app/editor/src/features/admin/filters/FilterSettingsForm.tsx
```diff
@@ -65,10 +65,11 @@ export const FilterSettingsForm: React.FC<IFilterSEttingsFormProps> = ({
       var values = { ...settings };
       values[key] = value;
       if (key === 'dateOffset') {
-        values = { ...settings, startDate: undefined, endDate: undefined };
+        values = { ...values, startDate: undefined, endDate: undefined };
       } else if (key === 'startDate' || key === 'endDate') {
-        values = { ...settings, dateOffset: undefined };
+        values = { ...values, dateOffset: undefined };
       }
+      values = { ...values, validDateOffset: key === 'dateOffset' && value === 3 };
```
kylgarmor, line 72: is the purpose of `validDateOffset` to determine if the date filter is a range vs an offset?  Seems a bit funny that we have two values `dateOffset` and `validDateOffset` which seem to overlap functionality

CelloLiu: No, it is not for that, in this case it is used for telling if we should use the value 3 or not as in some case that should represent ALL instead, it could be a bit confusing as naming is hard sometimes :)

Fosol: It's very confusing.  The Editor app uses the following values.
0 = today
1 = 24 hours
2 = 48 hours
3 = all
What values is the Subscriber app using?  We should be consistent.

Fosol (changes requested):

app/editor/src/features/admin/filters/FilterSettingsForm.tsx
CelloLiu: In Editor app we use 3 for either ALL or 3 in different places such as the home search and the admin/filters Query, so validDateOffset is here for telling the difference between the two, and we can rename it to whatever that make senses.

Fosol: Standardize the values.
 
Perhaps the following makes more sense
0 = All
1 = Today
2 = 24 hours
3 = 48 hours

Fosol (changes requested):

app/editor/src/features/admin/filters/FilterSettingsForm.tsx
Fosol: Remove this logic on this line.

app/editor/src/features/admin/filters/FilterSettingsForm.tsx
```diff
@@ -65,10 +65,11 @@ export const FilterSettingsForm: React.FC<IFilterSEttingsFormProps> = ({
       var values = { ...settings };
       values[key] = value;
       if (key === 'dateOffset') {
-        values = { ...settings, startDate: undefined, endDate: undefined };
+        values = { ...values, startDate: undefined, endDate: undefined };
```
Fosol, line 68: Keep this line

app/editor/src/features/admin/filters/FilterSettingsForm.tsx
```diff
@@ -65,10 +65,11 @@ export const FilterSettingsForm: React.FC<IFilterSEttingsFormProps> = ({
...
       } else if (key === 'startDate' || key === 'endDate') {
-        values = { ...settings, dateOffset: undefined };
+        values = { ...values, dateOffset: undefined };
```
Fosol, line 70: Keep this line

libs/npm/core/src/utils/elastic/generatePublishedOnQuery.ts
Fosol, line 1: undo changes on this file.

libs/npm/core/src/hooks/api-editor/interfaces/IFilterSettingsModel.ts
Fosol, line 1: Undo changes on this file.

Fosol (changes requested):

app/editor/src/features/admin/filters/FilterSettingsForm.tsx
CelloLiu: Removed, however this rollback has caused a [Search ALL] problem on the home page, which should have now been resolved in this PR as well. Thanks

libs/npm/core/src/utils/elastic/generatePublishedOnQuery.ts
CelloLiu: Removed, however this rollback has caused a [Search ALL] problem on the home page, which should have now been resolved in this PR as well. Thanks

libs/npm/core/src/hooks/api-editor/interfaces/IFilterSettingsModel.ts
CelloLiu: Rollbacked, however this rollback has caused a [Search ALL] problem on the home page, which should have now been resolved in this PR as well. Thanks

app/editor/src/features/content/hooks/useElasticsearch.ts
```diff
@@ -31,7 +31,11 @@ export const useElasticsearch = () => {
         startDate: filter.startDate ? moment(filter.startDate).toISOString() : undefined,
         endDate: filter.endDate ? moment(filter.endDate).toISOString() : undefined,
         dateOffset:
-          filter.startDate || filter.endDate ? undefined : filter.timeFrame ? +filter.timeFrame : 0,
+          filter.startDate || filter.endDate || filter.timeFrame === 3
```
Fosol, line 34: Thanks for this fix!

Fosol (approved):

Fosol (changes requested):
Please update with the latest `tno-core:0.0.356` library I published for you.

Fosol (approved):

PR #1211, opened 2023-11-01
Fix view report instance

Fosol:
Introduced a regression issue on the subscriber site when I changed the route for view a report instance from `preview` to `view`.

### DIFF
--- a/libs/net/kafka/KafkaListener`.cs
+++ b/libs/net/kafka/KafkaListener`.cs
@@ -1,5 +1,5 @@
-using System.Text.Json;
 using System.Collections.Concurrent;
+using System.Text.Json;
 using Confluent.Kafka;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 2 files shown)


libs/net/kafka/KafkaListener`.cs
```diff
@@ -168,7 +168,7 @@ public async Task ConsumeAsync(Func<ConsumeResult<TKey, TValue>, Task> action, C
             var consumeResult = this.Consumer!.Consume(cancellationToken);
             if (consumeResult != null)
             {
-                _logger.LogInformation("Message received from Kafka topic: '{topic}' key:'{key}'", consumeResult.Topic, consumeResult.Message.Key);
+                _logger.LogDebug("Message received from Kafka topic: '{topic}' key:'{key}'", consumeResult.Topic, consumeResult.Message.Key);
```
Fosol, line 171: Less noise on details we don't need for default logging.

PR #1213, opened 2023-11-01
fix: #tno-1590 - made duplicate detection more robust

kylgarmor:
- since we cant delete in non-PROD, we have to do it the long way

### DIFF
--- a/services/net/image/ImageAction.cs
+++ b/services/net/image/ImageAction.cs
@@ -167,14 +167,13 @@ public class ImageAction : IngestAction<ImageOptions>
                             // update the existing refernce published on date with the new date
                             reference.PublishedOn = newReference.PublishedOn;
 
-                            // What about the worst case scenario: one Editor changes it in MMI and another Editor changes it in TNO?
                             Logger.LogDebug($"Image Ingest : UPDATE : [{manager.Ingest.IngestType?.Name ?? "Image"} - {manager.Ingest.Name}] : {path}");
                         } else {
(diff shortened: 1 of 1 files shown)


services/net/image/ImageAction.cs
```diff
@@ -154,18 +156,35 @@ public ImageAction(IApiService api, IOptions<ImageOptions> options, ILogger<Imag
...
+                        Logger.LogDebug($"Image Ingest : ADD NEW : [{manager.Ingest.IngestType?.Name ?? "Image"} - {manager.Ingest.Name}] : {path}");
+                    } else {
+                        // if we already have a contentReference, check if this is actually an update to the image
+                        if (newReference.PublishedOn > reference.PublishedOn)
+                        {
+                            isUpdatedSourceContent = true;
+                            // update the existing refernce published on date with the new date
+                            reference.PublishedOn = newReference.PublishedOn;
+
+                            // What about the worst case scenario: one Editor changes it in MMI and another Editor changes it in TNO?
```
Fosol, line 170: I don't think we care about this scenario.  In TNO they don't even look at front pages in the editor app.

services/net/image/ImageAction.cs
Fosol, line 1: Have you run your linter for nullability?

services/net/image/ImageAction.cs
kylgarmor: that comment was acopy/paste from lesewhere.  not relevant

Fosol (approved):

services/net/image/ImageAction.cs
kylgarmor: Ive fixed the two possible null issues in `ImageAction.cs`.  I installed SonarLint and it's mad about all sorts of stuff...

PR #1219, opened 2023-11-03
TNO-1749: View reports

jdtoombs:
Backend was previously set up, just had to tweak the front end a little bit to make this work

Additionally:
- report rows now have pointer when hovering
- clicking icons in my reports still work as `stopPropegation` has been added to their onClicks

preview-reports

### DIFF
--- a/app/subscriber/src/features/my-reports/view/ReportInstancePreview.tsx
+++ b/app/subscriber/src/features/my-reports/view/ReportInstancePreview.tsx
@@ -1,13 +1,13 @@
 import { SearchWithLogout } from 'components/search-with-logout';
 import React from 'react';
 import { useParams } from 'react-router-dom';
-import { useReportInstances } from 'store/hooks';
+import { useReports } from 'store/hooks';
 import { Col, IReportResultModel, Loading, Show } from 'tno-core';
 
 import * as styled from './styled';
 
 const ReportInstancePreview: React.FC = () => {
-  const [{ viewReportInstance }] = useReportInstances();
+  const [{ previewReport }] = useReports();
   const { id } = useParams();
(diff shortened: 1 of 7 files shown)


app/subscriber/src/features/my-reports/view/ReportInstancePreview.tsx
Fosol, line 1: naming and path is confusing and inconsistent.
A user will "view" an report instance because it's a published report.
A user will "preview" a report because it hasn't been published and they're just viewing a possibility if it was published.
What is the purpose of this component and page?

Fosol (changes requested):

app/subscriber/src/features/my-reports/view/ReportInstancePreview.tsx
jdtoombs: Ahh, ok that makes more sense I think. I was getting confused in the ticket
image
But on the subscriber side it would just be viewing a published report rather than viewing it in an email right?

app/subscriber/src/features/my-reports/view/ViewReportInstance.tsx
```diff
@@ -18,14 +18,14 @@ const ReportInstancePreview: React.FC = () => {
     async (reportId: number) => {
       try {
         setIsLoading(true);
-        const response = await viewReportInstance(reportId);
+        const response = await previewReport(reportId);
```
Fosol, line 21: Doesn't make sense again.  What's the purpose of this component page?  Is it to "view" an instance or "preview" a report?

Fosol (changes requested):

jdtoombs:
Ticket being reworked

PR #1224, opened 2023-11-06
View Report Page

jdtoombs:
- new page to view generated reports on the subscriber site
- a user may view a report if it is their own, or it is made public 

viewreport

### DIFF
--- a/libs/net/dal/Services/ReportService.cs
+++ b/libs/net/dal/Services/ReportService.cs
@@ -105,6 +105,7 @@ public class ReportService : BaseService<Report, int>, IReportService
             .Include(r => r.Owner)
             .Include(r => r.Template).ThenInclude(t => t!.ChartTemplates)
             .Include(r => r.Sections).ThenInclude(s => s.Filter)
+            .Include(r => r.Instances)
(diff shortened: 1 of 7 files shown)


Fosol (approved):
Make sure the endpoint is returning the existing generate report subject/body fields and not recompiling it again.

jdtoombs:
Oops forgot to comment back, looks like it is ok afterall! @Fosol

libs/net/dal/Services/ReportService.cs
```diff
@@ -105,6 +105,7 @@ public IEnumerable<Report> Find(ReportFilter filter)
             .Include(r => r.Owner)
             .Include(r => r.Template).ThenInclude(t => t!.ChartTemplates)
             .Include(r => r.Sections).ThenInclude(s => s.Filter)
+            .Include(r => r.Instances)
```
Fosol, line 108: This needs to be removed.  Think about what this does.  It returns every instance generated for the report.  After a year it's 365 items in an array, of which at best only one will ever be viewed.  I'll remove it from my next PR.

PR #1227, opened 2023-11-07
TNO-1892: Filter media type, add page, all but source filtering

mihailistov:
Add filter media type page

### DIFF
--- a/app/subscriber/src/features/filter-media/FilterMedia.tsx
+++ b/app/subscriber/src/features/filter-media/FilterMedia.tsx
@@ -1,0 +1,113 @@
+import { MsearchMultisearchBody } from '@elastic/elasticsearch/lib/api/types';
+import { DateFilter } from 'components/date-filter';
+import { determineColumns } from 'features/home/constants';
+import { createFilterSettings } from 'features/press-gallery/utils';
+import moment from 'moment';
+import React from 'react';
+import { FiRefreshCcw } from 'react-icons/fi';
+import { useNavigate } from 'react-router-dom';
+import { useContent, useLookup, useLookupOptions } from 'store/hooks';
+import {
+  FieldSize,
+  FlexboxTable,
+  generateQuery,
+  IContentModel,
+  IFilterSettingsModel,
+  IOptionItem,
+  Row,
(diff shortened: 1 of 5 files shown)


jdtoombs (commented):
It looks good, will just have to do some testing after my PR is merged. 

A few things to note:
1. Product Search Group must be set through the editor app. Do this by navigating to Admin > Sources > and change the Product Search Group field and save
2. You can view `useFilterOptions.ts` to see how I have previously separated the sources by their respective product
3. When my PR is merged and you pull in the code you will have to perform a `make db-refresh` as well as a `make refresh n=editor` & `make refresh n=subscriber`. I believe those can all be run from the root directory

kylgarmor:
just curious if there is an issue if a product doesnt have a media type associated with it, what happens?

app/subscriber/src/features/filter-media/FilterMedia.tsx
```diff
@@ -0,0 +1,120 @@
...
+      try {
+        const res: any = await findContentWithElasticsearch(filter, false);
+        setResults(res.hits.hits.map((h: { _source: IContentModel }) => h._source));
+      } catch {}
+    },
+    [findContentWithElasticsearch],
+  );
+
+  React.useEffect(() => {
+    setSettings(
```
Fosol, line 51: Curious if you see the double event when you load the page with this?  You don't need to change anything, but I suspect there is one.

app/subscriber/src/features/filter-media/styled/FilterMedia.tsx
```diff
@@ -0,0 +1,69 @@
...
+      margin-right: auto;
+      input {
+        margin-top: 0.25em;
+      }
+    }
+  }
+  .table {
+    width: 100%;
+    .group {
+      background-color: #f9f9f9 !important;
```
Fosol, line 30: One day it would actually be nice to use themes correctly.  I was hoping someone would clean up the global css colours and such so that we could consistently use them in our styled components.  You don't need to do this here, I'm just highlighting it.

Fosol (approved):

PR #1228, opened 2023-11-07
tno-2026 - fix memory issue service

kylgarmor:
- update confluent.kafka libraya to 2.3.0
- switched KafkaAdmin to singleton in all usages

### DIFF
--- a/libs/net/kafka/KafkaListener`.cs
+++ b/libs/net/kafka/KafkaListener`.cs
@@ -1,4 +1,3 @@
-using System.Collections.Concurrent;
 using System.Text.Json;
 using Confluent.Kafka;
 using Microsoft.Extensions.Logging;
@@ -20,7 +19,7 @@ public class KafkaListener<TKey, TValue> : IKafkaListener<TKey, TValue>, IDispos
     private readonly ILogger _logger;
     private bool _disposed = false;
     private bool _open = false;
-    private readonly ConcurrentDictionary<ConsumeResult<TKey, TValue>, bool> _consumeResults = new();
+    private int _resultConsumerTracker = 0;
     #endregion
 
     #region Properties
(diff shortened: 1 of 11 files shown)


libs/net/kafka/KafkaListener`.cs
```diff
@@ -244,7 +242,7 @@ public void Commit(ConsumeResult<TKey, TValue> result)
         this.Consumer?.Commit(result);
         _logger.LogDebug("Message committed from topic:'{topic}' key:'{key}'", result.Topic, result.Message.Key);
 
-        if (this.IsLongRunningJob) _consumeResults.TryRemove(result, out _);
+        if (this.IsLongRunningJob) Interlocked.Decrement(ref _resultConsumerTracker);
```
Fosol, line 245: Thanks for this.

services/net/content/ContentManager.cs
```diff
@@ -221,7 +225,48 @@ private void ListenerStopHandler(object sender, EventArgs e)
...
+            {
+                this.Listener.Stop();
+                this.State.Stop();
+            }
+            else
+            {
+                await ProcessSourceContentAsync(result);
+
+                // Inform Kafka this message is completed.
+                this.Listener.Commit(result);
```
Fosol, line 241: If we're not pausing do we need to commit here?

kylgarmor: I was thinking that as well, but I left it in place as that's the same patter as the notification service uses.

My guess is that Kafka will ignore the commit because we have told it to autocommit.  If we could determine the config of the Listener at runtime we could wrap that `.Commit(` in an `if` statement.  Because that flag is part of the Kafka config as opposed to service Options, there didn't seem to be an easy way to check...

Fosol (approved):

PR #1229, opened 2023-11-07
TNO-2012: New way to identify product

jdtoombs:
Now in the Sources Admin section a user may select which product they would like to associate with it.

i.e) Vancouver Sun could be tagged as Daily Print by updating the Product Search Group field

The logic changes have been applied so that the advanced search uses this field (in `useFilterOptions`) instead of `productId` which could interfere with the ingest.

image


app/editor/src/features/admin/sources/constants/defaultSource.ts
```diff
@@ -14,6 +14,7 @@ export const defaultSource: ISourceForm = {
   useInTopics: false,
   ownerId: '',
   productId: '',
+  productSearchGroupId: '',
```
kylgarmor, line 17: is it possible that a source might actually be associated with multiple Product types?  I thought I heard Scott or maybe Carolynn mention something like that?

kylgarmor (approved):
Approved with the caveat that we may change this later to change it to a one-to-many relation ship later

PR #1239, opened 2023-11-08
TNO-1938 Persist press gallery filters

fbarreta:
- Persist press gallery filters
- Adapt the initial query to use the persisted filters

### DIFF
--- a/app/subscriber/src/store/slices/content/interfaces/IContentState.ts
+++ b/app/subscriber/src/store/slices/content/interfaces/IContentState.ts
@@ -2,10 +2,12 @@ import {
   IContentListAdvancedFilter,
   IContentListFilter,
 } from 'features/content/list-view/interfaces';
-import { IContentModel, IPaged } from 'tno-core';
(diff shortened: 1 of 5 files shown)


app/subscriber/src/store/slices/content/interfaces/IContentState.ts
```diff
@@ -2,10 +2,12 @@ import {
...
   IContentListFilter,
 } from 'features/content/list-view/interfaces';
-import { IContentModel, IPaged } from 'tno-core';
+import { IContentModel, IOptionItem, IPaged } from 'tno-core';
 
 export interface IContentState {
   filter: IContentListFilter;
   filterAdvanced: IContentListAdvancedFilter;
   content?: IPaged<IContentModel>;
+  dateFilter?: IOptionItem | null;
```
Fosol, line 11: Change to a single object for the `pressGalleryFilter` so that it's more intuitive when maintaining the store.

Fosol (changes requested):

PR #1242, opened 2023-11-09
TNO-1876: Fix image uploads in editor

mihailistov:
- Fixed the image aspect ratio

Also noticed other bugs such as:
- Upon selecting a file and uploading, no image preview shown, this is fixed now
- Upon deleting a file, the file name would remain, and the upload component would remain in a weird state that might be confusing to users

Before:
Nov-08-2023 22-52-09

After:
Nov-09-2023 12-09-14


app/editor/src/features/content/form/components/upload/Upload.tsx
```diff
@@ -57,26 +57,44 @@ export const Upload: React.FC<IUploadProps> = ({
...
   ...rest
 }) => {
+  const [loading, setLoading] = React.useState(true);
   const { isShowing, toggle } = useModal();
   const { values, setFieldValue } = useFormikContext<IContentForm>();
   const [, { ffmpeg }] = useWorkOrders();
   const { isShowing: showFFmpegModal, toggle: toggleFFmpeg } = useModal();
 
-  const [file, setFile] = React.useState<IFile>();
+  const [file, setFile] = React.useState<IFile | undefined>();
```
Fosol, line 66: Not required as typescript will know it can be undefined in this case because it's not set.  Same with below `useState()` calls.

app/editor/src/features/content/form/components/upload/Upload.tsx
```diff
@@ -57,26 +57,44 @@ export const Upload: React.FC<IUploadProps> = ({
...
 
   React.useEffect(() => {
-    if (!!initFile) {
-      setFile(undefined);
+    if (!!initFile && loading) {
+      setFile(initFile);
+      setFileName(generateName(initFile));
+      setLoading(false);
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Fosol, line 82: Always explain why you are not including all possible dependencies.  According to 'people' this means we're doing something wrong.  I however can't seem to figure out how to not cause infinite loops in some cases.

Fosol (approved):
Does this also work when navigating to the next content item with the Next/Previous button?

app/editor/src/features/content/form/components/upload/Upload.tsx
mihailistov: In all my React dev experience this situation is unavoidable... I will try to see if there's a better way going forward. For now I'll put a note.

Fosol (approved):

Fosol (approved):

PR #1251, opened 2023-11-13
TNO-1989: Fix issue with Editor papers default filter/query

mihailistov:


app/editor/src/store/slices/content/contentSlice.ts
```diff
@@ -33,7 +32,26 @@ export const initialContentState: IContentState = {
     logicalOperator: LogicalOperator.Contains,
     searchTerm: '',
   },
-  filterPaper: defaultPaperFilter(),
```
mihailistov, line 36: this doesn't actually work, redux initial state doesn't play nice with logic like this, plus best practice is to start it in an empty state, and load what we need

not sure if default paper filters is something that'll ever change.. if it's not then we can consider hard-coding the default state

Fosol: I think this used to be a constant.

Fosol (approved):

PR #1253, opened 2023-11-15
TNO-1774 Update ingest services

Fosol:
All Ingest Services will now update their last run time each iteration.  Also they will maintain an updated list of the current ingest configuration.  This should resolve issues where the ingest was not getting the latest updates when their configuration was changed.

Moved and renamed a few files for clarity.


libs/net/services/Actions/IIngestAction`.cs
```diff
@@ -20,6 +20,6 @@ public interface IIngestAction<TOptions> : IServiceAction<TOptions>
     /// <param name="name"></param>
     /// <param name="data"></param>
     /// <param name="cancellationToken"></param>
-    public Task<ServiceActionResult> PerformActionAsync<T>(IIngestServiceActionManager manager, string? name = null, T? data = null, CancellationToken cancellationToken = default) where T : class;
+    public Task<ServiceActionResult> PerformActionAsync<T>(IIngestActionManager manager, string? name = null, T? data = null, CancellationToken cancellationToken = default) where T : class;
```
Fosol, line 23: Renamed for clarity

libs/net/services/Actions/Managers/IIngestActionManager.cs
```diff
@@ -3,36 +3,42 @@
...
-    public Dictionary<string, object> Values { get; }
+    Dictionary<string, object> Values { get; }
     #endregion
 
     #region Methods
+    /// <summary>
+    /// Make AJAX request and update the ingest.
+    /// </summary>
+    /// <returns></returns>
+    Task<IngestModel> UpdateIngestStateAsync(int failedAttempts = 0);
```
Fosol, line 27: Provides a way to update the status from the service manager.

libs/net/services/Actions/Managers/IngestActionManager`.cs
```diff
@@ -65,14 +65,11 @@ public class IngestActionManager<TOptions> : ServiceActionManager<TOptions>, IIn
     /// Verify that the schedule should run the action.
     /// </summary>
     /// <returns></returns>
-    protected override async Task<bool> PreRunAsync()
+    protected override Task<bool> PreRunAsync()
     {
-        // Always fetch the latest version of the ingest and update this manager with it.
-        var ingest = await this.Api.HandleRequestFailure(async () => await this.Api.GetIngestAsync(this.Ingest.Id), this.Options.ReuseIngests, this.Ingest);
```
Fosol, line 71: We should already have the latest ingest configuration, there is no need to make this request too.

PR #1255, opened 2023-11-16
TNO-1965 Fix Blacks Ingest

Fosol:
When Blacks Press sends us corrupted XML files the File Monitor Service fails to be able to ingest all the stories in the file.  This update should fix the most common issue that occurs.

### DIFF
--- a/services/net/filemonitor/FilemonitorAction.cs
+++ b/services/net/filemonitor/FilemonitorAction.cs
@@ -24,7 +24,9 @@ namespace TNO.Services.FileMonitor;
 /// </summary>
 public class FileMonitorAction : IngestAction<FileMonitorOptions>
 {
-    #region Properties
+    #region Variables
+    // Regex to find the story closing tag.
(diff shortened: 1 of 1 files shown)


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -678,6 +680,16 @@ private static string GetItemSourceCode(IngestModel ingest, string paperName, Di
             var xmlTxt = ReadFileContents(filePath, ingest);
             if (xmlTxt != null)
             {
+                // Files regularly have the 'last' story corrupting in the <story> node.
+                // This results in the whole file being invalid.
+                // Check for and fix the common failure.
+                if (!_storyTag.Match(xmlTxt).Success)
+                {
+                    xmlTxt = $"{xmlTxt}</story><date>{DateTime.Now:MM-dd-yyyy}</date></bcng>";
```
Fosol, line 688: Append commonly missing data.
This does however mean we'll be missing content in this story and possibly any other stories that should be below this story.

PR #1260, opened 2023-11-16
TNO-1965 Fix File Monitor Service

Fosol:
Minor change to remove invalid `try+catch` and extra update of the ingest status.

### DIFF
--- a/services/net/filemonitor/FilemonitorAction.cs
+++ b/services/net/filemonitor/FilemonitorAction.cs
@@ -470,6 +470,8 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
                         await ImportArticleAsync(manager, item);
                     }
                 }
+                // This ingest has just completed running for one content item.
+                await manager.UpdateIngestStateAsync(manager.Ingest.FailedAttempts);
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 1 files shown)


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -680,42 +678,31 @@ private static string GetItemSourceCode(IngestModel ingest, string paperName, Di
     /// <returns></returns>
     private XmlDocument? GetValidXmlDocument(string filePath, IngestModel ingest)
     {
-        try
```
Fosol, line 683: Don't need this extra `try+catch` as the caller already handles errors.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -470,6 +470,8 @@ private async Task GetXmlArticlesAsync(string dir, IIngestActionManager manager,
                         await ImportArticleAsync(manager, item);
                     }
                 }
+                // This ingest has just completed running for one content item.
+                await manager.UpdateIngestStateAsync(manager.Ingest.FailedAttempts);
```
Fosol, line 474: Move to only update status on success, as the error handler already updates the status.

PR #1263, opened 2023-11-17
TNO-2050 Fix tooltip

Fosol:
Fixing issue that was causing the tooltips to expand.

I could not find what was causing it, but my guess is something is bubbling down to the tooltip to set the height at 100%.

image


app/editor/src/App.tsx
```diff
@@ -50,14 +50,8 @@ function App() {
...
       <ToastContainer />
-      <Tooltip style={{ zIndex: '999' }} variant="info" id="main-tooltip" place="top" float />
-      <Tooltip
-        style={{ zIndex: '999' }}
-        variant="info"
-        id="main-tooltip-right"
-        place="right"
-        float
-      />
+      <Tooltip variant="info" id="main-tooltip" place="top" float />
```
Fosol, line 53: Moved styling to css.

app/editor/src/index.scss
```diff
@@ -183,6 +183,19 @@ h4 {
   }
 }
 
+// *************************************************
+// Tooltip
+// *************************************************
+#main-tooltip {
+  height: unset !important;
```
Fosol, line 190: I hate `!important` but I can't find what's causing the expand...

PR #1264, opened 2023-11-17
TNO-2058 Add SignalR

Fosol:
Adding SignalR to the Subscriber app.


app/subscriber/src/components/layout/DefaultLayout.tsx
```diff
@@ -26,6 +27,7 @@ export const DefaultLayout: React.FC<ILayoutProps> = ({ children, ...rest }) =>
   const keycloak = useKeycloakWrapper();
   const { setToken } = React.useContext(SummonContext);
   useToastError();
+  useApiHub();
```
Fosol, line 30: This will initiate the connection.

app/subscriber/src/store/hooks/signalr/useApiHub.ts
Fosol, line 1: I don't like have a duplicate of this.  It may make sense to move this to tno-core.

PR #1268, opened 2023-11-21
Tno 1609 export report data to excel

fbarreta:
Export report to excel spreadsheet.
Need tno-core npm publish.


api/net/Areas/Subscriber/Controllers/ReportController.cs
```diff
@@ -354,6 +355,28 @@ public async Task<IActionResult> Generate(int id, [FromQuery] bool regenerate =
...
+    /// <summary>
+    /// Generate an Excel document report.
+    /// </summary>
+    /// <param name="id"></param>
+    /// <returns></returns>
+    [HttpPost("{id}/export")]
+    [Produces("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")]
+    [ProducesResponseType((int)HttpStatusCode.OK)]
+    [SwaggerOperation(Tags = new[] { "Report" })]
+    public FileResult Generate(int id)
```
Fosol, line 367: This method will causing confusion due to naming and location.  It doesn't generate a report, it generates an Excel document.  We have other methods named Generate which generate the report.  Also the `id` is not for a report, instead it is for an instance of a report.  Move this whole method to the `ReportInstanceController` as it's generating an Excel document based on the report instance.

- Rename to `GenerateExcel`
- Move to `ReportInstanceController`

api/net/Areas/Subscriber/Controllers/ReportController.cs
```diff
@@ -354,6 +355,28 @@ public async Task<IActionResult> Generate(int id, [FromQuery] bool regenerate =
...
+    {
+        var helper = new ReportXlsExport(_reportInstanceService, _serializerOptions);
+
+        var report = helper.GenerateReport(id);
+
+        using var stream = new MemoryStream();
+        report.Write(stream);
+        var bytes = stream.ToArray();
+
+        return File(bytes, "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", "cbra.xlsx");
```
Fosol, line 377: Rename file `cbra.xlsx` to `{report.Name}.xlsx`.  Handle invalid characters.

libs/net/reports/ReportXlsExport.cs
```diff
@@ -0,0 +1,543 @@
...
+                return workbook;
+            }
+
+            var content =  serialized.Content.ToList().OrderBy(o => o.Content?.PublishedOn);
+
+            var tonesCount = new TonesCount();
+
+            content.
+            Select(s => new {
+                s.Content?.Product?.Name,
```
Fosol, line 109: Change `Product` to `MediaType`

libs/npm/core/src/hooks/api-subscriber/useApiSubscriberReports.ts
```diff
@@ -85,6 +87,13 @@ export const useApiSubscriberReports = (
         }`,
       );
     },
+    exportReport: (reportId: number) => {
+      return download({
+        url: `/subscriber/reports/${reportId}/export`,
+        method: 'post',
+        fileName: `myreport${reportId}.xlsx`,
```
Fosol, line 94: Use the name of the file provided by the endpoint if possible.  It should be `{report.name}.xlsx`.  See above comment.

libs/net/reports/ReportXlsExport.cs
```diff
@@ -0,0 +1,543 @@
...
+  }
+  #endregion
+
+  #region Methods
+        /// <summary>
+        /// Creates a new instance of a Report object.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns>XSSFWorkbook</returns>
+        public XSSFWorkbook GenerateReport(int id)
```
Fosol, line 93: This is a misleading method.  It doesn't generate a report, it generates an Excel document.  Also the `id` argument is not a `Report.Id`, it is a `ReportInstance.Id`.  We have other methods with the same name that actually generate a report.  We don't want to cause confusion.

Rename the method to `GenerateExcel`.  
Rename the argument to `instanceId`.

Fosol (changes requested):

app/subscriber/src/features/my-reports/view/ReportSnapshotEdit.tsx
```diff
@@ -33,6 +33,18 @@ export const ReportSnapshotEdit: React.FC = () => {
     [generateReport, setValues],
   );
 
+  const handleExport = React.useCallback(async () => {
+    try {
+      if (instance?.id) {
+        await toast.promise(exportReport(instance?.id), {
+          pending: 'Downloading file',
```
jdtoombs, line 40: Might be overkill but if we have some sort of status update message for other downloads in the application we could make an enum for this

api/net/Areas/Subscriber/Controllers/ReportInstanceController.cs
```diff
@@ -209,6 +210,28 @@ public async Task<IActionResult> SendToAsync(int id, string to)
         return new JsonResult(new ReportInstanceModel(instance));
     }
 
+    /// <summary>
+    /// Generate an Excel document report.
+    /// </summary>
+    /// <param name="instanceId"></param>
+    /// <returns></returns>
+    [HttpPost("{instanceId}/export")]
```
kylgarmor, line 218: not a big deal, but wondering whether semantically speaking should this be a GET rather than a POST?

fbarreta: done ! :)

Fosol (approved):

PR #1274, opened 2023-11-24
TNO-1818 Order Sections by Attribute

fbarreta:
Added:

- TNO Core List of attributes
- Subscriber Drop Down Selector
- Editor Drop Down Selector
- New attribute on Section Settings model
- Order by this new attributes on Report Generation


api/net/Areas/Helpers/ReportHelper.cs
```diff
@@ -64,6 +64,29 @@ public class ReportHelper : IReportHelper
...
 
     #region Methods
+    /// <summary>
+    /// Order the content based on the session field.
+    /// </summary>
+    /// <param name="content"></param>
+    /// <param name="orderByField"></param>
+    /// <returns>Ordered Content</returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    public static ContentModel[] OrderBySectionField(ContentModel[] content, string orderByField)
```
Fosol, line 74: You will need to add this logic to the `ReportingManager.cs` otherwise when the report is generated and sent by email it won't sort by the field.  If you can come up with a clean way to centralize this logic so that it isn't duplicated it would be create.  Currently the `ReportHelper` has a bit of duplication that will cause maintenance issues going forward.

Fosol (changes requested):

Fosol (approved):

PR #1275, opened 2023-11-25
TNO-2075: Goodbye nest, and each feature to have its own store

jdtoombs:
That was more work than expected..

- Ripping out models that were associated with NEST
- Changing all `findContent()` to `findContentWithElasticSearch`
- Add filter store for each feature to maintain state independent of each other


jdtoombs:
@Fosol - just need to publish but this is ready for review

app/editor/yarn.lock
```diff
@@ -10604,7 +10604,7 @@ __metadata:
     sass-extract-js: 0.4.0
     sass-extract-loader: 1.1.0
     styled-components: 5.3.9
-    tno-core: 0.0.371
+    tno-core: 0.0.372
```
Fosol, line 10607: Your lock file is pointing to a different version than the package.

app/subscriber/yarn.lock
```diff
@@ -10629,7 +10629,7 @@ __metadata:
     sass-extract-js: 0.4.0
     sass-extract-loader: 1.1.0
     styled-components: 5.3.9
-    tno-core: 0.0.371
+    tno-core: 0.0.372
```
Fosol, line 10632: same issue here

Fosol (changes requested):

Fosol (approved):

PR #1278, opened 2023-11-28
TNO-1598: Remove footer, add info shield

jdtoombs:
Quick PR to remove footer as per Bobbi request, and add info shield in the sidebar.

info-shield


app/subscriber/src/components/info/InfoShield.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Tooltip } from 'react-tooltip';
+import { Col } from 'tno-core';
+
+import * as styled from './styled';
+
+export const InfoShield: React.FC = () => {
+  return (
+    <styled.InfoShield className="info-shield">
+      <FaShield data-tooltip-id="info-shield" />
+      <Tooltip clickable className="react-tooltip" id="info-shield" place="top" delayHide={800}>
```
jdtoombs, line 12: add a few ms delay with `delayHide` so it doesn't disappear if you accidentally mouse off while moving to a link.

Fosol (approved):

PR #1287, opened 2023-11-30
feat: #TNO-1937 - allow admins to action product requests

kylgarmor:
- TNO-CORE updated to 0.0.375
- admins now get an email when a user requests to subscribe or unsubscribe from a product
- no email is generated if a user cancels their request

**Editor app - Product screen with subscription request shown**
image

**Subscriber screen with various products in various states shown below:**
image


libs/net/entities/AdminConfigurableSettingNames.cs
```diff
@@ -0,0 +1,15 @@
+using TNO.Core.Json;
+
+namespace TNO.Entities;
+
+/// <summary>
+/// Provides report status.
+/// </summary>
+public enum AdminConfigurableSettingNames
```
Fosol, line 8: Confused how this is used?

kylgarmor:
It’s used the same way as a string const would be. I just use a tostring to get the value to look up.
On Nov 29, 2023 at 4:40 PM -0800, Jeremy Foster ***@***.***>, wrote:

@Fosol commented on this pull request.

________________________________

In libs/net/entities/AdminConfigurableSettingNames.cs<https://github.com/bcgov/tno/pull/1287#discussion_r1410025757>:

> @@ -0,0 +1,15 @@
+using TNO.Core.Json;
+
+namespace TNO.Entities;
+
+/// <summary>
+/// Provides report status.
+/// </summary>
+public enum AdminConfigurableSettingNames


Confused how this is used?

libs/net/entities/AdminConfigurableSettingNames.cs
kylgarmor: It’s used the same way as a string const would be. I just use a tostring to get the value to look up.

kylgarmor: Just noticed the summary is wrong from copy pasta

Fosol (approved):

PR #1292, opened 2023-11-30
TNO-2061: Autofill Evening Overview summary

mihailistov:
Screenshot 2023-11-30 at 12 39 46 AM


Fosol (approved):

jdtoombs (approved):

app/editor/src/features/reports/av-overview/styled/AutoComplete.tsx
```diff
@@ -0,0 +1,50 @@
+import styled, { css } from 'styled-components';
```
jdtoombs, line 1: @Fosol - I find this very easy to read, wonder if we should start implementing this way? (Rather than one root styled component for each component)

PR #1294, opened 2023-11-30
fix: #TNO-937 - fixes for subscriber site

kylgarmor:
- subscriber who had never been subscribed before couldn't request subscription
- unapproved requests for subscription were not showing up
- fixed cursor issue on the shared section component

### DIFF
--- a/app/editor/src/features/admin/products/constants/subscriberColumns.tsx
+++ b/app/editor/src/features/admin/products/constants/subscriberColumns.tsx
@@ -91,14 +91,10 @@ export const subscriberColumns = (
           )}
           disabled={true}
           title={
-            product.subscribers.some(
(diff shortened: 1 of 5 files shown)


app/editor/src/features/admin/products/constants/subscriberColumns.tsx
```diff
@@ -91,10 +91,14 @@ export const subscriberColumns = (
           )}
           disabled={true}
           title={
-            product.subscribers.filter((u) => u.id === cell.original.id)[0]
```
Fosol, line 94: Why going through the array twice?  Can't you just do `product.subscribers.filter((u) => u.id === cell.original.id)?.requestedIsSubscribedStatus ? 'Request is to SUBSCRIBE' : 'Request is to UNSUBSCRIBE'`?

Fosol (approved):

PR #1296, opened 2023-12-01
Folder fixes and style enhancements 

jdtoombs:
In this PR:

- strip repetitive css and use existing component made
- fix bug where you cannot rename folder
- apply common table theme to my folders
- allow user to hit enter when renaming a folder to confirm save
- change icons
- made a simple util for `handleEnterPress()` to be used with inputs throughout the app on their `onKeyDown` prop

folder-bugs-enhance

### DIFF
--- a/app/subscriber/src/features/utils/handleEnterPressed.ts
+++ b/app/subscriber/src/features/utils/handleEnterPressed.ts
@@ -1,0 +1,10 @@
+import { KeyboardEvent } from 'react';
+
+export const handleEnterPressed = (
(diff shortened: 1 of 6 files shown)


app/subscriber/src/features/utils/handleEnterPressed.ts
```diff
@@ -0,0 +1,10 @@
+import { KeyboardEvent } from 'react';
```
jdtoombs, line 1: Will try to replace repetitive code with this util when I com across it

Fosol (approved):

PR #1298, opened 2023-12-01
Fix editor app

Fosol:
Not certain how this hasn't broken then application in the past, but this should fix the issue.


app/editor/src/features/home/InfoText.tsx
```diff
@@ -29,7 +29,9 @@ export const InfoText: React.FC = () => {
         <li>Articles related to major stories.</li>
       </ul>
       <Show visible={systemMessage.isEnabled}>
-        <p className="system-message">{parse(systemMessage.message) ?? ''}</p>
+        <p className="system-message">
+          {systemMessage.message ? parse(systemMessage.message) : ''}
```
Fosol, line 33: This was causing the whole site to crash.

PR #1300, opened 2023-12-02
fix: #tno-2090 - radio content content type broken

kylgarmor:
- apparently the library we used to determine contentype type is wrong
- an m4a file should map to audio/mp4 NOT audio/m4a
- switched to new library
- updated all references
- removed unused references
- small fix for content migrator to fix issue with null settings


Fosol (approved):

Fosol:
Thanks for solving this mystery

PR #1309, opened 2023-12-07
[UPDATED] TNO-1894, 1782: Add content to folders, new components, new styling

jdtoombs:
Updated images...

In this PR:

- now able to add content to report (need to verify I am doing it correctly...)
- revamp view content bar
- revamp add to folder
- add to report actions and ui

image
image
image
image


app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const [activeReport, setActiveReport] = React.useState<IReportModel | null>(null);
+  React.useEffect(() => {
+    if (!myReports.length) {
+      findMyReports().catch(() => {});
+    }
+    // Only do this on init.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  /** Adds the content to the active report. */
```
jdtoombs, line 26: I feel like this isn't the correct way... let me know. I essentially look for the latest instance and add the content to it with the existing content that is there

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const addContentToReport = React.useCallback(
+    (section: string) => {
+      if (!activeReport || !activeReport.instances.length) {
+        return;
+      }
+
+      // convert the content to an instance content
+      const convertedContent: IReportInstanceContentModel = {
```
jdtoombs, line 34: Will extract

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
jdtoombs: Updating now to handle case where no instances are present

app/subscriber/src/components/bar/styled/Bar.tsx
```diff
@@ -1,16 +1,18 @@
 import styled from 'styled-components';
 
-export const Bar = styled.div`
+export const Bar = styled.div<{ $vanilla?: boolean }>`
```
Fosol, line 3: What's with the use of the `$`?

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+      const convertedContent = toInstanceContent(content, activeReport, section);
+      const update = (report: IReportModel) => {
+        updateReport(report, true)
+          .then(() => toast.success(`${content.length} storie(s) have been added to report.`))
+          .catch(() => {});
+      };
+
+      if (!!activeReport.instances.length) {
+        // get the latest instance content and append to it
+        const instContent = [
```
Fosol, line 44: The current instance will always be `0` index position.  We return them in descending order always

Fosol: We never want to add content to a sent report instance.  A new instance will need to be generated in this case.
`ReportInstance.SentOn == null` to allow updating content.
We should add an error in the API if anyone attempts to update a sent instance.

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -0,0 +1,114 @@
...
+          }
+        });
+        const report = {
+          ...activeReport,
+          instances: instances,
+        };
+        updateReport(report, true).catch(() => {});
+      }
+
+      if (!activeReport.instances.length) {
```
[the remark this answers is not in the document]
Fosol: Don't come up with a new way to do this.  The My Reports already has this logic all worked out.  When you edit a report it will automatically generate an instance if you don't have one already.

image

Refer to ReportEdit.tsx

image

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+          instances: [newInstance],
+        };
+        update(report);
+      }
+    },
+    [activeReport, content, updateReport],
+  );
+
+  // ensure no concurrency errorrs rather than getting from profile store
+  React.useEffect(() => {
```
Fosol, line 81: Change this logic to match above example.

app/subscriber/src/components/tool-bar/utils/newReportInstance.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  IReportInstanceContentModel,
+  IReportInstanceModel,
+  IReportModel,
+  ReportStatusName,
+} from 'tno-core';
+
+export const newReportInstance = (
```
Fosol, line 8: No need for this

app/subscriber/src/components/tool-bar/utils/toInstanceContent.ts
```diff
@@ -0,0 +1,25 @@
+import { IContentModel, IReportInstanceContentModel, IReportModel } from 'tno-core';
+
+/**
+ * Converts an array of content to an array of report instance content.
+ * @param content The content to convert.
+ * @param report The report to convert the content for.
+ * @param sectionName The section name to convert the content for.
+ */
+export const toInstanceContent = (
```
Fosol, line 9: I would simplify.  You should already know the instance at this point.

Use these arguments `instanceId, sectionName, content, sortOrder`

app/subscriber/src/components/tool-bar/utils/toInstanceContent.ts
```diff
@@ -0,0 +1,25 @@
...
+  report: IReportModel,
+  sectionName: string,
+): IReportInstanceContentModel[] => {
+  return content.map((c) => {
+    return {
+      contentId: c.id,
+      content: c,
+      sectionName: sectionName,
+      instanceId: !!report.instances.length ? report.instances[report.instances.length - 1].id : 1,
+      createdBy: c.createdBy,
```
Fosol, line 20: You don't need `createdBy` or `createdOn`

app/subscriber/src/features/home/home-filters/styled/HomeFilters.tsx
```diff
@@ -14,4 +15,17 @@ export const HomeFilters = styled(Row)`
     font-size: 0.5em !important;
   }
   font-family: Roboto', sans-serif;
+      button {
+      min-width: 5rem;
+      display: flex;
+      justify-content: center;
+      &.active {
+        background-color: ${(props) => props.theme.css.defaultRed} !important;
```
Fosol, line 23: `!important` means we're doing something wrong most of the time

Fosol (changes requested):

app/subscriber/src/components/bar/styled/Bar.tsx
jdtoombs: Just from the styled-components docs, seems to be their convention, not a big deal to remove though

app/subscriber/src/features/home/home-filters/styled/HomeFilters.tsx
jdtoombs: this is just moved code from another area, the next ticket on the list screen will remove this anyways (removed already on current branch)

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
jdtoombs: Gotcha, ok will update and based on below comments as well!

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -27,12 +27,17 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content }) =>
...
       if (!activeReport) {
         return;
       }
 
-      const convertedContent = toInstanceContent(content, activeReport, section);
+      const convertedContent = toInstanceContent(
+        content,
+        activeReport.instances[0].id,
+        sectionName,
+        0,
```
jdtoombs, line 39: @Fosol, updated again. But quick question, how do we determine the sort order from here? Should it always go to the top/bottom?

jdtoombs: Does it make sense to have it as an argument here when it will apply to `1...n` pieces of content being passed?

app/subscriber/src/features/content/view-content/styled/ViewContent.tsx
```diff
@@ -2,6 +2,10 @@ import styled from 'styled-components';
 
 export const ViewContent = styled.div`
   .info-bar {
+    margin-left: -1.25em;
```
jdtoombs, line 5: negative margins so bg color fills full page section (ignoring padding)

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
Fosol: Sort order would default to the top (`sortOrder = 0`), but Scott has mentioned he wants a way to control this.  Perhaps focus on getting this to work before we provide a way for the user to choose the sort order value.

Fosol (approved):

PR #1310, opened 2023-12-07
TNO-2037 Add mobile report

Fosol:
Reports now support the ability to send two types of emails to subscribers.  The Full Text version of the report, or the Link Only version of the report.  When adding subscribers to reports the choice can be made for each user.

## Summary

- Add DB migration 1.0.98
- Published tno-core:0.0.?
- Update Editor app
- Updated Report Admin Subscribers
- Fixed bug introduced with the email image fix.  Had to temporarily remove images from reports.  Will create a separate PR to put them back in.
- Fixed bug in Openshift configuration so that the Reporting Service can communicate with the Charts API.

## Report Admin Subscribers

image


.vscode/settings.json
```diff
@@ -24,8 +24,8 @@
   ],
   "java.configuration.updateBuildConfiguration": "interactive",
   "editor.codeActionsOnSave": {
-    "source.organizeImports": false,
-    "source.fixAll.eslint": true
+    "source.organizeImports": "never",
```
Fosol, line 27: vscode did this automatically.  I suspect an update.

app/editor/src/features/admin/av-overviews/OverviewSubscribers.tsx
```diff
@@ -21,7 +27,7 @@ export const OverviewSubscribers: React.FC = () => {
   }, []);
 
   const handlePageChange = React.useCallback(
-    async (page: ITablePage, table: ITableInternal<IUserModel>) => {
+    async (page: ITablePage, table: ITableInternal<IUserAVOverviewModel>) => {
```
Fosol, line 30: Fix subscriber model

app/editor/src/features/admin/reports/ReportFormSections.tsx
```diff
@@ -120,11 +120,6 @@ export const ReportFormSections = () => {
               Control the output of the report. Redirect users to view the report on the website
               instead of in their email.
             </p>
-            <FormikCheckbox
```
Fosol, line 123: No longer control it with this configuration setting

app/subscriber/src/features/my-reports/view/ReportEdit.tsx
```diff
@@ -138,10 +138,16 @@ export const ReportEdit: React.FC = () => {
                 </Col>
                 <Col gap="0.5rem">
                   <Row gap="1rem" justifyContent="flex-end">
+                    <Action
```
Fosol, line 141: We can now get to the config page easier

libs/net/ches/ChesService.cs
```diff
@@ -237,7 +237,8 @@ public async Task<EmailResponseModel> SendEmailAsync(IEmail email)
             email.Cc = email.Cc?.NotNullOrWhiteSpace();
             email.Bcc = email.Bcc?.NotNullOrWhiteSpace();
 
-            ConvertInlineImagesToAttachments(email);
+            // TODO: Need to fix mail merge before this can be implemented again.
+            // ConvertInlineImagesToAttachments(email);
```
Fosol, line 241: @kylgarmor removed temporarily

libs/net/ches/Models/IEmail.cs
```diff
@@ -1,12 +1,17 @@
-using TNO.Core.Converters;
 using System;
 using System.Collections.Generic;
 using System.Text.Json.Serialization;
+using TNO.Core.Converters;
 
 namespace TNO.Ches.Models
 {
-    public interface IEmail: IEmailBase
+    public interface IEmail
```
Fosol, line 8: Rolled back to fix reports

libs/net/ches/Models/IEmailMerge.cs
```diff
@@ -1,15 +1,60 @@
...
+using System.Text.Json.Serialization;
+using TNO.Core.Converters;
 
 namespace TNO.Ches.Models
 {
     /// <summary>
     /// IEmailMerge interface, provides a structure to manage generating multiple emails with a single template.
     /// </summary>
-    public interface IEmailMerge: IEmailBase
+    public interface IEmailMerge
```
Fosol, line 10: Rolled back to fix reports

libs/net/dal/Services/FolderService.cs
```diff
@@ -31,11 +31,7 @@ public IEnumerable<Folder> FindAll()
             .Include(f => f.Owner)
             .Include(f => f.Filter)
             .Include(f => f.Events).ThenInclude(f => f.Schedule)
-            .Include(f => f.ContentManyToMany).ThenInclude(f => f.Content)
-            .Include(f => f.ContentManyToMany).ThenInclude(f => f.Content).ThenInclude(c => c!.Source)
-            .Include(f => f.ContentManyToMany).ThenInclude(f => f.Content).ThenInclude(c => c!.MediaType)
-            .Include(f => f.ContentManyToMany).ThenInclude(f => f.Content).ThenInclude(c => c!.Series)
-            .Include(f => f.ContentManyToMany).ThenInclude(f => f.Content).ThenInclude(c => c!.TonePools)
+            .Include(f => f.ContentManyToMany)
```
Fosol, line 34: This may break the subscriber folders.

libs/npm/core/src/components/table/styled/FlexboxTable.tsx
```diff
@@ -65,8 +65,13 @@ export const FlexboxTable = styled.div<ITableStyleProps>`
         background-color: rgb(233, 236, 239);
       }
 
-      &:hover:not(.active) {
-        filter: brightness(75%);
+      &:nth-child(even):hover:not(.active) {
```
Fosol, line 68: Fixed a hover bug

services/net/reporting/appsettings.json
```diff
@@ -23,6 +23,11 @@
     "ViewContentUrl": "https://mmi.gov.bc.ca/view/",
     "RequestTranscriptUrl": "https://mmi.gov.bc.ca/api/subscriber/work/orders/transcribe/"
   },
+  "Charts": {
```
Fosol, line 26: Fix bug in Openshift.

libs/npm/core/src/hooks/api/interfaces/IUserModel.ts
```diff
@@ -31,8 +31,4 @@ export interface IUserModel extends IAuditColumnsModel {
   reports?: IReportModel[];
   reportInstances?: IReportInstanceModel[];
   notifications?: INotificationModel[];
-  // This model is often used to identify the user is subscribed to a report or notification.
```
kylgarmor, line 34: nice!

libs/npm/core/src/hooks/api/interfaces/IUserProductModel.ts
```diff
@@ -0,0 +1,10 @@
+import { IUserModel } from '.';
+
+export interface IUserProductModel extends IUserModel {
+  /** Whether the user is subscribed to this product. */
+  isSubscribed: boolean;
+  /** Whether the user has requested to be subscribed. */
+  requestedIsSubscribedStatus?: boolean; // TODO: I find this status confusing
```
kylgarmor, line 7: naming things is hard - I'm open to suggestions
- `requestedIsSubscribedStatus` : Does the user want to subscribe or unsubscribe - Scott apparently wants to know either way.  This value in combination with the `isSubscribed` can be used so that we can infer "They are subscribed but they want to unsubscribe"
- `subscriptionChangeActioned` : Has an admin actioned this request.  Used to show/infer the status of the subscription on both the subscriber and admin page

services/net/reporting/ReportingManager.cs
```diff
@@ -351,9 +354,19 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
             }
         }
 
-        var to = report.Subscribers.Where(s => !String.IsNullOrWhiteSpace(s.User?.Email) && s.IsSubscribed).Select(s => s.User!.Email).ToArray();
-        var subject = await this.ReportEngine.GenerateReportSubjectAsync(report, sectionContent, false);
-        var body = await this.ReportEngine.GenerateReportBodyAsync(report, sectionContent, null, false);
+        // Generate and send report to subscribers who want an email with a link to the website.
+        // We do this first because we don't want to save the output of this in the instance.
+        var linkOnlyFormatSubscribers = report.Subscribers.Where(s => s.IsSubscribed && LinkOnlyFormats.Contains(s.Format) && !String.IsNullOrWhiteSpace(s.User?.Email)).ToArray();
+        var linkOnlyFormatBody = linkOnlyFormatSubscribers.Any() ? await this.ReportEngine.GenerateReportBodyAsync(report, sectionContent, null, true, false) : "";
```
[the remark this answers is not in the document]
kylgarmor: running them in a specific order to avoid a problem seems a little fragile to me.  Would it be better to have a mechanism for creating the reportbody without saving the instance at all?

services/net/reporting/ReportingManager.cs
```diff
@@ -366,44 +379,57 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
...
             ?? throw new InvalidOperationException("Report instance failed to be returned by API")) : null;
 
-        if (request.SendToSubscribers && (to.Any() || !String.IsNullOrEmpty(request.To)))
+        if (request.SendToSubscribers)
         {
-            try
-            {
-                // Send the email.
-                var response = await SendEmailAsync(request, to, subject, body, $"{report.Name}-{report.Id}");
+            var failed = false; // I don't like this implementation because one group of emails could be successful, but the other fail.
```
kylgarmor, line 390: you could implement a mechanism to set a "reasonable" delay on the first email,  capture the id of the first email request and if you encounter a failure sending the second email, you call ches to cancel the first one

services/net/reporting/ReportingManager.cs
```diff
@@ -366,44 +379,57 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
...
+                }
+                catch (ChesException ex)
                 {
-                    model.Id = instanceModel.Id;
-                    model.Version = instanceModel.Version ?? 0;
-                    model.Status = ReportStatus.Failed;
-                    model.Content.ForEach(ric => ric.InstanceId = model.Id);
-                    model.Response = JsonDocument.Parse(JsonSerializer.Serialize(ex.Data["error"], _serializationOptions));
-                    await this.Api.UpdateReportInstanceAsync(model);
+                    responseModel.FullTextFormatResponse = JsonDocument.Parse(JsonSerializer.Serialize(ex.Data["error"], _serializationOptions));
```
kylgarmor, line 417: I would put a comment here to denote what's happening.  My guess is that you are storing something to do with the error in the report instance for later analysis

services/net/reporting/ReportingManager.cs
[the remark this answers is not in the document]
kylgarmor: or should different format mean different/seperate instance for each?

kylgarmor (commented):
added a few comments - no show stoppers

services/net/reporting/ReportingManager.cs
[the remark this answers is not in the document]
Fosol: I was going back and forth on this and trying to think of other possibilities.  I'm not a fan of the two separate email groups as part of a single operation, but it seemed the best option in this case.

- We only want a single instance of a report when it is published
- When we generate an instance, we want all reports to reference it so that when a user goes to the website and views it, it'll be already generated and identical each time
- The link only email is really the same instance of the report and must point to the instance with content

services/net/reporting/ReportingManager.cs
Fosol: Yes we have always been capturing the CHES response in our report and notification instances.  I had to modify this to support capturing two responses is all.

services/net/reporting/ReportingManager.cs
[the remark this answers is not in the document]
Fosol: Originally I designed it so that it was a different report or instance to handle the different format.  This however didn't meet the business requirements.

PR #1312, opened 2023-12-08
Frontpage Gallery

fbarreta:
I created a component that is shared now between the My Frontpage session and the Frontpage that accepts the an IContentModel[].

image

### DIFF
--- a/app/subscriber/src/components/front-page-gallery/FrontPageGallery.tsx
+++ b/app/subscriber/src/components/front-page-gallery/FrontPageGallery.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { useContent, useNavigateAndScroll } from 'store/hooks';
+import { IContentModel, Row } from 'tno-core';
+
+import * as styled from './styled';
+
+interface IFrontPageGallery extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * Content array.
+   */
+  frontpages: IContentModel[];
+}
+
+export const FrontPageGallery: React.FC<IFrontPageGallery> = ({ frontpages }) => {
+  const [srcUrls, setSrcUrls] = React.useState<any[]>([]);
+  const navigateAndScroll = useNavigateAndScroll();
+  const [, { stream }] = useContent();
+
+  React.useEffect(() => {
+    setSrcUrls([]);
+    if (!!frontpages) {
+      frontpages.forEach((x) => {
+        if (x.fileReferences?.length) {
+          stream(x.fileReferences[0].path).then((result) => {
+            setSrcUrls((srcUrls) => [...srcUrls, { url: result, id: x.id }]);
+          });
+        }
+      });
+    }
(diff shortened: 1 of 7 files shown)


app/subscriber/src/components/front-page-gallery/FrontPageGallery.tsx
```diff
@@ -0,0 +1,47 @@
...
+  frontpages: IContentModel[];
+}
+
+export const FrontPageGallery: React.FC<IFrontPageGallery> = ({ frontpages }) => {
+  const [srcUrls, setSrcUrls] = React.useState<any[]>([]);
+  const navigateAndScroll = useNavigateAndScroll();
+  const [, { stream }] = useContent();
+
+  React.useEffect(() => {
+    setSrcUrls([]);
```
Fosol, line 20: Why do you do this?  You've already set the initial value on line 15.  It will also result in the state getting rendered twice any time `frontpages, stream` dependencies change.

fbarreta: Because it's keeps adding when I change the date filter:
image

app/subscriber/src/components/front-page-gallery/FrontPageGallery.tsx
```diff
@@ -0,0 +1,47 @@
...
+    if (!!frontpages) {
+      frontpages.forEach((x) => {
+        if (x.fileReferences?.length) {
+          stream(x.fileReferences[0].path).then((result) => {
+            setSrcUrls((srcUrls) => [...srcUrls, { url: result, id: x.id }]);
```
Fosol, line 25: This line is why it's adding.  Rewrite this part instead of causing a double render.

Fosol (changes requested):

app/subscriber/src/components/front-page-gallery/FrontPageGallery.tsx
Fosol: Try one of the following.

```
  React.useEffect(() => {
    const srcUrls = frontPages
      .filter((fp) => fp.fileReferences?.length)
      .map(async (fp) => {
        const result = await stream(fp.fileReferences![0].path);
        return { url: result, id: fp.id };
      });
    setSrcUrls(srcUrls);
  }, []);

  React.useEffect(() => {
    const srcUrls = Promise.all(
      frontPages
        .filter((fp) => fp.fileReferences?.length)
        .map(async (fp) => {
          const result = await stream(fp.fileReferences![0].path);
          return { url: result, id: fp.id };
        }),
    );
    srcUrls.then((results) => setSrcUrls(results));
  }, []);
  ```

Fosol: A link with more information about awaiting a map
https://advancedweb.hu/how-to-use-async-functions-with-array-map-in-javascript/

Fosol (approved):

PR #1315, opened 2023-12-13
Select all button issues and responsiveness

jdtoombs:
In this PR:

- make content tool bar mobile friendly
- allow user to select all 

I am not a huge fan of this solution, I was trying desperately to get it to trigger the `onChange` event within the flexbox table but no luck. Had to add some additional logic.

But, this has been something I have been staring at for a while - so please let me know if you see a better way.

image

The quick filter buttons could use some love, wondering if we could come up with some mobile icons to use:
image

### DIFF
--- a/app/subscriber/src/components/basic-search/styled/BasicSearch.tsx
+++ b/app/subscriber/src/components/basic-search/styled/BasicSearch.tsx
@@ -55,6 +55,7 @@ export const BasicSearch = styled(Row)`
   /* MOBILE SEARCH  */
   .search-mobile {
     display: flex;
+    max-width: 13em;
     margin-top: auto;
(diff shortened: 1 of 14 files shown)


kylgarmor:
maybe a silly question, but would a "Select all" for a table with row check boxes be in the column heading above the row check boxes?

jdtoombs:
> maybe a silly question, but would a "Select all" for a table with row check boxes be in the column heading above the row check boxes?

@Fosol - If I could do this it would be a lot cleaner - but let me know what you think! For things like the search page I will have to do something slightly diff though.

Fosol:
Look at the latest PR I pushed on this.

PR #1318, opened 2023-12-13
TNO-2008 Share Stories

fbarreta:
ShareMenu Component was created to send notifications with selected content to the selected colleague.
image
Tooltip with email list.

image
Confirmation modal.

Some changes were made on Modal component to accommodate more UI components, and also control to the confirm button was added (to block the button in case all the required fields were not selected).
Along these some changes to hooks will require tno-core update.

A new endpoint/model were created on Subscriber area to handle the notification send.


Fosol:
We'll definitely need to clean up the UI to get closer to the mocks provided by Bobbi

api/net/Areas/Subscriber/Controllers/NotificationController.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        var username = User.GetUsername() ?? throw new NotAuthorizedException("Username is missing");
+        var user = _userService.FindByUsername(username) ?? throw new NotAuthorizedException("User does not exist");
+
+        var request = new NotificationRequestModel(NotificationDestination.NotificationService, new { })
+        {
+            NotificationId = notification.Id,
+            ContentId = contentId,
+            RequestorId = user.Id,
+            To = to,
```
Fosol, line 87: `to` will be provided by the user returned by `userId`.

api/net/Areas/Subscriber/Models/NotificationModel.cs
Fosol, line 1: Models belong to a controller.  Review how all other controllers are implemented.  Move to folder associated with controller.

api/net/Areas/Subscriber/Models/NotificationModel.cs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// <summary>
+    /// get/set - The foreign key to the template.
+    /// </summary>
+    public int TemplateId { get; set; }
+
+    /// <summary>
+    /// get/set - The Razor templates to generate the notification.
+    /// </summary>
+    public NotificationTemplateModel? Template { get; set; }
```
Fosol, line 27: This model doesn't need to return the template.  Only include what is needed.  Also review the properties below whether they are needed.  If the web app isn't going to use it, not sense serializing it.

api/net/Areas/Subscriber/Models/NotificationModel.cs
```diff
@@ -0,0 +1,131 @@
...
+    #endregion
+
+    #region Methods
+    /// <summary>
+    /// Creates a new instance of a Notification object.
+    /// </summary>
+    /// <param name="options"></param>
+    /// <param name="includeTemplate"></param>
+    /// <returns></returns>
+    public Entities.Notification ToEntity(JsonSerializerOptions options, bool includeTemplate = false)
```
Fosol, line 97: Are you ever going to send this model back to the API?  I doubt it.  Why include this method?

api/net/Areas/Subscriber/Models/NotificationTemplateModel.cs
Fosol, line 1: Is this even needed?

app/subscriber/src/components/share-menu/ShareMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+   * The content that has been selected to add to a folder
+   */
+  content: IContentModel[];
+}
+
+/** Component that renders the button that gives users access to a sub menu that will allow them to add selected content to
+ * an existing folder. Or create a new one.
+ */
+export const ShareMenu: React.FC<IShareSubMenuProps> = ({ content }) => {
+  const [options, setOptions] = React.useState<IOptionItem[]>([]);
```
Fosol, line 28: Organization convention is as follows.

Hook dependencies first.
React.useState second.
Component variables third.
React.useEffect and functions last.

app/subscriber/src/components/share-menu/ShareMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+        toast.error(`${Settings.DefaultAlert} setting needs to be configured.`);
+      }
+    } catch {}
+  };
+
+  const message = `Share ${content.length} selected content${
+    content.length > 1 ? 's' : ''
+  } with colleague ?`;
+
+  const selectColleague = (
```
Fosol, line 68: Should this be using a `React.useMemo`?

app/subscriber/src/store/hooks/subscriber/useColleagues.ts
```diff
@@ -32,6 +42,12 @@ export const useColleagues = (): [IColleagueController] => {
         );
         return response.data;
       },
+      sendNotification: async (notificationId: number, to: string, contentId?: number) => {
```
Fosol, line 45: Rename to `share` and update to match the endpoint changes above.

libs/npm/core/src/hooks/api/subscriber/useApiSubscriberColleagues.ts
```diff
@@ -36,5 +36,12 @@ export const useApiSubscriberColleagues = (
         { data: model },
       );
     },
+    sendNotification: (notificationId: number, to: string, contentId?: number) => {
```
Fosol, line 39: Update according to the changes above.

Fosol (changes requested):

fbarreta:
> We'll definitely need to clean up the UI to get closer to the mocks provided by Bobbi

I'll try an hybrid approach between tooltip and the modal to confirm. I'll update it and let you know.

api/net/Areas/Subscriber/Controllers/NotificationController.cs
fbarreta: I don't think that's the case, because it will be sent to a colleague, if I set it to the user object, it will be sent to it's own email.

kylgarmor: I'm pretty sure that Jeremy is referring to the new `userId` parameter he has suggested on the route.  On this topic though you did mention sharing to multiple colleagues during standup. So that might end up being an array of `colleague userId` ???

fbarreta: The property **to** from the NotificationRequestModel requires a string with the email address. Since the emails will be sent individually, no arrays are required.
image
Looks like the notification object is correct.
I just want to double (triple) check it ...

Fosol: The `NotificationRequestModel` requires a string.
The endpoint you are building is to share a story with a colleague.  A colleague must be an existing user in the solution.  When a user wants to share a story, they must provide the `userId` of the user they want to share with.  You can then find the `email` for the colleague.

fbarreta: Ok ! Got it !

api/net/Areas/Subscriber/Controllers/ContentController.cs
```diff
@@ -281,5 +293,39 @@ public IActionResult Delete(ContentModel model)
...
+    /// <param name="contentId"></param>
+    /// <param name="colleagueId"></param>
+    /// <param name="notificationId"></param>
+    /// <returns></returns>
+    [HttpPost("{contentId}/share")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(NotificationModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Notification" })]
+    public async Task<IActionResult> ShareAsync(long? contentId, int colleagueId, int notificationId)
```
Fosol, line 309: `contentId` should never be null

Fosol (changes requested):

Fosol:
Resolve all the branch conflicts and I'll publish a new version of `tno-core` for you.

Fosol:
@fbarreta update the PR with `tno-core:0.0.387`.  Update both the editor and subscriber apps.

fbarreta:
> @fbarreta update the PR with `tno-core:0.0.387`. Update both the editor and subscriber apps.

Done !

Fosol (approved):

PR #1324, opened 2023-12-14
TNO-2045 Add tab validation

Fosol:
We have a new component FormikTab that supports validation of each tab.  It also identifies which tab has issues.

## Configure Report

image

### DIFF
--- a/app/subscriber/src/components/tabs/FormikTabs.tsx
+++ b/app/subscriber/src/components/tabs/FormikTabs.tsx
@@ -1,0 +1,87 @@
+import { FormikErrors, useFormikContext } from 'formik';
+import { set, transform } from 'lodash';
+import React from 'react';
+import { ValidationError } from 'yup';
+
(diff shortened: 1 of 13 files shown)


app/subscriber/src/components/tabs/FormikTabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+      });
+      Promise.all(results).then((tabs) => {
+        tabs.forEach((tab) => {
+          setTouched({ ...touched, ...(tab.touched ?? {}) });
+          setStatus({ ...errors, ...tab.errors });
+        });
+        setTabs(tabs);
+      });
+    }
+    // Only validate when submitting.
```
Fosol, line 82: I don't like hooks and useEffects...  I had to do this to detect when the parent form was being submitted so that I can validate each tab.  If you include any other dependencies it loops.

PR #1335, opened 2023-12-18
TNO-2111 Fix report schedule

Fosol:
Schedule validation was failing because the deserialization was including the `0` value in the enum array.  Fixed the API, and also updated the UI to not pass the value to the API.

## Summary

- Published `tno-core:0.0.389`
- Fixed report schedules


app/editor/src/features/admin/folders/FolderFormCollect.tsx
```diff
@@ -161,36 +162,43 @@ export const FolderFormCollect: React.FC = () => {
                 label="Monday"
                 name={`events.0.runOnWeekDays`}
                 value={ScheduleWeekDayName.Monday}
+                onBeforeChange={(value) => selectWeekDays(value)}
```
Fosol, line 165: Added callback to clean up the selected values and remove `NA` values when there are other values selected.

libs/net/core/Extensions/EnumExtensions.cs
```diff
@@ -25,18 +25,23 @@ public static class EnumExtensions
...
         where T : Enum
     {
-        return Enum.GetValues(typeof(T))
+        var values = Enum.GetValues(typeof(T))
             .Cast<T>()
             .Where(v => value.HasFlag(v))
             .Select(v => (int)(object)v)
             .ToArray();
+
+        if (values.Length > 1 && !includeZero) return values.Where((v) => v != 0).ToArray();
```
Fosol, line 43: We don't want to include a `0` enum value if the array has other values.

PR #1336, opened 2023-12-19
TNO-2045 Improve reports

Fosol:
Provided a way to quickly change the sort order and the section for content in a report.

## Summary

- Updated Subscriber app
- Published tno-core:0.0.390

## Edit Report

image


app/subscriber/src/features/my-reports/utils/sortContent.ts
```diff
@@ -19,7 +21,7 @@ export const sortContent = (
       index === 0 || rows[index - 1].sectionName !== i.sectionName ? 0 : sectionIndex + 1;
     return {
       ...i,
-      sortOrder: sectionIndex,
+      sortOrder: updateSortOrder ? sectionIndex : i.sortOrder,
```
Fosol, line 24: We only need to resort when we regenerate the content

app/subscriber/src/features/my-reports/view/components/ReportContentSection.tsx
```diff
@@ -59,6 +73,33 @@ export const ReportContentSection: React.FC<IReportContentSectionProps> = ({
     [instance, setFieldValue],
   );
 
+  const handleChangeSection = React.useCallback(
```
Fosol, line 76: We can now change the section with a select component instead of drag and drop

app/subscriber/src/features/my-reports/view/components/ReportContentSection.tsx
```diff
@@ -59,6 +73,33 @@ export const ReportContentSection: React.FC<IReportContentSectionProps> = ({
...
+            : c,
+        ),
+        (c) => `${c.contentId}-${c.sectionName}`,
+      );
+      setFieldValue('instances.0.content', sortReportContent(values, content));
+    },
+    [setFieldValue, values],
+  );
+
+  const handleChangeSortOrder = React.useCallback(
```
Fosol, line 93: We can now change the sort with a text component instead of drag and drop

PR #1338, opened 2023-12-19
TNO-2069 Fit image to div in view content

fbarreta:
image

Scale image content according to div.


kylgarmor (approved):
Demonstrated as working

PR #1348, opened 2023-12-22
Tno 1920 hide innactive tags

fbarreta:
A script was added to set the tags with the name:  '-- UPDATE MIGRATED TAG NAME --' to disabled.

ReactSelect component requires isDisabled property, not isEnabled on IOptionItem.

image
Old content with disabled tag.

image
New content with disabled tag.


api/net/Areas/Editor/Controllers/LookupController.cs
```diff
@@ -162,7 +162,7 @@ public async Task<IActionResult> FindAllAsync()
         var series = _seriesService.FindAll();
         var contributors = _contributorService.FindAll();
         var metrics = _metricService.FindAll();
-        var tagServices = _tagService.FindAll();
+        var tagServices = _tagService.FindAllEnabled();
```
Fosol, line 165: Don't do this, as it will cause issues when content references a disabled tag.

api/net/Areas/Subscriber/Controllers/LookupController.cs
```diff
@@ -128,7 +128,7 @@ public IActionResult FindAllAsync()
         var license = _licenseService.FindAll();
         var series = _seriesService.FindAll();
         var ministers = _ministerService.FindAll();
-        var tagServices = _tagService.FindAll();
+        var tagServices = _tagService.FindAllEnabled();
```
Fosol, line 131: Don't do this, as it will cause issues when content references a disabled tag.  I would think that we still want to be able to search for content with these old tags.

Fosol (commented):
Please only hide them on the special display for tags.

Fosol (changes requested):
.

libs/npm/core/src/components/form/options/interfaces/OptionItem.ts
```diff
@@ -6,21 +6,21 @@ export class OptionItem<T extends string | number | undefined = string | number
...
     this.value = value;
-    this.isEnabled = isEnabled;
+    this.isDisabled = isEnabled;
   }
 
   static create<T extends string | number>(
     label: string | React.ReactElement,
     value: T,
-    isEnabled: boolean = true,
+    isDisabled: boolean = true,
```
Fosol, line 21: I think I prefer your implementation `isDisabled` because html options use the `disabled` flag.  However, we'll need to review any other places this logic may need to be fixed.  Currently the Editor app only shows enabled values in many of the select components.

app/editor/src/features/content/form/components/tags/Tags.tsx
```diff
@@ -31,9 +31,11 @@ export const Tags: React.FC<ITagsProps> = () => {
     return {
       label: tag.code,
       value: tag.id,
+      isDisabled: !tag.isEnabled,
     } as IOptionItem;
   });
-
+  // eslint-disable-next-line no-console
+  console.log(tagOptions);
```
Fosol, line 38: Remove `console.log`

Fosol (approved):

PR #1350, opened 2023-12-23
CSS Things

jdtoombs:
- Fixed numerous scrolling and alignment issues
- remove double header, now placed at grid level instead of entering on every page
- remove `SearchWithLogout` component as it is no longer used, `Header` is used instead


app/subscriber/src/features/my-reports/styled/MyReports.tsx
```diff
@@ -28,6 +28,14 @@ export const MyReports = styled.div`
     }
   }
 
+  .page-section {
+    margin-top: 0;
+  }
+
+  .my-reports-content {
+    height: calc(100dvh - 260px);
```
jdtoombs, line 36: `dvh` is handy for mobile, as it accounts for potential UI from things like the Safari app showing the URL bar at the top of the screen

Fosol (approved):
Happy to see the header cleaned up

PR #1363, opened 2023-12-29
TNO-1881 Update report template to format time on header

fbarreta:
tno-1881-formated
Formatted time


libs/net/dal/Migrations/1.0.106/Up/PostUp/00-WeekdayTemplateUpdate.sql
```diff
@@ -0,0 +1,125 @@
+DO $$
+BEGIN
+
+UPDATE public.report_template SET
```
Fosol, line 4: What did you update, and wouldn't you also need to update the weekend template?

fbarreta: I updated this:
@section.Name @section.StartTime hr
to this:
@section.Name @(section.StartTime?.Length == 8 ? section.StartTime.Substring(0, 5) : section.StartTime) @(section.StartTime?.Length > 0 ? "hr" : "")

My local weekend template was outdated, but I've update with dev's and implemented the same changes.

Fosol (approved):

PR #1370, opened 2024-01-03
TNO-2000 Set source sort order

fbarreta:
Add the source sortOrder param when calling the home page.

image

### DIFF
--- a/app/subscriber/src/features/home/Home.tsx
+++ b/app/subscriber/src/features/home/Home.tsx
@@ -57,11 +57,14 @@ export const Home: React.FC = () => {
   const fetchResults = React.useCallback(
     async (filter: MsearchMultisearchBody) => {
       try {
+        if (sortBy === 'source') {
+          filter.sort = [{ 'source.sortOrder': 'asc' }, { publishedOn: 'desc' }];
+        }
         const res: any = await findContentWithElasticsearch(filter, false);
         setContent(res.hits.hits.map((h: { _source: IContentModel }) => h._source));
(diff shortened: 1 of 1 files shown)


Fosol (approved):

app/subscriber/src/features/home/Home.tsx
```diff
@@ -57,11 +57,14 @@ export const Home: React.FC = () => {
...
       try {
+        if (sortBy === 'source') {
+          filter.sort = [{ 'source.sortOrder': 'asc' }, { publishedOn: 'desc' }];
+        }
         const res: any = await findContentWithElasticsearch(filter, false);
         setContent(res.hits.hits.map((h: { _source: IContentModel }) => h._source));
       } catch {}
     },
-    [findContentWithElasticsearch],
+    [findContentWithElasticsearch, sortBy],
```
Fosol, line 67: I don't think we want to send the `source.sortOrder` value to Elasticsearch.  It will skew the search results importance.  Can you instead just sort the results of the search by the `source.sortOrder`?

Fosol (changes requested):

app/subscriber/src/features/home/Home.tsx
fbarreta: I don't like the idea of port processing information on the frontend, if we can get it already processed at the backend.
This update orders the result, but I would still prefer to have it already ordered by a DB engine then a javascript sort function.

Fosol (approved):

PR #1374, opened 2024-01-04
Tno 2144 Parse Editor Tags

fbarreta:
https://github.com/bcgov/tno/assets/26801490/dabf7852-a74f-45ba-918c-50d91dd06493

All the tags between brackets and separated by comma or space will be parsed and added to the tags component.
(With the changes, it will not check the middle of content, just the end, so the video is outdated)

### DIFF
--- a/app/editor/src/features/content/form/ContentStoryForm.tsx
+++ b/app/editor/src/features/content/form/ContentStoryForm.tsx
@@ -21,6 +21,7 @@ import { isSummaryRequired } from './utils';
 export interface IContentStoryFormProps {
   contentType: ContentTypeName;
   summaryRequired?: boolean;
+  checkTags?: Function;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


app/editor/src/features/content/form/ContentStoryForm.tsx
```diff
@@ -41,6 +43,26 @@ export const ContentStoryForm: React.FC<IContentStoryFormProps> = ({
     initSummaryRequired ?? isSummaryRequired(values),
   );
 
+  const getTags = (e: string) => {
+    if (!checkTags) {
+      return;
+    }
+    const regex = new RegExp(/\[(.*?)\]/g);
```
Fosol, line 50: Is there any value to only parse the end of string (trim for empty spaces)?  The client will never add tags to the middle of a document.  This would also remove the risk of some random data coming in from 3rd parties.

fbarreta: It would make my life easier ... hahaha
Probably there is a way, I'll just google it and make it work.
:)

fbarreta: Found it ! And works good.

Fosol (approved):

PR #1379, opened 2024-01-05
TNO-2044 Tags major fixes

fbarreta:
- Remove on blur property from editor that was stripping the tags
- Set field value on change. It was holding the validation while inserting new medias
- Enabled case insensitive tags


Fosol (changes requested):
I have published `tno-core:0.1.5` for you.  Update PR and we'll merge.  Thanks for solving this riddle.  Looks like I moved the component into our core library and we lost the history on who wrote this extraction.

Fosol (approved):

PR #1380, opened 2024-01-05
Fix basic search

Fosol:
Resolved regression bug introduced into the basic search.

My Saves Searches will now load the full page results instead of the Advanced Search.  Clicking the GO ADVANCED will load the component to edit the search.  I've added the gear icon back which will go directly to the advanced search.

image

### DIFF
--- a/app/subscriber/src/components/basic-search/BasicSearch.tsx
+++ b/app/subscriber/src/components/basic-search/BasicSearch.tsx
@@ -1,14 +1,20 @@
 import { filterFormat } from 'features/search-page/utils';
-import { handleEnterPressed } from 'features/utils';
+import { handleEnterPressed, isNumber } from 'features/utils';
 import { FaPlay, FaSearch } from 'react-icons/fa';
 import { useNavigate } from 'react-router';
+import { useParams } from 'react-router-dom';
 import { useContent } from 'store/hooks';
-import { Button, Row, Text, toQueryString } from 'tno-core';
+import { Button, IFilterSettingsModel, Row, Text, toQueryString } from 'tno-core';
 
 import * as styled from './styled';
 
+export interface IBasicSearchProps {
+  onSearch?: (filter: IFilterSettingsModel) => void;
+}
+
 /** Basic search functionality (just search term), and an option to get to the advanced filter */
-export const BasicSearch: React.FC = () => {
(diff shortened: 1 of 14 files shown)


app/subscriber/src/components/basic-search/BasicSearch.tsx
```diff
@@ -52,7 +61,7 @@ export const BasicSearch: React.FC = () => {
         Search
         <FaPlay />
       </Button>
-      <p onClick={() => navigate('/search/advanced')}>GO ADVANCED</p>
+      <p onClick={() => navigate(`/search/advanced/${filterId}`)}>GO ADVANCED</p>
```
Fosol, line 64: This provides a way to load the active filter.

app/subscriber/src/features/my-searches/hooks/useElastic.ts
```diff
@@ -10,18 +11,19 @@ import { generateMustNotQuery, generateQuery, IFilterSettingsModel, Settings } f
  */
 export const useElastic = () => {
   const [{ isReady, settings }] = useLookup();
-
-  const [frontPageImagesMediaTypeId, setFrontPageImagesMediaTypeId] = React.useState(0);
+  const [{ frontPageImagesMediaTypeId }, { storeSettingsFrontPageImagesMediaTypeId }] =
```
Fosol, line 14: Moved logic to redux to reduce the visibility of how horrible React state is.  Now only the very first call will have the issue.  Before it would be every time this hook was used the first time.

app/subscriber/src/features/search-page/SearchPage.tsx
```diff
@@ -63,39 +62,6 @@ export const SearchPage: React.FC<ISearchType> = ({ showAdvanced }) => {
     }
   }, [activeFilter, getFilter, filterId, init, storeFilter, storeSearchFilter]);
 
-  // function that bolds the searched text only if advanced filter is enabled for it
-  const formatSearch = React.useCallback(
```
Fosol, line 67: Moved to separate function file

app/subscriber/src/features/utils/isNumber.ts
```diff
@@ -0,0 +1,7 @@
+export function isNumber(str: string | undefined | null) {
```
Fosol, line 1: Should be moved to `tno-core`

PR #1381, opened 2024-01-05
fix: #tno-2151 - fixed reset settings

kylgarmor:
- added visual hint for advanced search settings when section is collapsed
- added tooltips to icons in collapsible section
- fixed check box labels to be clickable for "search in"
- altered layout for advanced search to remove excess margins
- removed a bunch of uneccesary css
- In the screen grab below, Sentiment and Media Types both have search criteria set, as hinted at by the indicator (which has a tooltip)
image

### DIFF
--- a/app/subscriber/src/features/search-page/components/advanced-search/constants/DefaultAdvancedSearch.ts
+++ b/app/subscriber/src/features/search-page/components/advanced-search/constants/DefaultAdvancedSearch.ts
@@ -13,7 +13,7 @@ export const defaultAdvancedSearch = {
   page: '',
   publishedEndOn: '',
   publishedStartOn: '',
-  // search: '', // don't reset this...
+  search: '',
   section: '',
   sentiment: [],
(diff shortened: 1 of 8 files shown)


app/subscriber/src/features/search-page/components/advanced-search/constants/DefaultAdvancedSearch.ts
```diff
@@ -13,8 +13,9 @@ export const defaultAdvancedSearch = {
   page: '',
   publishedEndOn: '',
   publishedStartOn: '',
-  search: '',
+  // search: '', // don't reset this...
```
Fosol, line 16: Won't this cause a undefined error on initial load?  Why not just override this when reseting?

kylgarmor: changed as per suggestion

Fosol (approved):

PR #1385, opened 2024-01-06
Dev Tool: Create styled-component 

jdtoombs:
This is a simple shell script to create a `styled-component` with our standard practices.

When creating a `styled-component` it can be quite tedious making a `/styled` directory followed by the component and an `index.ts` file in both the styled directory, and working directory.

With the following script a user may run `create-styled-component.sh <component-name>`.

The above command will do the following

1. create styled folder in working directory if none exists already
2. add `<component-name>.tsx` and export statement to `/styled/index.ts` (will create `index.ts`  if none exists, or append)
3. create <component-name>.tsx and add basic imports to use styled component
4. add export statement to working directory `index.ts` if it exists

Running `create-styled-component.sh Component` on an empty directory would create the following:
```
├── Component.tsx
└── styled
    ├── Component.tsx
    └── index.ts

1 directory, 3 files
```

*Note: I created a symbolic link in order to be able to just call the script without the full path name*

This was done with: 
`sudo ln -s /home/jdtoombs/code/work/tno/tools/scripts/create-styled-component.sh /usr/local/bin/create-styled-component.sh`

I believe this is possible on Windows as well with:
`winln -s target link` in GitBash

I have not tested this on Windows yet, so some tweaking may be required to instructions!

### DIFF
--- a/tools/scripts/create-styled-component.sh
+++ b/tools/scripts/create-styled-component.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+# Create a styled component with the given name, this script will also create the styled folder if none exists already as well as the
+# index.ts file if none exists already. If they do exist, it will append the export statement to the index.ts file.
+# Usage: create-styled-component.sh <component-name>
+# Example: create-styled-component.sh Button
+# Create a symbolic link to this file in your /usr/local/bin directory to use it from anywhere.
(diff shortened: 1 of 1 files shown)


Fosol (approved):
Nice utility function

PR #1388, opened 2024-01-09
TNO-2113 Share to an email

fbarreta:
Share content with an email that is not in colleague list.


app/subscriber/package.json
```diff
@@ -53,7 +53,7 @@
     "react-tooltip": "5.10.0",
     "redux-logger": "3.0.6",
     "styled-components": "5.3.9",
-    "tno-core": "0.1.5"
+    "tno-core": "./tno-core-0.1.5.tgz"
```
mihailistov, line 56: i don't think we should commit this change 😄

app/subscriber/yarn.lock
```diff
@@ -10568,7 +10568,7 @@ __metadata:
     sass-extract-js: 0.4.0
     sass-extract-loader: 1.1.0
     styled-components: 5.3.9
-    tno-core: 0.1.5
+    tno-core: ./tno-core-0.1.5.tgz
```
mihailistov, line 10571: here too

Fosol (changes requested):
Regrettably we'll need to get Bobbi to provide you some mocks for this work.  We cannot allow users to enter freeform email address to share content.  They can only share stories with subscribers.

Presently for testing we allow reports to be sent to an email address, but this is only visible to administrators.

PR #1392, opened 2024-01-09
TNO-2162 Fix reports

Fosol:
When a report is published (sent to subscribers) the instance status will be changed to submitted.

## Summary

- Updated Editor app
- Updated Subscriber app
- Published `tno-core:0.1.6`


api/net/Areas/Subscriber/Controllers/ReportInstanceController.cs
```diff
@@ -243,6 +243,9 @@ public async Task<IActionResult> PublishAsync(int id)
         var instance = _reportInstanceService.FindById(id) ?? throw new NoContentException("Report does not exist");
         if (instance.OwnerId != user.Id) throw new NotAuthorizedException("Not authorized to publish this report"); // User does not own the report
 
+        instance.Status = Entities.ReportStatus.Submitted;
```
Fosol, line 246: Change report instance state to submitted immediately so that the UI can lock down.

app/subscriber/src/features/my-reports/view/components/ReportSendForm.tsx
```diff
@@ -34,11 +34,17 @@ export const ReportSendForm: React.FC = () => {
   const handlePublish = React.useCallback(
     async (id: number) => {
       try {
-        await publishReportInstance(id);
+        const updatedInstance = await publishReportInstance(id);
+        setFieldValue(
+          'instances',
+          values.instances.map((i) =>
```
Fosol, line 40: Keep the content as it wouldn't have been changed.

PR #1395, opened 2024-01-09
TNO-1948: Fix select checkbox issue

mihailistov:
Jan-09-2024 12-59-41

### DIFF
--- a/app/editor/src/features/content/list-view/components/tool-bar/filter/InputOption.tsx
+++ b/app/editor/src/features/content/list-view/components/tool-bar/filter/InputOption.tsx
@@ -40,11 +40,9 @@ export const InputOption: React.FC<any> = ({
     style,
   };
 
-  /** ensures that the element is unchecked in the menu when the item is removed from the selection */
+  /** ensures that the element is checked/unchecked in the menu when the item is added/removed from the selection */
   React.useEffect(() => {
-    if (!isSelected) {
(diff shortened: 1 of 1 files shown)


app/editor/src/features/content/list-view/components/tool-bar/filter/InputOption.tsx
```diff
@@ -40,11 +40,9 @@ export const InputOption: React.FC<any> = ({
     style,
   };
 
-  /** ensures that the element is unchecked in the menu when the item is removed from the selection */
+  /** ensures that the element is checked/unchecked in the menu when the item is added/removed from the selection */
   React.useEffect(() => {
-    if (!isSelected) {
-      (document.getElementById(children) as HTMLInputElement).checked = false;
-    }
+    (document.getElementById(children) as HTMLInputElement).checked = isSelected;
```
Fosol, line 45: using the DOM isn't ideal.  I wonder why this was implemented this way originally?

Fosol (approved):

app/editor/src/features/content/list-view/components/tool-bar/filter/InputOption.tsx
mihailistov: @Fosol good call, I didn't really question it. I just pushed some changes that is an alternative, more minimalist way. I tested as best I could and don't see any weird side effects. Let me know if you want me to revert it back to my initial commit.

PR #1397, opened 2024-01-10
TNO-2159: Possible fix for my reports not always showing

mihailistov:


app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -18,9 +18,7 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content }) =>
   const [activeReport, setActiveReport] = React.useState<IReportModel>();
   const [reportId, setReportId] = React.useState<number | null>(null);
   React.useEffect(() => {
-    if (!myReports.length) {
-      findMyReports().catch(() => {});
-    }
+    findMyReports().catch(() => {});
```
Fosol, line 21: We don't want to do this.  It's an expensive operation.
Something is updating the state and removing all the other reports.  We need to find what is causing that.

Fosol (changes requested):

PR #1400, opened 2024-01-11
fix - #TNO-2200 - changed mechanism for content to be assigned a score

kylgarmor:
- moved score calculation to API layer
- scores are not set in the UI layer anymore
- added a new optional field to IFolderContentModel to return the calculated max score for use in the UI
- added several missing rules
- updated tno-core to 0.1.9


api/net/Areas/Admin/Controllers/FolderController.cs
```diff
@@ -81,16 +87,26 @@ public IActionResult FindById(int id)
...
     /// <param name="id"></param>
+    /// <param name="includeMaxTopicScore">return the max topic score for each piece of content in the folder</param>
     /// <returns></returns>
     [HttpGet("{id}/content")]
     [Produces(MediaTypeNames.Application.Json)]
     [ProducesResponseType(typeof(IEnumerable<FolderContentModel>), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Folder" })]
-    public IActionResult GetContent(int id)
+    public IActionResult GetContent(int id, bool includeMaxTopicScore = false)
```
Fosol, line 97: I wonder if we even need it here, but it doesn't hurt to have it.

api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -206,6 +214,22 @@ public async Task<IActionResult> AddAsync(ContentModel model)
         var newContent = (Content)model;
         newContent.OwnerId = user.Id;
         newContent.PostedOn = newContent.Status == ContentStatus.Publish || newContent.Status == ContentStatus.Published ? DateTime.UtcNow : null;
+
+        var topicScore = TopicScoreHelper.GetScore(_topicScoreRuleService.FindAll(), model.PublishedOn?.TimeOfDay, model.SourceId, model.Body.Length, model.Section, model.Page, model.SeriesId);
+        if (topicScore != null) {
```
Fosol, line 219: Is there a reason this extra logic isn't the same in all places?  Also could it be moved to the `TopicScoreHelper` to centralize it?

Fosol (approved):

Fosol:
Just need to fix React issue

api/net/Areas/Helpers/TopicScoreHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+                }
+                else
+                {
+                    content.TopicsManyToMany.First().Score = topicScore.Value;
+                }
+            }
+        }
+
+        /// <inheritdoc/>
+        public void SetContentScore(ref ContentModel model)
```
Fosol, line 61: Objects are by reference already, why specify?

Fosol (approved):

PR #1401, opened 2024-01-11
[Ready ✅] TNO-2141: Allow users to enter tags from text area in advanced search

jdtoombs:
Ignore GIF quality using ffmpeg and having driver issues...

Allow users to enter tags from text area in advanced search

- fixes bug causing app to crash when `[` is entered
- can enter `[tag]` in text area to add valid tag, will automatically make uppercase if not already

Working on one more thing - leaving on draft.

tags

Also updated manage folder and allow user to add content to report from here:
image


jdtoombs:
@Fosol - it doesn't actually save it to the query it removes them before storing it to the search, its just an additional way for the user to add tags as per [TNO-2141](https://apps.itsm.gov.bc.ca/jira/browse/TNO-2141). I can scrap the changes if we want though

Fosol (commented):
If the square brackets and the tags are extracted before saving then this will work.

Fosol (approved):

PR #1402, opened 2024-01-11
Fix reports

Fosol:
Updated My Reports Settings tab.

image

### DIFF
--- a/app/subscriber/src/components/section/styled/Section.tsx
+++ b/app/subscriber/src/components/section/styled/Section.tsx
@@ -17,7 +17,6 @@ export const Section = styled.div<ISectionProps>`
     align-items: center;
     padding: 0.25rem 0.5rem;
     background-color: ${(props) => props.theme.css.bkTertiary};
-    flex: 1;
     border-top-right-radius: 0.25rem;
     border-top-left-radius: 0.25rem;
     border-bottom-right-radius: ${(props) => (!props.open ? '0.25rem' : 'unset')};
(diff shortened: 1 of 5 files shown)


app/subscriber/src/components/section/styled/Section.tsx
```diff
@@ -17,7 +17,6 @@ export const Section = styled.div<ISectionProps>`
     align-items: center;
     padding: 0.25rem 0.5rem;
     background-color: ${(props) => props.theme.css.bkTertiary};
-    flex: 1;
```
Fosol, line 20: This was causing the header to balloon when there was little content in it.

app/subscriber/src/features/my-reports/admin/styled/ReportAdmin.tsx
```diff
@@ -125,4 +129,9 @@ export const ReportAdmin = styled.div`
       background: ${(props) => props.theme.css.highlightSecondary};
     }
   }
+
+  .section {
```
Fosol, line 133: Adding borders to sections to make them more visible

app/subscriber/src/features/my-reports/view/styled/ReportEdit.tsx
```diff
@@ -38,6 +38,11 @@ export const ReportEdit = styled.div`
         padding-bottom: 0;
       }
     }
+
+    .section {
```
Fosol, line 42: Adding borders to sections to make them more visible

PR #1403, opened 2024-01-11
TNO-2096 Fix to allow check box selection

fbarreta:
Still keeping the behavior on clicking the whole line to navigate to content, but allowing the click on the check box.

image

### DIFF
--- a/app/subscriber/src/features/search-page/SearchPage.tsx
+++ b/app/subscriber/src/features/search-page/SearchPage.tsx
@@ -181,9 +181,12 @@ export const SearchPage: React.FC<ISearchType> = ({ showAdvanced }) => {
                       >
                         <Col className="cols">
                           <Row>
-                            <Col className="checkBoxColumn" alignItems="center">
+                            <Col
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/search-page/SearchPage.tsx
```diff
@@ -183,6 +183,7 @@ export const SearchPage: React.FC<ISearchType> = ({ showAdvanced }) => {
                           <Row>
                             <Col className="checkBoxColumn" alignItems="center">
                               <Checkbox
+                                onClick={(e) => e.stopPropagation()}
```
Fosol, line 186: Will this have a very narrow focus?  One of the issues users have expressed is that they accidently click on the row and it makes it hard to select content.

fbarreta: As far as I tested works good. But for sanity check, I moved the stop propagation to the column. So now the user have a little room for error.

Fosol (approved):

PR #1404, opened 2024-01-11
TNO-2159: Fix my reports missing sometimes

mihailistov:


app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -17,11 +17,14 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content }) =>
...
   const [reportId, setReportId] = React.useState<number | null>(null);
+
   React.useEffect(() => {
-    if (!myReports.length) {
+    // Only load myReports if content is ready, avoids double-fetch
+    // since this entire component is re-rendered anytime content changes
+    if (!myReports.length && content?.length) {
       findMyReports().catch(() => {});
     }
-    // Only do this on init.
```
mihailistov, line 24: Found another issue where this was actually being double-called because the entire component was rendered twice, once when there's no content, and once when content is finally loaded.

Fosol (approved):
Awesome debugging!

PR #1408, opened 2024-01-12
TNO-2159: Additional fix for reports missing content sometimes

mihailistov:


app/subscriber/src/features/my-reports/admin/ReportAdmin.tsx
```diff
@@ -129,7 +129,7 @@ export const ReportAdmin: React.FC<IReportAdminProps> = ({ path: defaultPath = '
       if (existingReport) {
         setReport(toForm(existingReport));
       } else {
-        getReport(reportId, true)
+        getReport(reportId)
```
mihailistov, line 132: I reversed this back to how it was before as we don't need report content in ReportAdmin page

app/subscriber/src/features/my-reports/ReportCard.tsx
```diff
@@ -73,7 +75,13 @@ export const ReportCard: React.FC<IReportCardProps> = ({ report, onDelete }) =>
...
         </>
       }
-      onChange={(open) => open && fetchReport(report.id)}
+      onChange={(open) => {
+        const existingReport = myReports.find((r) => r.id === report.id);
+        // only need to fetch if we have fetched prior data but it is missing report instance data
+        if (open && !existingReport?.instances?.length) {
+          fetchReport(report.id);
+        }
+      }}
```
mihailistov, line 84: Another optimization, whereas prior to this we were re-fetching the report data every time someone expanded the ReportCard, even if the report had already been fetched

app/subscriber/src/features/my-reports/view/ReportEdit.tsx
```diff
@@ -75,12 +75,13 @@ export const ReportEdit: React.FC = () => {
...
 
   // Case 2: Some report data loaded in store, check if current report is present
-  // if it is, set it to that report, if not, try to fetch it
+  // AND has content, if so, set it to that report, if not, fetch report w/ content
   React.useEffect(() => {
     const reportId = parseInt(id ?? '0');
     if (!!reportId && myReports?.length) {
       const existingReport = myReports.find((r) => r.id === reportId);
-      if (existingReport) {
+      const hasFetchedContent = existingReport?.instances.some((r) => r.content?.length);
```
mihailistov, line 83: This is the best I can do given the current data structure. I am assuming that if there's _any_ content present for any of the instances, that means we have fetched content for this report before, so there's no need to refetch

Here's the rough breakdown of what data is present & when:

- fetchMyReports() -> returns reports with no instance data, so `instances: []`
- getReport(id) -> returns report with instances data but NO content, `instances: [{ ...instanceData, content: [] }]`
- getReport(id, true) -> returns report + instance + **content**
- generateReport(id) -> reports instance + **content**

As long as one of `getReport(id, true)` or `generateReport(id)` have been called by this point. All content data needed should be present.

app/subscriber/src/features/my-reports/ReportCard.tsx
Fosol: This should work for most cases, but doesn't account for the background processes which run auto-reports.

app/subscriber/src/features/my-reports/view/ReportEdit.tsx
Fosol: I think it's a reasonable approach.

Fosol (approved):

PR #1412, opened 2024-01-12
feat: #tno-2221 - added preview button for event of the day

kylgarmor:


Fosol (approved):
I wouldn't have bothered with the DB migration due to the issue of reports not always existing in each environment.  Hope this won't break when I push to PROD

kylgarmor:
That report should exist as its created by a prior migration...

PR #1421, opened 2024-01-16
Tno 2139 filter content by transcription

fbarreta:
- When the list is already rendered, a call is made to retrieve the work orders on the last two days.
- New Column added to display transcript status
- If there are work orders, the new column is filled
- Added option at Show Only to display the pending transcripts
- When the Option is checked, a frontend "filter" function is called to display that status.


app/editor/src/features/content/list-view/hooks/useColumns.tsx
```diff
@@ -118,5 +118,12 @@ export const useColumns = ({ fetch }: IColumnProps): ITableHookColumn<IContentSe
       width: '55px',
       cell: (cell) => <Status value={cell.original.status} />,
     },
+    {
+      accessor: 'transcript',
+      label: 'Transcript',
+      hAlign: 'center',
+      width: '155px',
+      cell: (cell) => <div>{cell.original.transcriptStatus}</div>,
```
Fosol, line 126: can you use the same icons from the transcription page?

Fosol (approved):

PR #1423, opened 2024-01-16
TNO-2149: Move basic search to header

jdtoombs:
**In this PR:**
- basic search will appear in the header when not viewing a search page
- small styling changes to accommodate this

**Things to consider:**
- need to talk to Bobbi on how this will work on mobile, as the horizontal screen real estate is not the same in the header. spent a good chunk of time trying to figure out a "nice" solution but was not happy with it
- may need some new designs to accompany this and or use the old search bar on mobile (will reach out to her) and create another ticket

image

### DIFF
--- a/app/subscriber/src/components/basic-search/BasicSearch.tsx
+++ b/app/subscriber/src/components/basic-search/BasicSearch.tsx
@@ -41,7 +41,7 @@ export const BasicSearch = ({ onSearch, inHeader }: IBasicSearchProps) => {
   };
 
   return (
-    <styled.BasicSearch $inHeader={inHeader}>
+    <styled.BasicSearch inHeader={inHeader}>
       <label>SEARCH FOR: </label>
       <Row className="icon-search">
         <FaSearch onClick={() => handleSearch()} className="search-icon" />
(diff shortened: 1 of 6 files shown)


app/subscriber/src/components/basic-search/BasicSearch.tsx
```diff
@@ -39,7 +41,7 @@ export const BasicSearch = ({ onSearch }: IBasicSearchProps) => {
   };
 
   return (
-    <styled.BasicSearch>
+    <styled.BasicSearch $inHeader={inHeader}>
```
Fosol, line 44: What's with the `$`

jdtoombs: It's just the way they do it on the [styled-component docs](https://styled-components.com/docs/basics#passed-props), probably just a way to differentiate between a styled component prop/ regular prop. Not required by any means though, I could take it out!

jdtoombs: Gives me flashbacks to jQuery lol. I am honestly not a huge fan, I think I am trying to be too "by the books". So far it has resulted in me forgetting the `$` multiple times when passing a prop.

kylgarmor: I like it conceptually but I would need a rule to remind me to use the `$` sign ;)

jdtoombs: I'll take it out for now as we don't currently use this as a standard. Doesn't make much sense unless we changed it everywhere

Fosol (approved):

PR #1428, opened 2024-01-18
feat: #tno-2100 - show checkboxes for advanced search

kylgarmor:
- created check box display for all sections that previously display select boxes
- added ability to swap between check box and select
- select between checkbox and select is stored in a cookie
- update tno core to 0.1.16


app/subscriber/src/features/search-page/components/advanced-search/components/sections/MediaSection.tsx
```diff
@@ -1,13 +1,16 @@
...
 
 import { SubMediaGroups } from '../../constants';
 import { ISubMediaGroupExpanded } from '../../interfaces';
+import { IFilterDisplayProps } from './IFilterDisplayProps';
 import { determineSelectedMedia, sortableMediaOptions } from './utils';
 
-export interface IMediaSectionProps {
+export interface IMediaSectionProps extends IFilterDisplayProps {
+  // /** whether to display filter choices as checkboxes or a select box */
+  // displayFiltersAsDropdown: boolean;
```
Fosol, line 13: Something doesn't look right here.  Below it's not complaining this is commented out.

app/subscriber/src/features/search-page/components/advanced-search/components/sections/SeriesSection.tsx
```diff
@@ -20,18 +31,60 @@ export const SeriesSection: React.FC = () => {
...
+            </Button>
+            /
+            <Button
+              variant={ButtonVariant.link}
+              onClick={() => storeFilter({ ...filter, seriesIds: [] })}
+            >
+              Deselect All
+            </Button>
+          </div>
+          {seriesOptions.map((item, index) => (
```
Fosol, line 53: Just a note regarding using `index` for they `key` in many of your implementations.  This is generally not a reliable key and should only be used when it's the only option.  However, there are safe use-cases for it.  I don't know whether these are safe or not.

app/subscriber/src/features/search-page/components/advanced-search/components/sections/MediaSection.tsx
kylgarmor: good catch, I was experimenting with extending the interface.

Fosol (approved):

PR #1434, opened 2024-01-19
fix: #tno-2236 - fix spacing issues

kylgarmor:
- fixed numerous alignment and spacing issues
- content in the header is now of equal height and is vertically centered
- removed unneccsary padding from header
- changed height calculation for the left nav to avoid weird spacing issue
- removed extra svg padding from collpased navigation
- stop logout action from wrapping at narrower viewport
- removed dead css

### DIFF
--- a/app/subscriber/src/components/header/Header.tsx
+++ b/app/subscriber/src/components/header/Header.tsx
@@ -1,7 +1,7 @@
 import { BasicSearch } from 'components/basic-search';
 import { UserProfile } from 'components/user-profile';
 import React from 'react';
-import { Row, Show } from 'tno-core';
+import { Link, Row, Show } from 'tno-core';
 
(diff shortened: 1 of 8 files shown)


app/subscriber/src/components/header/Header.tsx
```diff
@@ -28,13 +28,13 @@ export const Header: React.FC<IHeaderProps> = ({
     <styled.Header className="header">
       <Row>
         {showLogo && (
-          <div className="logo-container">
+          <Link to="/landing/home" title="MMI Home page" className="logo-container">
```
kylgarmor, line 31: clicking on the log on the header now takes you to the home page of the site

jdtoombs (approved):
You know I am all for it :D

PR #1435, opened 2024-01-19
small css things

jdtoombs:
Just putting this up because I know @kylgarmor is working on some CSS as well and want to make sure we don't step on each other's toes!

- remove bottom border from headline to match new mock up
- make page section top margin consitent with all components

image

### DIFF
--- a/app/subscriber/src/components/section/styled/PageSection.tsx
+++ b/app/subscriber/src/components/section/styled/PageSection.tsx
@@ -7,7 +7,7 @@ export const PageSection = styled.div<{ $ignoreMinWidth?: boolean; $ignoreLastCh
   background: ${(props) => props.theme.css.bkSecondary};
   box-shadow: ${(props) => props.theme.css.boxShadow};
   align-self: stretch;
-  margin: 1rem;
+  margin: 0 1rem 1rem 1rem;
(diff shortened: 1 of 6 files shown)


app/subscriber/src/components/section/styled/PageSection.tsx
```diff
@@ -7,7 +7,7 @@ export const PageSection = styled.div<{ $ignoreMinWidth?: boolean; $ignoreLastCh
   background: ${(props) => props.theme.css.bkSecondary};
   box-shadow: ${(props) => props.theme.css.boxShadow};
   align-self: stretch;
-  margin: 1rem;
+  margin: 0 1rem 1rem 1rem;
```
jdtoombs, line 10: @kylgarmor - seems like this is ok to do and let's us remove the other margin 0s for the top. but let me know if you think otherwise!

Fosol (approved):

app/subscriber/src/components/section/styled/PageSection.tsx
kylgarmor: perfect!

kylgarmor:
so nice to see overloaded css go away

PR #1437, opened 2024-01-19
feat: #tno-1988 - added link to folder or report within existing toast popup

kylgarmor:
- just added link into existing toast to avoid adding more UI clutter


Fosol (approved):
Does it stay open long enough?

kylgarmor:
only time will tell :)  We can make it stay open longer if the user has an issue.  We can also make it stay open indefinitely.

PR #1443, opened 2024-01-20
TNO-2192: Fix issue calculating next report send date

mihailistov:


app/subscriber/src/features/my-reports/utils/calcNextSend.ts
```diff
@@ -17,17 +17,15 @@ interface ISchedule {
...
-    timeZone: 'UTC',
+  const currentTime: Date = new Date();
+  const currentWeekDay = currentTime.toLocaleDateString('en-US', {
+    timeZone: 'America/Vancouver',
+    weekday: 'long',
+  });
+  const currentTimeOfDay = currentTime.toLocaleTimeString('en-US', {
+    // TODO: Store User's preferred timezone in profile/settings to use for coverting/storing
+    // Server currently compares time in server local time, which is PST. Hence, use PST.
+    timeZone: 'America/Vancouver',
```
mihailistov, line 28: Need a sanity check... I don't think last sent date is even relevant? You can't have a future send be in anytime before the present time... so we can just use the current time for calculating next send. @Fosol

Fosol: It depends... The last sent date does indicate whether it was sent.  However, I think your logic is sound.  Whether it was sent or not, the next calculated send date would be in the future anyway.

Fosol (approved):

PR #1444, opened 2024-01-22
Fix chart formatter logic

Fosol:
Fixing Charts API bug.


api/node/src/utils/data.ts
```diff
@@ -6,8 +6,8 @@
 export const convertBase64ConfigToChartJsConfig = (data: string) => {
   const chartJsConfigAsJson = Buffer.from(data, 'base64').toString('ascii');
   const chartJsConfig = JSON.parse(chartJsConfigAsJson);
-  if (chartJsConfig.plugins.datalabels.formatter !== null) {
-    // the expection here is for a lambda type function
+  if (chartJsConfig.plugins?.datalabels?.formatter !== null) {
```
Fosol, line 9: Need to handle nullable/undefined situations or the charts fail.

PR #1452, opened 2024-01-23
TNO-2195 Bugs

jdtoombs:
1. Fix sync issues, list view count is now updated and in sync when adding/removing from the config screen
2. Fix some styling
3. Allow filter select to be clear-able on the configure folder screen
4. Remove unused code

### DIFF
--- a/app/subscriber/src/components/table/styled/SubscriberTableContainer.tsx
+++ b/app/subscriber/src/components/table/styled/SubscriberTableContainer.tsx
@@ -48,6 +48,12 @@ export const SubscriberTableContainer = styled(Col)`
 
     background: ${(props) => props.theme.css.bkWhite};
 
+    // rather than updating the tno-core package and adding props to control the zebra striping of the table with more props, we'll just override the css here
+    // as this wrapper will be applied to most tables in the subscriber app.
+    &:nth-child(2n) {
+      background: inherit !important;
+    }
+
     &:hover {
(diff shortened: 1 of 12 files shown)


app/subscriber/src/components/table/styled/SubscriberTableContainer.tsx
```diff
@@ -48,6 +48,12 @@ export const SubscriberTableContainer = styled(Col)`
 
     background: ${(props) => props.theme.css.bkWhite};
 
+    // rather than updating the tno-core package and adding props to control the zebra striping of the table with more props, we'll just override the css here
+    // as this wrapper will be applied to most tables in the subscriber app.
+    &:nth-child(2n) {
+      background: inherit !important;
```
Fosol, line 54: ugh, why not upgrade tno-core?

Fosol (approved):

app/subscriber/src/components/table/styled/SubscriberTableContainer.tsx
jdtoombs: Lemme do that, just wasn't sure if it was worth a prop or not :laughing:

jdtoombs: Updated, no more `!important`

Fosol (approved):

PR #1455, opened 2024-01-24
[UPDATED] TNO-2136, 2184: New filter by media 

jdtoombs:
** Going to make a seperate ticket to make some of this code shared as `MediaSection` and `FilterMedia` have some duplicate code in their `useEffects()` **

In this PR:

- new filter by media component
image
- alphabetic filter for longer lists
image
- hint indicator and actions when no content is returned (clickable and will take you to that filter)
image
- updates to date filter
- action bar added to list view
- two column layout

### DIFF
--- a/app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
+++ b/app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
@@ -1,7 +1,7 @@
 import { useFilterOptions } from 'components/navbar';
 import { PageSection } from 'components/section';
 import React from 'react';
-import { useContent } from 'store/hooks';
+import { useContent, useLookup } from 'store/hooks';
 import { Col, ISourceModel, Row, Show } from 'tno-core';
 
 import { MediaFilterTypes } from './constants';
@@ -11,25 +11,20 @@ import { alphabetArray } from './utils';
 
 /** The FilterMediaLanding Consists of the filter to narrow down the results of media, as well as the results. It is returned in a two column layout*/
 export const FilterMediaLanding: React.FC = () => {
-  const {
-    dailyPrint,
-    sources,
-    weeklyPrint,
-    cpWire,
-    talkRadio,
-    onlinePrint,
-    television,
-    newsRadio,
-  } = useFilterOptions();
+  const { dailyPrint, weeklyPrint, cpWire, talkRadio, onlinePrint, television, newsRadio } =
+    useFilterOptions();
+  const [{ mediaTypes }] = useLookup();
+  console.log(mediaTypes);
   const [
     {
       mediaType: { filter },
     },
-    { findContentWithElasticsearch, storeMediaTypeFilter: storeFilter },
+    { storeMediaTypeFilter: storeFilter },
   ] = useContent();
   const [activeFilter, setActiveFilter] = React.useState<MediaFilterTypes>(
     MediaFilterTypes.DAILY_PRINT,
   );
+
   const [activeLetter, setActiveLetter] = React.useState<string>('All');
   const [narrowedOptions, setNarrowedOptions] = React.useState<ISourceModel[]>([]);
   const [activeSource, setActiveSource] = React.useState<ISourceModel | null>(null);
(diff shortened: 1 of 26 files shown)


app/subscriber/src/features/filter-media/constants/MediaFilterTypes.ts
```diff
@@ -0,0 +1,9 @@
+export enum MediaFilterTypes {
```
Fosol, line 1: Is there a reason to have an enum for these?  We have a look up which provides all of our media types and it will never have an issue if the names change.

kylgarmor: I second this.

app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
```diff
@@ -0,0 +1,146 @@
+import { useFilterOptions } from 'components/navbar';
+import { PageSection } from 'components/section';
+import React from 'react';
+import { useContent } from 'store/hooks';
+import { Col, ISourceModel, Row, Show } from 'tno-core';
+
+import { MediaFilterTypes } from './constants';
+import { FilterMedia } from './FilterMedia';
+import * as styled from './styled';
+import { alphabetArray } from './utils';
```
kylgarmor, line 10: should we be using a constant for this?  This can lead to a situation where letters show up for which there are no options associated.

app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const {
+    dailyPrint,
+    sources,
+    weeklyPrint,
+    cpWire,
+    talkRadio,
+    onlinePrint,
+    television,
+    newsRadio,
+  } = useFilterOptions();
```
kylgarmor, line 23: I have refactored `useFilterOptions` away in my latest PR.  It wasn't being used anywhere else so I moved it inline with the component.  Might need to move it back out... I'd be happy to assist.

app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ] = useContent();
+  const [activeFilter, setActiveFilter] = React.useState<MediaFilterTypes>(
+    MediaFilterTypes.DAILY_PRINT,
+  );
+  const [activeLetter, setActiveLetter] = React.useState<string>('All');
+  const [narrowedOptions, setNarrowedOptions] = React.useState<ISourceModel[]>([]);
+  const [activeSource, setActiveSource] = React.useState<ISourceModel | null>(null);
+
+  const labels = Object.values(MediaFilterTypes);
+  const determineOptions = () => {
```
kylgarmor, line 38: @jdtoombs Take a look at the way I've refactored the `MediaSection` component.  I think you can do away with the enum and this `determineOptions` lookup.

app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
```diff
@@ -0,0 +1,146 @@
...
+        return newsRadio;
+      case MediaFilterTypes.CP_NEWS:
+        return cpWire;
+      default:
+        return [];
+    }
+  };
+
+  React.useEffect(() => {
+    if (activeFilter === MediaFilterTypes.WEEKLY_PRINT) {
```
kylgarmor, line 60: It might be better to map the use of this to a magic number of options under the section to determine whether to use a letter or the magic `All` option.  Arguably the dictionary of letters is more useful when there's more than can be seen in one block of content.  In your screen grab that would be around 18.

app/subscriber/src/features/filter-media/utils/alphabetArray.ts
```diff
@@ -0,0 +1,9 @@
+/** used to create an array of the alphabet for the filter media narrow down options */
```
kylgarmor, line 1: I think we should re-think this, or alternatively, de-activate letters for which there are no items underneath that letter when the navigation appears.

jdtoombs:
@Fosol / @kylgarmor  - updated just need to rebase with new changes as well as get rid of hard coded strings

app/subscriber/src/features/filter-media/utils/alphabetArray.ts
jdtoombs: Agreed; however, that may be an iterative ticket. Just stuck to what I was given for this one!

app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
```diff
@@ -0,0 +1,161 @@
...
+  ] = useContent();
+  const [{ sources, mediaTypes }] = useLookup();
+  const [activeFilter, setActiveFilter] = React.useState<ISubMediaGroupItem>();
+
+  const [activeLetter, setActiveLetter] = React.useState<string>('All');
+  const [narrowedOptions, setNarrowedOptions] = React.useState<ISourceModel[]>([]);
+  const [activeSource, setActiveSource] = React.useState<ISourceModel | null>(null);
+  const [subMediaGroups, setSubMediaGroups] = React.useState<ISubMediaGroupItem[]>();
+
+  React.useEffect(() => {
```
jdtoombs, line 27: Making another ticket to get rid of this duplication and share it (also used in `MediaSection`

Fosol (approved):

PR #1456, opened 2024-01-24
feat: #tno-2107 - removed need for media type id settings and lookups

kylgarmor:
- refactored away the use of the media type id settings in favor of iterating over the media types and sources to create a dictionary
- removed duplicate MediaSection component
- modified MediaSection to be a bit more self-encapsulated
- removed dead code


app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
```diff
@@ -253,8 +250,8 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ onSearch }) =>
               >
                 <MediaSection
                   displayFiltersAsDropdown={displayFiltersAsDropdown}
-                  setMediaGroupExpandedStates={setMediaGroupExpandedStates}
-                  mediaGroupExpandedStates={mediaGroupExpandedStates}
+                  sources={sources}
+                  mediaTypes={mediaTypes}
```
Fosol, line 254: I like

Fosol (approved):

app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
kylgarmor: yah there didn't seem to be a reason to expose that state outside the component or if there was, it has been lost.  Also, pushing the list of sources and mediaTypes in this way means it doesnt rely directly on redux.  We are probably doing that way more than we need to.  It should be parent component gets the stuff from lookup and passes it down to its children

PR #1461, opened 2024-01-26
 feat: #2224 - subscribers will be kept in sync between weekend and weekday av overview

kylgarmor:
- if another template for STAT holiday is added, just add another value to the AVOverviewTemplateType enum

### DIFF
--- a/api/net/Areas/Admin/Controllers/AVOverviewController.cs
+++ b/api/net/Areas/Admin/Controllers/AVOverviewController.cs
@@ -102,8 +102,52 @@ public class AVOverviewController : ControllerBase
     [SwaggerOperation(Tags = new[] { "Evening Overview" })]
     public IActionResult Update(AVOverviewTemplateModel model)
     {
+        // update the template the user is editing
         var result = _overviewTemplateService.UpdateAndSave((Entities.AVOverviewTemplate)model);
-        var template = _overviewTemplateService.FindById(result.TemplateType) ?? throw new InvalidOperationException("Overview Section does not exist");
+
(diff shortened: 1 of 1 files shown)


api/net/Areas/Admin/Controllers/AVOverviewController.cs
```diff
@@ -102,8 +102,52 @@ public IActionResult Add(AVOverviewTemplateModel model)
...
+            template = _overviewTemplateService.FindById(templateType) ?? throw new InvalidOperationException("Overview Template does not exist");
+            // update the subscription status for matching users
+            foreach (var subscriber in result.SubscribersManyToMany)
+            {
+                var target = template.SubscribersManyToMany.FirstOrDefault((s) => s.UserId == subscriber.UserId);
+                if (target != null) target.IsSubscribed = subscriber.IsSubscribed;
+            }
+
+            // get a list of existing userids so we can determine additions and removals
+            HashSet<int> existingSubscriberUserIds = new HashSet<int>(template.SubscribersManyToMany.Select((s) => s.UserId));
```
Fosol, line 127: Nice.  I generally use the Equal override, but your solution doesn't require wiring up the other objects.

Fosol (approved):

api/net/Areas/Admin/Controllers/AVOverviewController.cs
kylgarmor: equal would not have worked here I think because they are different not only be subscription status, but by the `templateType` or any of the other properties such as the timestamps and such :)

PR #1462, opened 2024-01-26
Tno 2020 list view styling

fbarreta:
Implementations:

- Group by time and order functions implemented
- Media player icon (toggle player)
- Video player
- Audio Player


Fosol (approved):
Thanks

PR #1466, opened 2024-01-26
TNO-2245 Change report summary

Fosol:
Reports will no longer show the Editor Summary but instead will show the Subscriber Summary (if they enter one).

The Subscriber Summary will be included in the report template output.


app/subscriber/src/features/my-reports/view/components/ContentForm.tsx
```diff
@@ -71,7 +71,7 @@ export const ContentForm: React.FC<IContentFormProps> = ({
         <Wysiwyg
           name={`instances.0.content.${index}.versions.${userId}.body`}
           label="Body"
-          value={content.versions?.[userId]?.body ?? content.body ?? ''}
+          value={content.versions?.[userId]?.body ?? content.isApproved ? content.body ?? '' : ''}
```
Fosol, line 74: Fixed bug where an unapproved transcript was being displayed.

PR #1471, opened 2024-01-31
TNO 2106 - My Minister screen

fbarreta:
Overall functionality:

- Filter the Ministers base on user setting
- Iterates this list and make a call for each Minister using the Minister name, Minister aliases and date as parameters
- Get the results and check mentions to each other Minister and populate the Object, also add a count to that minister content.

image

UI Functionalities:

- Checkboxes are displayed with the minister name and content count
- It can be clicked to hide/show a certain Minister content
- A tag to is displayed to show the mentions to other ministers on that content

### DIFF
--- a/app/subscriber/src/features/my-minister/MyMinister.tsx
+++ b/app/subscriber/src/features/my-minister/MyMinister.tsx
@@ -1,6 +1,6 @@
 import { MsearchMultisearchBody } from '@elastic/elasticsearch/lib/api/types';
+import { DateFilter } from 'components/date-filter';
 import { ContentListActionBar } from 'components/tool-bar';
-import { determineColumns } from 'features/home/constants';
 import { filterFormat } from 'features/search-page/utils';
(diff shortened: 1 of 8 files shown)


app/subscriber/src/features/my-minister/MyMinister.tsx
Fosol, line 1: Does this make a separate request for each minister and then visually separate each minister results?

fbarreta: Exactly. For each Minister we have a call.
ministerModels.forEach(async (m) => {
          const res = await fetchResults(
            generateQuery(
              filterFormat({
                ...filter,
                search: makeSimpleQueryString(m.aliases.split(',')),
                startDate: filter.startDate,
                endDate: filter.endDate,
                inByline: true,
                inHeadline: true,
                inStory: true,
                defaultSearchOperator: 'or',
              }),
            ),
          );

Fosol (approved):

PR #1487, opened 2024-02-02
fix: #tno-2237 - moved creationDateOfLastImport to IngestState

kylgarmor:
- renamed ingest_service table to ingest_state to match it's underlying entity
- added creation_date_of_last_item to ingest_state
- removed creationDateOfLastImport from ingest.configuration
- migrated value of ingest.configurationcreationDateOfLastImport to ingest_state.creation_date_of_last_item
- moving the field means that there wont be a new Ingest version created every time an import occurs


libs/net/services/Helpers/ApiService.cs
```diff
@@ -333,19 +333,6 @@ public async Task<IEnumerable<TNO.API.Areas.Services.Models.Ingest.IngestModel>>
         var jsonString = JsonSerializer.Serialize(ingest);
         return await RetryRequestAsync(async () => await this.OpenClient.SendAsync<TNO.API.Areas.Services.Models.Ingest.IngestModel>(url, HttpMethod.Put, JsonContent.Create(ingest)));
     }
-
-    /// <summary>
-    /// Make an HTTP request to the api to update the ingest configuration.
-    /// </summary>
-    /// <param name="ingest"></param>
-    /// <returns></returns>
-    public async Task<TNO.API.Areas.Services.Models.Ingest.IngestModel?> UpdateIngestConfigAsync(
```
Fosol, line 342: How are you updating the ingest with the latest info if you've removed this?  I don't see the corresponding `UpdateIngestStateAsync` method

kylgarmor: `UpdateIngestStateAsync` already existed.  We already use it for updating `FailedAttempts`
I added a new method in the `IngestActionManager` specific to the new state value `UpdateIngestStateCreationDateOfLastItemAsync`

Fosol (approved):

PR #1488, opened 2024-02-02
TNO-2256: Update list view for today's commentary

jdtoombs:
**This PR includes:**
- logic fixes
- remove `HomeFilters` and make `MediaTypeFilters` a dynamic component that can access different stores
- fix in line media bug
- remove dead code


Fosol (commented):
There are a lot of string constants in this.  Can you provide some context to their purpose?

app/subscriber/src/components/media-type-filters/utils/determineStore.ts
```diff
@@ -0,0 +1,12 @@
+import { IContentState } from 'store/slices';
```
jdtoombs, line 1: When passing a store to the `MediaTypeFilter` this util will determine the appropriate `storeFilter` method to call. This reduces the number of props we have to pass to the filter.

app/subscriber/src/components/media-type-filters/MediaTypeFilters.tsx
```diff
@@ -2,22 +2,35 @@ import React, { useEffect, useState } from 'react';
...
 import { Button, ButtonHeight, ContentTypeName, IFilterSettingsModel } from 'tno-core';
 
-import { HomeFilterType } from '../constants';
+import { MediaFilterTypes } from './constants';
 import * as styled from './styled';
+import { determineStore } from './utils';
 
-export interface IHomeFilterProps {}
+export interface IMediaTypeFiltersProps {
+  filterStoreName:
```
jdtoombs, line 10: valid redux store names, could just use `string` and an error would be thrown when deconstructing anyways, but this provides better auto-completion

app/subscriber/src/components/media-type-filters/constants/MediaFilterTypes.ts
```diff
@@ -0,0 +1,8 @@
+export enum MediaFilterTypes {
```
jdtoombs, line 1: Removed `HomeFilterTypes` and add `MediaFilterTypes`. Used in switch cases in `MediaFilter`

app/subscriber/src/store/hooks/subscriber/useContent.ts
```diff
@@ -24,7 +24,7 @@ interface IContentController {
   storeMyMinisterFilter: (filter: IFilterSettingsModel) => void;
   storeTopStoriesFilter: (filter: IFilterSettingsModel) => void;
   storeFrontPageFilter: (filter: IFilterSettingsModel) => void;
-  storeTodayCommentaryFilter: (filter: IFilterSettingsModel) => void;
+  storeTodaysCommentaryFilter: (filter: IFilterSettingsModel) => void;
```
jdtoombs, line 27: spelling error

app/subscriber/src/components/media-type-filters/MediaTypeFilters.tsx
jdtoombs: tried `filterStoreName: keyof IContentState`; however, does not behave with dynamic deconstruction (line 31)

app/subscriber/src/components/media-type-filters/constants/MediaFilterTypes.ts
Fosol: I'm worried this is going back to hardcoded magic strings for media type filters again.  Kyle recently refactored much of this to use our dynamically managed list.

jdtoombs: I'll transfer that over to the newly implemented way - they differ slightly from the mockup but I don't think it should matter

jdtoombs: Will ping once updated - fixing a few things

kylgarmor: this feels like a step backwards.  Since media types are essentially a list which the user can change we should completely avoid having magic strings which match the current values in the list.

app/subscriber/src/components/media-type-filters/FilterOptions.tsx
```diff
@@ -27,78 +40,78 @@ export const HomeFilters: React.FC<IHomeFilterProps> = () => {
     mediaTypeIds: [],
   };
 
-  const handleFilterClick = (type: HomeFilterType) => {
+  const handleFilterClick = (type: FilterOptionTypes) => {
```
jdtoombs, line 43: @kylgarmor - this is where that enum is being used (it was poorly named MediaTypeFilter by me), but as you can see it is its own beast with `contentType`, `mediaTypeId`, and `source`. Spoke with Jeremy and for now we are going to live with it as the grouping does not line up with anywhere else in the app. Renamed to `FilterOptionTypes` for lack of a better name

jdtoombs:
@Fosol - updated after our chat!

Fosol (approved):

PR #1489, opened 2024-02-02
TNO-2123: Fix date picker colours and UI/UX

mihailistov:
- Fix buggy behaviour with start dates before end date and end dates after start date
- CSS, colour updates from Bobbi
- Fix query filter for start but no end date, and end but no start date
- Fixed folder schedule date picker, no months or year nav before
- Only allow scheduling dates from today on
- Only allow filtering for dates on or before today's date

Folder date picker:
Before:
Screenshot 2024-02-02 at 1 17 14 PM

After:
Screenshot 2024-02-02 at 2 43 45 PM

Bugs with advanced search date picker:
Before:
Screenshot 2024-02-02 at 12 57 45 PM

After:
Screenshot 2024-02-02 at 1 01 07 PM

### DIFF
--- a/app/subscriber/src/components/date-filter/DateFilter.tsx
+++ b/app/subscriber/src/components/date-filter/DateFilter.tsx
@@ -62,6 +62,7 @@ export const DateFilter: React.FC<IDateFilterProps> = ({ filter, storeFilter }) 
         <FaCalendarDay className="calendar" onClick={() => setOpen(true)} />
         <ReactDatePicker
           open={open}
+          maxDate={new Date()}
           disabled
           dateFormat="dd-MMM-y"
           onChange={(e) => storeFilter({ ...filter, startDate: e?.toISOString() })}
--- a/app/subscriber/src/features/my-folders/styled/ConfigureFolder.tsx
+++ b/app/subscriber/src/features/my-folders/styled/ConfigureFolder.tsx
@@ -55,19 +55,19 @@ export const ConfigureFolder = styled(PageSection)`
   }
   .react-datepicker__time-list-item--selected,
   .react-datepicker__close-icon::after {
-    background-color: rgb(151, 29, 41) !important;
+    background-color: ${(props) => props.theme.css.btnRedColor}; !important;
   }
   .react-datepicker__close-icon::after {n
     background-color: ${(props) => props.theme.css.btnBkPrimary};
   }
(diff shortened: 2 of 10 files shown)


app/subscriber/src/components/date-filter/DateFilter.tsx
```diff
@@ -62,6 +62,7 @@ export const DateFilter: React.FC<IDateFilterProps> = ({ filter, storeFilter })
         <FaCalendarDay className="calendar" onClick={() => setOpen(true)} />
         <ReactDatePicker
           open={open}
+          maxDate={new Date()}
```
Fosol, line 65: Won't this always default today?  We don't want a value in the component if the user didn't select one.

app/subscriber/src/features/my-folders/styled/ConfigureFolder.tsx
```diff
@@ -53,9 +53,31 @@ export const ConfigureFolder = styled(PageSection)`
...
+  .react-datepicker__day--selected {
+    background-color: rgb(151, 29, 41);
+  }
+
+  .react-datepicker__day--keyboard-selected,
+  .react-datepicker__month-text--keyboard-selected,
+  .react-datepicker__quarter-text--keyboard-selected,
+  .react-datepicker__year-text--keyboard-selected {
+    &:hover {
+      background-color: rgb(151, 29, 41);
```
jdtoombs, line 78: Might be worth adding these colors to our variables - other than that looks good!

app/subscriber/src/components/date-filter/DateFilter.tsx
mihailistov: @Fosol yes exactly, it just sets a limit on the furthest date into the future that the user can select, it doesn't actually set any values... everywhere we use DateFilter currently this logic is valid

mihailistov: Just to elaborate, these are the places that will have the calendar component with a max selectable date as today. I think they're all valid, are any of these pages areas you would need to pick a date in the future (beyond today)?

Screenshot 2024-02-02 at 6 25 04 PM

If we need to add a prop to pass into `DateFilter` component, I can definitely do that, but right now everywhere would use the same maxDate, so it's not necessary yet. If we need to customize that, we can add the prop later

Fosol (approved):

PR #1512, opened 2024-02-08
TNO-2293 Fix unapproved user

Fosol:
Fixing linting issues.

Fixing logic to extract and apply roles to an first time authenticated user.

### DIFF
--- a/api/net/Keycloak/CssHelper.cs
+++ b/api/net/Keycloak/CssHelper.cs
@@ -135,14 +135,14 @@ public class CssHelper : ICssHelper
         // If user doesn't exist, add them to the database.
         if (user == null)
         {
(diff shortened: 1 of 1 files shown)


api/net/Keycloak/CssHelper.cs
```diff
@@ -189,19 +189,20 @@ private async Task AddOrUpdateUserAsync(string username, Entities.User? user, TN
                 user.EmailVerified = principal.GetEmailVerified() ?? false;
 
                 // Apply the preapproved roles to the user.
-                var roles = await UpdateUserRolesAsync(key, user.Roles.Split(",").Select(r => r[1..^1]).ToArray());
```
Fosol, line 192: I don't understand this logic here.  Why would we remove the first character from every role?

PR #1515, opened 2024-02-08
Featured Content - Fix erroneous search of "Homepage" -> "Featured Story"

jdtoombs:
Was a discrepancy in my local as I had not updated my admin settings

### DIFF
--- a/app/subscriber/src/features/home/Home.tsx
+++ b/app/subscriber/src/features/home/Home.tsx
@@ -2,13 +2,13 @@ import { MsearchMultisearchBody } from '@elastic/elasticsearch/lib/api/types';
 import { ContentList } from 'components/content-list';
 import { DateFilter } from 'components/date-filter';
 import { ContentListActionBar } from 'components/tool-bar';
-import { filterFormat, getFilterActions } from 'features/search-page/utils';
+import { filterFormat } from 'features/search-page/utils';
 import { createFilterSettings } from 'features/utils';
 import { IContentSearchResult } from 'features/utils/interfaces';
 import moment from 'moment';
 import React, { useMemo } from 'react';
 import { useContent, useLookup } from 'store/hooks';
-import { ActionName, generateQuery, IContentModel, IFilterSettingsModel, Row } from 'tno-core';
+import { generateQuery, IContentModel, IFilterSettingsModel, Row } from 'tno-core';
 
 import * as styled from './styled';
 
@@ -54,7 +54,15 @@ export const Home: React.FC = () => {
   );
 
   const homePage = React.useMemo(() => {
-    return getFilterActions(actions)[ActionName.Homepage];
+    const featured = actions.find((a) => a.name === 'Featured Story');
+    if (!featured?.id) return undefined;
+    const featuredFilterValue = {
+      id: featured?.id,
+      value: String(true),
+      valueType: featured?.valueType,
+    };
+
+    return featuredFilterValue;
   }, [actions]);
 
   React.useEffect(() => {
--- a/app/subscriber/src/features/home/Home.tsx
+++ b/app/subscriber/src/features/home/Home.tsx
@@ -8,7 +8,7 @@ import { IContentSearchResult } from 'features/utils/interfaces';
 import moment from 'moment';
 import React, { useMemo } from 'react';
 import { useContent, useLookup } from 'store/hooks';
-import { generateQuery, IContentModel, IFilterSettingsModel, Row } from 'tno-core';
+import { generateQuery, IContentModel, IFilterSettingsModel, Row, Settings } from 'tno-core';
 
 import * as styled from './styled';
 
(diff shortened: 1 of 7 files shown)


app/subscriber/src/features/home/Home.tsx
```diff
@@ -54,7 +54,15 @@ export const Home: React.FC = () => {
   );
 
   const homePage = React.useMemo(() => {
-    return getFilterActions(actions)[ActionName.Homepage];
+    const featured = actions.find((a) => a.name === 'Featured Story');
```
jdtoombs, line 57: Not sure of what the best way to do this would be other than hard coding / making another magic enum

jdtoombs: As the names are managed by admins on the site

Fosol: This is what Settings are for.  We created them for dynamic constants.

jdtoombs: I may be over complicating this, but in the admin page for `Actions` the admin updates the name of said actions but the `action.name` should really be paired up with `Settings` in some way? Unless we want them to change the action name (in _Admin -> Actions_), then additionally update a variable under _Admin -> Settings_

jdtoombs: Will adjust

jdtoombs: Had a d'oh moment, but updated now @Fosol

app/subscriber/src/features/home/Home.tsx
```diff
@@ -53,15 +55,32 @@ export const Home: React.FC = () => {
     [findContentWithElasticsearch],
   );
 
+  const featuredItemsName = React.useMemo(() => {
+    const value = appSettings?.find((s) => s.name === Settings.FeaturedItemsName)?.value;
+    if (!value && !!appSettings?.length)
```
jdtoombs, line 60: just in case its not set

Fosol: Don't use a featured name in Settings.  Use the key so that in the future the admin can change the name and everything works.  Just like it was Homepage and now is Featured Story.

Fosol (changes requested):

app/subscriber/src/features/home/Home.tsx
jdtoombs: Just need some clarification, `Settings.FeaturedItemsName` is the key for "Featured Story" and it can be configured in Settings if the name changes again

jdtoombs: image

jdtoombs: Let me know if I am misunderstanding @Fosol

Fosol: The Settings should almost always point to primary keys.  An admin can one day change the name of "Featured Stories" to "Today's Featured Stories".  Just like they changed it from "Homepage" to "Featured Stories".  The setting would keep working if it used the primary key.

With your implementation it would break.

Fosol (changes requested):

app/subscriber/src/features/home/Home.tsx
jdtoombs: Ah I was avoiding using `id` because of some overthought that the action could change resulting in a different `id`. Will fix

Fosol: A primary key cannot change.  If they deleted the action then the filter would stop working.  Which would be expected behaviour.  They would need to then go to the settings and choose a new action to filter on.

app/subscriber/src/features/home/Home.tsx
```diff
@@ -53,15 +55,32 @@ export const Home: React.FC = () => {
...
+      toast.error(
+        'No FeaturedItemsName found in settings. Please contact your administrator to update.',
+      );
+    return value;
+  }, [appSettings]);
+
   const homePage = React.useMemo(() => {
-    return getFilterActions(actions)[ActionName.Homepage];
-  }, [actions]);
+    const featured = actions.find((a) => a.name === featuredItemsName);
```
Fosol, line 68: Do not need this.

app/subscriber/src/features/home/Home.tsx
```diff
@@ -53,15 +55,32 @@ export const Home: React.FC = () => {
     [findContentWithElasticsearch],
   );
 
+  const featuredItemsName = React.useMemo(() => {
+    const value = appSettings?.find((s) => s.id === Settings.FeaturedItem)?.value;
+    if (!value && !!appSettings?.length)
+      toast.error(
+        'No FeaturedItemsName found in settings. Please contact your administrator to update.',
```
Fosol, line 62: Update the key name.

Fosol (changes requested):

Fosol (approved):

PR #1520, opened 2024-02-09
TNO-2312: Add active item indicator 

jdtoombs:
Add indicator for which nav option the user is on

- also update `yarn.lock` forgot to include this on last PR

image


app/subscriber/src/components/navbar/Navbar.tsx
```diff
@@ -27,12 +32,21 @@ export const Navbar: React.FC<INavbarProps> = ({ options }) => {
   const groupedOptions = options?.filter((option) => !!option.groupName);
   const groupByName = _.groupBy(groupedOptions, 'groupName');
 
+  const determineClassName = React.useCallback((path: string) => {
+    if (window.location.pathname.includes(path)) return 'active';
```
Fosol, line 36: React will not know when `window.location.pathname` changes natively.  I believe you need to use the `useLocation()`.

Fosol (changes requested):

app/subscriber/src/components/navbar/Navbar.tsx
jdtoombs: updated!

Fosol (approved):

PR #1523, opened 2024-02-12
feat: #tno-2305 - add row layout styling for front page images

kylgarmor:


libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -2391,6 +2391,10 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasColumnType("integer")
                         .HasColumnName("ingest_id");
 
+                    b.Property<DateTime?>("CreationDateOfLastItem")
```
Fosol, line 2394: This worries me.  It indicates our migration has issues.

kylgarmor: yah I will double check that actually because something did happen when I created the initial migration here.  I assumed I had just copy pasted incorrectly or something.

Fosol (approved):

PR #1528, opened 2024-02-13
fix: #tno-2301 - nav under content in paper is in sync

kylgarmor:
- if the user navigates in a content tab/window the currently highlited row will be selected


Fosol (approved):
That didn't seem as difficult as I thought it would be.

PR #1533, opened 2024-02-14
[UPDATED] TNO-2272, 2253: Tech debt and feedback for filter media

jdtoombs:
In this PR:

- new hook for accessing `subMediaGroups`
- add time to filter media results
- add series name to filter media results
- add full name to narrowed options in filtered media 
- remove narrowed options for `Events`

image


app/subscriber/src/features/utils/createSubMediaGroups.ts
```diff
@@ -0,0 +1,60 @@
...
+import { IMediaTypeModel, ISourceModel } from 'tno-core';
+
+/**
+ * Create Sub Media Groups
+ * @param sources
+ * @param mediaTypes
+ * @param setSubMediaGroups
+ * @param setMediaGroupExpanded - optional
+ */
+export const createSubMediaGroups = (
```
Fosol, line 14: This is a perfect example of something that should become a hook.  If you're passing state stuff from other components, just move the state stuff to the hook so that the component can pull it out instead like all other hooks.

Fosol (changes requested):

app/subscriber/src/features/utils/createSubMediaGroups.ts
jdtoombs: Just debugging some dependency bug with the new hook, will ping once I fix it

jdtoombs: New commit [here!](https://github.com/bcgov/tno/pull/1533/commits/0503937675c1a01028e9f0795215a27dae802995) @Fosol

Fosol (approved):

PR #1536, opened 2024-02-15
feat: #tno-2270 - address feedback from stakeholder

kylgarmor:
- add grouping to the topic drop down
- needed to use the regular react-select instead of the overridden version as it was proving too difficult

### DIFF
--- a/app/editor/src/features/admin/event-of-the-day/EventOfTheDayList.tsx
+++ b/app/editor/src/features/admin/event-of-the-day/EventOfTheDayList.tsx
@@ -92,7 +92,6 @@ const EventOfTheDayList: React.FC = () => {
     findAllTopics()
       .then((data) => {
         setAllTopics(data);
-        // setTopicOptions(getSortedTopicOptions(data));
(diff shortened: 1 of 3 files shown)


app/editor/src/features/admin/event-of-the-day/EventOfTheDayList.tsx
```diff
@@ -86,7 +92,8 @@ const EventOfTheDayList: React.FC = () => {
     findAllTopics()
       .then((data) => {
         setAllTopics(data);
-        setTopicOptions(getSortedTopicOptions(data));
+        // setTopicOptions(getSortedTopicOptions(data));
```
Fosol, line 95: clean up?

Fosol (approved):

app/editor/src/features/admin/event-of-the-day/EventOfTheDayList.tsx
kylgarmor: done

PR #1544, opened 2024-02-16
Press gallery fixes

jdtoombs:
- group date content on frontend instead of sending 7 days worth of elastic requests
- allow user to enter each filter, not one or the other
- clean up some code

Leaving in draft as I am playing with some things still. 

Also correct me if I am wrong - but grouping based on contributor aliases in headline, byline, body etc.. on the frontend would probably be slower than utilizing elastic search? 

Date is easy enough to group; however, I figured it may be best to get the hits from individual elastic search queries still (for each press member). Going to try and get an aggregate response for just the hits on that query. (in progress)

Another option which I would be a fan of.... is removing the hits completely from the drop down lists. Especially since the user can apply both filters at once now it just adds unnecessary complexity imo.

### DIFF
--- a/app/subscriber/src/features/press-gallery/PressGallery.tsx
+++ b/app/subscriber/src/features/press-gallery/PressGallery.tsx
@@ -34,10 +34,12 @@ export const PressGallery: React.FC = () => {
 
   const [content, setContent] = React.useState<IContentSearchResult[]>([]);
   const [pressMembers, setPressMembers] = React.useState<IPressMember[]>([]);
+  const [initialLoad, setInitialLoad] = React.useState(false);
   const [selected, setSelected] = React.useState<IContentModel[]>([]);
   const [dateOptions, setDateOptions] = React.useState<IDateOptions[]>([]);
   const [aliases, setAliases] = React.useState<string[]>([]);
   const [loading, setLoading] = React.useState(false);
+  const [groupedContent, setGroupedContent] = React.useState<any[]>([]);
   const [pressSettings] = React.useState<IFilterSettingsModel>(
     createFilterSettings(`${moment().startOf('day')}`, `${moment().subtract('2', 'weeks')}`),
   );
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/press-gallery/PressGallery.tsx
```diff
@@ -92,6 +88,23 @@ export const PressGallery: React.FC = () => {
     setDateOptions(dates);
   }, []);
 
+  // group content by date on the frontend instead of addtioanl fetches for each date
+  React.useEffect(() => {
+    if (!!content.length) {
+      const grouped = content.reduce((acc: any, content) => {
```
jdtoombs, line 94: fix

app/subscriber/src/features/press-gallery/PressGallery.tsx
```diff
@@ -34,10 +34,12 @@ export const PressGallery: React.FC = () => {
 
   const [content, setContent] = React.useState<IContentSearchResult[]>([]);
   const [pressMembers, setPressMembers] = React.useState<IPressMember[]>([]);
+  const [initialLoad, setInitialLoad] = React.useState(false);
   const [selected, setSelected] = React.useState<IContentModel[]>([]);
   const [dateOptions, setDateOptions] = React.useState<IDateOptions[]>([]);
   const [aliases, setAliases] = React.useState<string[]>([]);
   const [loading, setLoading] = React.useState(false);
+  const [groupedContent, setGroupedContent] = React.useState<any[]>([]);
```
jdtoombs, line 42: fix

Fosol: Don't use `any`

jdtoombs:
have an idea closing for now

app/subscriber/src/features/press-gallery/PressGallery.tsx
jdtoombs: thats what the fix is for :stuck_out_tongue:, closing this PR for now though have an idea

jdtoombs: (was telling myself to fix)

PR #1545, opened 2024-02-17
feat: #tno-2270 - more features added

kylgarmor:
- fixed report so that it clears folder after report is sent
- fixed aggregations count - excluding part of a nested field in an inner aggregate doesnt exclude it from the parent
- fixed report engine so that it skipes disabled sections during render
- added Send action to the Event of the day page


app/editor/src/features/admin/event-of-the-day/EventOfTheDayList.tsx
```diff
@@ -244,6 +260,23 @@ const EventOfTheDayList: React.FC = () => {
...
+        onConfirm={async () => {
+          try {
+            await handlePublish();
+          } catch {
+            // Globally handled
+          } finally {
+            toggle();
+          }
+        }}
+      />
```
kylgarmor, line 279: Added Send feature to the Event of the Day page

libs/net/template/ReportEngine.cs
```diff
@@ -253,7 +253,7 @@ public IEnumerable<ContentModel> OrderBySectionField(IEnumerable<ContentModel> c
         {
             var aggregateSection = new Dictionary<string, ReportSectionModel>();
             // Collect all section content for report summary charts.
-            report.Sections.Where(section => section.SectionType == Entities.ReportSectionType.Content).ForEach(section =>
+            report.Sections.Where(section => section.SectionType == Entities.ReportSectionType.Content && section.IsEnabled).ForEach(section =>
```
kylgarmor, line 256: skip rendering content for disabled sections

libs/net/elastic/Models/Converters/AggregationsConverter.cs
```diff
@@ -50,13 +50,16 @@ private static AggregationRootModel ParseAggregationRootModels(ref Utf8JsonReade
...
+                    dto.DocCount = reader.GetInt64(); // this result is not reliable
                     break;
                 default:
                     dto.ChildAggregation = ParseAggregationModel(ref reader, propName);
                     break;
             }
         }
+        // calculate the sum DocCount of ChildAggregation.Buckets
+        if (dto.ChildAggregation.Buckets.Any())
+            dto.DocCount = dto.ChildAggregation.Buckets.Sum(b => b.DocCount);
```
kylgarmor, line 62: if there are child aggregations manually calculate the same ignoring the doc count from elastic.  Elastic doesn't exclude parent items when child is excluded

libs/net/template/ReportEngine.cs
```diff
@@ -262,7 +262,7 @@ public IEnumerable<ContentModel> OrderBySectionField(IEnumerable<ContentModel> c
                 }
             });
 
-            await report.Sections.Where(section => section.SectionType == Entities.ReportSectionType.MediaAnalytics).ForEachAsync(async section =>
+            await report.Sections.Where(section => section.SectionType == Entities.ReportSectionType.MediaAnalytics && section.IsEnabled).ForEachAsync(async section =>
```
kylgarmor, line 265: skip rendering content for disabled sections

libs/net/elastic/Models/Converters/AggregationsConverter.cs
```diff
@@ -105,7 +108,7 @@ private static IEnumerable<AggregationBucketModel> ParseAggregationBucketModels(
                 switch (propName.ToLower())
                 {
                     case var _ when propName.Equals(PropNameDocCount.ToLower()):
-                        bucket.DocCount = reader.GetInt64();
+                        bucket.DocCount = reader.GetInt64(); // unreliable
```
Fosol, line 111: If unreliable should you have it a TryParse instead?

kylgarmor: it's that the value is inherently wrong.  There might be a way to fix in the aggregation query itself, but this feels sane.
Part of the problem is that we have the "Not Applicable" Topic assigned to a valid "Topic Type". In the Elastic db that nested value is denormalized so when we group by TopicType and then TopicName, it doesnt care that "Issue:Not Applicable" has a child exclusion for "Not Applicable".  It just counts all the hits on "Issue".

Fosol (approved):

PR #1549, opened 2024-02-21
Tno 2273 evening overview feedback

fbarreta:
- Auto complete fix (avoid duplicates, and remove index check)
- Load content fix (get initial date from the formik set value, previously we were one step behind monitoring url params)
- Weekday and weekend template update preventing column wrap with a fixed width.


app/editor/src/features/reports/av-overview/OverviewGrid.tsx
```diff
@@ -343,14 +341,17 @@ export const OverviewGrid: React.FC<IOverviewGridProps> = ({ editable = true, in
...
                                         suggestions = summaries
                                           .sort()
-                                          .filter(
-                                            (v: Suggestion) =>
-                                              regex.test(v.text) && v.index !== itemIndex,
-                                          );
+                                          .filter((v: Suggestion) => regex.test(v.text))
+                                          .filter((item) => {
+                                            const duplicate = sSet.has(item.text);
+                                            sSet.add(item.text);
```
kylgarmor, line 352: should this be added if it's a duplicate?

kylgarmor (approved):

PR #1550, opened 2024-02-21
TNO-2287 Subscriber color coding by environment

fbarreta:
Background color coding according to environment.


jdtoombs (approved):
Looks good to me, and thanks for adding those scss variables :)

PR #1551, opened 2024-02-22
TNO-2215 Audio video layout fix

fbarreta:
image
Before

image
After

### DIFF
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -536,7 +536,7 @@ const ContentForm: React.FC<IContentFormProps> = ({
                     </Col>
                   </Show>
                 </Row>
-                <Row className="tab-section fvh">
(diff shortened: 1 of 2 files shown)


jdtoombs:
Looks good! Just one small thing, I noticed after the changes the Publish/Save without publish button is no longer appearing in one row as shown in your attached images:

image

kylgarmor:
hey Felipe.  Could you maybe change the percentage of the screen that the media section gets?  I dont see any reason for it to get 50% whereas the text-area could definitely benefit from more space

fbarreta:
> Looks good! Just one small thing, I noticed after the changes the Publish/Save without publish button is no longer appearing in one row as shown in your attached images:
> 
> image

It's because my window size, the default behaviour when the size is smaller, when it's regular size, it looks the same:
image

fbarreta:
> hey Felipe. Could you maybe change the percentage of the screen that the media section gets? I dont see any reason for it to get 50% whereas the text-area could definitely benefit from more space

Yeah ... looked better with 65%.
image

kylgarmor (approved):

PR #1555, opened 2024-02-23
fix: #tno-2352 - added logging and tweaked logic

kylgarmor:
- adopted a "fail fast" method when verifying ingest schedules
- log reason why ingest is not being run
- reset failures AFTER initial checks have taken place

### DIFF
--- a/libs/net/services/Actions/Managers/IngestActionManager`.cs
+++ b/libs/net/services/Actions/Managers/IngestActionManager`.cs
@@ -67,6 +67,16 @@ public class IngestActionManager<TOptions> : ServiceActionManager<TOptions>, IIn
     /// <returns></returns>
     protected override Task<bool> PreRunAsync()
     {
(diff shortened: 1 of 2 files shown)


libs/net/services/Actions/Managers/IngestActionManager`.cs
```diff
@@ -67,6 +67,16 @@ public class IngestActionManager<TOptions> : ServiceActionManager<TOptions>, IIn
     /// <returns></returns>
     protected override Task<bool> PreRunAsync()
     {
+        if (this.Ingest.IsEnabled && (!this.Ingest.IngestSchedules.Any() || this.Ingest.IngestSchedules.All(s => s.Schedule == null))) {
```
Fosol, line 70: Surprised this isn't already handled

Fosol (approved):

PR #1556, opened 2024-02-23
feat: #tno-2329 - moved Topic component into shared location

kylgarmor:
- updated Event of the day page and Content page top use shared component


app/editor/src/components/topic/styled/Topic.tsx
```diff
@@ -0,0 +1,37 @@
...
+    font-size: 0.75rem;
+    width: 1rem;
+    height: 1rem;
+    text-align: center;
+    font-weight: bold;
+    margin-right: 0.25rem;
+    text-transform: uppercase;
+  }
+  .option-hint.type-Proactive {
+    background-color: #006600;
```
jdtoombs, line 29: nit picking but wondering if we have a variable for these colors

Fosol (approved):

PR #1557, opened 2024-02-26
fix: #tno-2354 - sped up response time from adding a topic

kylgarmor:
- do not make a call to refresh ALL lookups just because we added a new topic
- when adding a new topic, just update Topics in redux and localstorage
- may be a new pattern here to increase responsiveness in lists?

### DIFF
--- a/app/editor/src/store/hooks/admin/useTopics.ts
+++ b/app/editor/src/store/hooks/admin/useTopics.ts
@@ -1,7 +1,14 @@
 import React from 'react';
 import { useAjaxWrapper, useLookup } from 'store/hooks';
 import { IAdminState, useAdminStore } from 'store/slices';
-import { IPaged, ITopicFilter, ITopicModel, useApiAdminTopics } from 'tno-core';
+import {
+  IPaged,
+  ITopicFilter,
+  ITopicModel,
+  saveToLocalStorage,
+  StorageKeys,
+  useApiAdminTopics,
+} from 'tno-core';
 
 interface ITopicController {
   findAllTopics: () => Promise<ITopicModel[]>;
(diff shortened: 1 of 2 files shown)


app/editor/src/store/hooks/admin/useTopics.ts
```diff
@@ -45,12 +52,14 @@ export const useTopics = (): [IAdminState, ITopicController] => {
...
         const response = await dispatch<ITopicModel>('add-topic', () => api.addTopic(model));
+        let items: ITopicModel[] = [];
+
         store.storeTopics((topics) => {
-          var items = [...topics];
+          items = [...topics];
           items.splice(model.sortOrder, 0, response.data);
           return items;
         });
-        await lookup.getLookups();
```
kylgarmor, line 53: this line is out of place here.  we should not have to retrieve ALL the lookups just because we added one new topic

app/editor/src/store/hooks/admin/useTopics.ts
```diff
@@ -45,12 +52,14 @@ export const useTopics = (): [IAdminState, ITopicController] => {
...
+        saveToLocalStorage(StorageKeys.Topics, items, store.storeTopics);
```
kylgarmor, line 62: store the updated list of Topics to LocalStorage

app/editor/src/store/hooks/lookup/useLookup.ts
```diff
@@ -219,14 +219,16 @@ export const useLookup = (): [ILookupState, ILookupController] => {
           'lookup',
         );
       },
-      getTopics: async () => {
+      getTopics: async (refresh?: boolean) => {
```
kylgarmor, line 222: strangely, we have this refresh parameter in the declaration of the method but then we dont use it.  This implementation seems to make sense.

PR #1566, opened 2024-02-28
fix: #tno-2002 - series - program cleanup

kylgarmor:
- added the ability to merge one Series Program with another
- from the "master" record, the user picks another record they want to merge
- when the records are merged, all content associated with the record that is merged are re-associated with the "master"
- the record which will now have no content associated with is is deleted.


app/editor/src/features/admin/series/SeriesMerge.tsx
```diff
@@ -0,0 +1,126 @@
...
+    (state as any)?.series ?? defaultSeries,
+  );
+  const [seriesOptions, setSeriesOptions] = React.useState<IOptionItem[]>([]);
+  const [mergeSeriesSourceOption, setMergeSeriesSourceOption] = React.useState<IOptionItem>();
+
+  const targetSeriesId = Number(id);
+
+  React.useEffect(() => {
+    if (!!targetSeriesId && targetSeries?.id !== targetSeriesId) {
+      setTargetSeries({ ...defaultSeries, id: targetSeriesId }); // Do this to stop double fetch.
```
Fosol, line 44: Make sure the double fetch isn't the idiotic think React does in development mode...

Fosol (approved):

app/editor/src/features/admin/series/SeriesMerge.tsx
kylgarmor: in all honesty, I just copied that boilerplate code from the original `SeriesForm` page

PR #1569, opened 2024-02-28
TNO-2290: Fix filter media type query bugs

mihailistov:
Reduced # of calls from ~10 to 1 on Filter Media page, by reducing redundant queries, and combining 5 separate queries for 5 days, into 1.

Before: 
Screenshot 2024-02-16 at 11 56 30 AM

After:
Screenshot 2024-02-28 at 3 28 45 PM


Fosol (approved):
Thanks for this, it'll make a big difference

PR #1570, opened 2024-02-29
fix: #tno-2359 - fixed filter and patched ReportingManager

kylgarmor:
- filter was breaking the report sending because it wasnt including the id
- reporting manager fails when report send is triggered if the content id is invalid


Fosol (approved):
Thanks

PR #1573, opened 2024-02-29
TNO-2336 Send email on failure

fbarreta:
Send email on service failure.

I came with a no email format. If it's OK. I'll replicate into the other services.

### DIFF
--- a/services/net/folder-collection/FolderCollectionManager.cs
+++ b/services/net/folder-collection/FolderCollectionManager.cs
@@ -3,6 +3,8 @@ using System.Text.Json.Nodes;
 using Confluent.Kafka;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using TNO.Ches;
+using TNO.Ches.Configuration;
 using TNO.Core.Exceptions;
 using TNO.Core.Extensions;
 using TNO.Elastic;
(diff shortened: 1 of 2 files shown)


services/net/folder-collection/FolderCollectionManager.cs
```diff
@@ -64,20 +78,63 @@ public class FolderCollectionManager : ServiceManager<FolderCollectionOptions>
...
+    public async Task SendEmailAsync(string subject, string message)
+    {
+        if (this.Options.SendEmailOnFailure)
+        {
+            try
+            {
+                var email = new TNO.Ches.Models.EmailModel(this.ChesOptions.From, this.Options.EmailTo, subject, message);
+                await this.Ches.SendEmailAsync(email);
+            }
+            catch (ChesException ex)
```
Fosol, line 114: Does the ChesService ever throw any other type of error?  If it does you'll need to handle it, or this process can break other processes.

Fosol:
Do we need this anymore?

services/net/folder-collection/FolderCollectionManager.cs
fbarreta: No, we can remove this PR.

fbarreta:
This changes will be addressed on this pr:
https://github.com/bcgov/tno/pull/1579

PR #1578, opened 2024-03-01
TNO-2280: Fix keyboard shortcuts on report page

mihailistov:


app/subscriber/src/features/my-reports/view/components/ReportEditForm.tsx
```diff
@@ -125,15 +125,6 @@ export const ReportEditForm: React.FC<IReportEditFormProps> = ({
   return (
     <PageSection
       tabIndex={0}
-      onKeyDown={(e) => {
```
mihailistov, line 128: we don't need this anymore as the window listener will pick up for events whenever the edit form is mounted

Fosol (approved):
awesome

app/subscriber/src/features/my-reports/view/components/ContentEditForm.tsx
```diff
@@ -146,18 +146,42 @@ export const ContentEditForm = ({
...
+  const handleKeyDown = React.useCallback(
+    (e: KeyboardEvent) => {
+      if (!form?.content) return;
+      if (e.code === 'Escape') onUpdate?.();
+      // added metaKey support for MAC
+      // allows use of Command key as well as Control key for windows
+      else if (e.ctrlKey || e.metaKey) {
+        if (e.code === 'Enter') handleAddUpdateContent(values, form);
+        else if (e.code === 'ArrowUp' || e.code === 'ArrowLeft') {
+          console.log('NAV PREV');
```
kylgarmor, line 158: do we need these `console.log` statements?

Fosol:
Does this handle input boxes and allow users to use the arrow keys within those?

mihailistov:
> Does this handle input boxes and allow users to use the arrow keys within those?

@Fosol yes, I just tested now to make sure, since the listener is on the window, it'll pick up the events no matter where the focus is

Fosol:
Yes would be bad.  If a user is typing in an input they don't want the data changing out from under them.

mihailistov:
> Yes would be bad. If a user is typing in an input they don't want the data changing out from under them.

What are the chances they'll press CTRL+Arrow keys in an input box? I think it's a bit odd to have hotkeys but disable them for inputs. If they want to nav back and forth they'll have to click outside of a input box to blur it, _and then_ press the hotkeys

I can add something to ignore the hotkeys though, if you think that's best

Fosol:
Definitely ignore the prev/next shortcut keys actions within an any kind of input.  That means ignore the page level prev/next actions you're implementing when a user is in an input.

mihailistov:
> Definitely ignore the prev/next shortcut keys actions within an any kind of input. That means ignore the page level prev/next actions you're implementing when a user is in an input.

An alternative would be prompting the user if there any unsaved changes, "Are you sure you want to navigate away from this content?". Just throwing ideas out there.

Fosol:
That isn't ideal.  If I'm in an input and I want to move the cursor to another word I don't want a prompt firing every time.  I just want the cursor to move in the in put.

mihailistov:
> That isn't ideal. If I'm in an input and I want to move the cursor to another word I don't want a prompt firing every time. I just want the cursor to move in the in put.

For UX perspective, Google Chrome has a shortcut CTRL+TAB to navigate to the next tab in the browser, this works no matter where the focus is, I could be typing in this input, and it'll still work. It would be awkward if I had to first click outside the input for it to work. Kind of eliminates the point of a shortcut.

Maybe we should use a different shortcut rather than CTRL+Arrow keys

For now I'll work on disabling it in inputs

Fosol:
Perhaps I'm misunderstanding this PR.  Is this to handle the previous/next actions to move from one content to another within a report?  If so when a content item is selected a form is displayed.  That form is to enable the user to edit the content properties.  When a user focuses on an input they are changing the property values.  A valid keypress is the arrow keys within an input to allow the user to move the cursor.

Will the arrow keys work within the input as they normally would?  Or will this PR change the behavior?

All we want to make certain of is users should be able to use the arrow keys within an input normally.

mihailistov:
> Perhaps I'm misunderstanding this PR. Is this to handle the previous/next actions to move from one content to another within a report? If so when a content item is selected a form is displayed. That form is to enable the user to edit the content properties. When a user focuses on an input they are changing the property values. A valid keypress is the arrow keys within an input to allow the user to move the cursor.
> 
> Will the arrow keys work within the input as they normally would? Or will this PR change the behavior?
> 
> All we want to make certain of is users should be able to use the arrow keys within an input normally.

No the window listener would override the normal behaviour of CTRL+Arrow Key which would normally just skip the cursor to the next word, so you're right in thinking that. I'm just asking I'm not sure if disabling the shortcut in the input is the best solution as it makes the shortcut harder to use

I'm probably overthinking this. I'll just disable the shortcut if an input is active. Sorry about the confusion!

Fosol:
Ah I forgot about the CTRL+Arrow combination, I was thinking it was just the Arrow key.  I think that's specific enough to not have to worry about user mistyping it.  This is probably good enough in this case.

PR #1580, opened 2024-03-01
TNO-2290: Fix DateFilter behavior

mihailistov:


app/subscriber/src/components/date-filter/DateFilter.tsx
```diff
@@ -16,66 +15,83 @@ export interface IDateFilterProps {
...
-      return;
-    } else if (storedFilter.startDate !== filter.startDate) {
-      storeFilter({
-        ...filter,
-        startDate: moment(filter.startDate).startOf('day').toISOString(),
-        endDate: moment(filter.startDate).endOf('day').toISOString(),
-      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [filter]);
```
Fosol, line 35: This will fire every time the filter changes.  Which would blow away a selected endDate if they had specified one.  Is this desirable?

mihailistov: It's a bit confusing but the concept of the DateFilter means we are only filtering for one day, or one 24-hr period. So the end date will always be 23:59:59 hours ahead of the startDate whenever using a DateFilter anywhere.

Also this only does anything on initial setup, that's the only time `if (!filter.startDate)` would be true. After first load, there will always be a startDate. This is the logic we had before I rewrote it in my last PR. I changed it back since I thought it was actually a bit safer/clearer since the confusion yesterday.

PR #1582, opened 2024-03-02
feat: #tno-1729 - baseline extract quotes service

kylgarmor:
- created extract-quotes service
- wired up local deployment to docker
- wired up remote deployment to openshift
- wired up quotes to Admin/ContentController
- TODO - actually extract quotes


openshift/kustomize/services/extract-quotes/overlays/dev/config-map.yaml
```diff
@@ -0,0 +1,43 @@
+---
+kind: ConfigMap
```
Fosol, line 2: not much of an overlay here.  Does this essentially replace the configs?

services/net/extract-quotes/ExtractQuotesManager.cs
```diff
@@ -0,0 +1,287 @@
...
+using TNO.Core.Extensions;
+using TNO.Kafka;
+using TNO.Kafka.Models;
+using TNO.Services.Managers;
+using TNO.Services.ExtractQuotes.Config;
+
+namespace TNO.Services.ExtractQuotes;
+
+/// <summary>
+/// ExtractQuotesManager class, provides a Kafka Consumer service which imports audio from all active topics.
```
Fosol, line 18: Might want to update comments

services/net/extract-quotes/ExtractQuotesManager.cs
```diff
@@ -0,0 +1,287 @@
...
+    /// <param name="request"></param>
+    /// <returns></returns>
+    private async Task ProcessIndexRequestAsync(ConsumeResult<string, IndexRequestModel> result)
+    {
+        this.Logger.LogDebug($"ProcessIndexRequestAsync:BEGIN:{result.Message.Key}");
+
+        this.Logger.LogInformation("Indexing content from Topic: {Topic}, Content ID: {Key}", result.Topic, result.Message.Key);
+        var model = result.Message.Value;
+
+        if (model.Action == IndexAction.Index || model.Action == IndexAction.Publish)
```
Fosol, line 254: Is this logic relevant?

services/net/extract-quotes/ExtractQuotesManager.cs
```diff
@@ -0,0 +1,287 @@
...
+        {
+            // TODO: Failures after receiving the message from Kafka will result in missing content.  Need to handle this scenario.
+            var content = await this.Api.FindContentByIdAsync(result.Message.Value.ContentId);
+            if (content != null)
+            {
+                // Update the unpublished content with the latest data and status.
+                await ExtractQuotesAsync(content);
```
Fosol, line 261: What are you going to do after extracting the quotes?  Update the database?  Will the frontend make a separate requests for quotes?

services/net/extract-quotes/ExtractQuotesManager.cs
kylgarmor: I thought it would be.  The story says that the extract quotes should listen to the index topic.  Based on that I figured we would only attempt to extract quotes from content if the action was index or publish

Fosol: One thing to be aware of is that your extracted quotes won't show up in Elasticsearch.  Not sure if they need to be.

Fosol (approved):

services/net/extract-quotes/ExtractQuotesManager.cs
kylgarmor: Conceptually, when we extract quotes for a piece of content it would be associated with it's parent content in the database.  In order for the quotes to be indexed by elastic, we would have to trigger a re-index, which could re-trigger the quote extraction service. Might have to think about that...

If the quote extraction service is the sole source of truth for extracting/associating quotes with content, then there doesnt need to be any UI around adding quotes to content.  We just need to display the ones which were extracted.

services/net/extract-quotes/ExtractQuotesManager.cs
kylgarmor: yah I kind of talked about that in one of my other comments.  Since the quote came from the content is there a need to have a seperate index of quotes?  The business might say, I want to find a quote from an individual that contains this.  If we dont use elastic for that kind of index, it wont be as *nice*

openshift/kustomize/services/extract-quotes/overlays/dev/config-map.yaml
kylgarmor: yes, this whole PR is just setting a baseline.  I'm sure there will be config specific to the action of extracting the quotes that comes later

docker-compose.yml
```diff
@@ -433,6 +433,39 @@ services:
     networks:
       - tno
 
+####################### CoreNLP #######################
+  corenlp:
```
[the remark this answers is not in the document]
kylgarmor: yes, that was my thinking.  It's not something we are developing against or adding on top of.

libs/net/dal/Services/QuoteService.cs
```diff
@@ -0,0 +1,70 @@
...
+        return quote;
+    }
+
+    /// <summary>
+    /// Attach a quote to a piece of content.
+    /// </summary>
+    /// <param name="content"></param>
+    /// <param name="quote"></param>
+    /// <returns></returns>
+    public IEnumerable<Quote> Attach(Content content, IEnumerable<Quote> quotes)
```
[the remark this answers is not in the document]
kylgarmor: makes sense.  Just added some validation to make sure it's set correctly.

api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -384,6 +388,37 @@ public async Task<IActionResult> UploadFile([FromRoute] long id, [FromQuery] lon
...
+        if (!quotes.Any()) return new JsonResult(new { Error = "No quotes added." })
+        {
+            StatusCode = StatusCodes.Status400BadRequest
+        };
+
+        _quoteService.Attach(content, Array.ConvertAll(quotes.ToArray(), item => (Quote)item));
+
+        await _kafkaMessenger.SendMessageAsync(_kafkaHubOptions.HubTopic, new KafkaHubMessage(HubEvent.SendAll, new KafkaInvocationMessage(MessageTarget.ContentUpdated, new[] { new ContentMessageModel(content, "file") })));
+
+        return new JsonResult(new ContentModel(content, _serializerOptions));
```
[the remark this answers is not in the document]
kylgarmor: fixed this issue

PR #1583, opened 2024-03-04
TNO-2333 Ingest Manager/services override fix

fbarreta:
Allow services to user their config while sending email alerts.

### DIFF
--- a/libs/net/services/Managers/IngestManager`.cs
+++ b/libs/net/services/Managers/IngestManager`.cs
@@ -162,6 +162,7 @@ public abstract class IngestManager<TActionManager, TOption> : ServiceManager<TO
                         // Update ingest with failure.
                         await manager.RecordFailureAsync(ex);
                         this.State.RecordFailure();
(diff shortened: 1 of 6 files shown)


libs/net/services/Managers/IngestManager`.cs
```diff
@@ -162,6 +162,7 @@ public override async Task RunAsync()
                         // Update ingest with failure.
                         await manager.RecordFailureAsync(ex);
                         this.State.RecordFailure();
+                        await this.SendEmailAsync($"Ingest '{ingest.Name}' failed to run. This is failure {manager.Ingest.FailedAttempts+1} out of {manager.Ingest.RetryLimit} maximum retries.", ex);
```
Fosol, line 165: This will send an email for every failure won't it?  We only want to get an email if the failure keeps occurring after retries.

fbarreta: Ok. Got it !

Fosol (approved):

PR #1587, opened 2024-03-05
feat: #tno-2230 - admitted defeat and have used iframe

kylgarmor:


libs/net/dal/Services/ContentService.cs
```diff
@@ -299,8 +300,10 @@ public async Task<IPaged<API.Areas.Services.Models.Content.ContentModel>> FindWi
         if (!filter.IsHidden.HasValue)
             filterQueries.Add(s => s.Term(t => t.IsHidden, filter.IsHidden));
 
-        if (filter.HasTopic == true)
+        if (filter.HasTopic == true) {
```
Fosol, line 303: Your linter isn't following standard C# code layout

Fosol (approved):

libs/net/dal/Services/ContentService.cs
kylgarmor: this file wasnt meant to be part of this PR.  Attempting to undo

PR #1588, opened 2024-03-05
TNO-2369 Use new methods on dashboard

fbarreta:
As @kylgarmor suggested, I used an endpoint he already created to enable/disable the ingest.
Also I created a similar endpoint just to reset the ingest state.
It does not involve complex object into two simple operations.


libs/net/dal/Services/BaseService`.cs
```diff
@@ -77,9 +77,25 @@ public virtual TEntity Update(TEntity entity)
 
     public virtual TEntity UpdateAndSave(TEntity entity)
     {
-        var result = Update(entity);
-        this.Context.CommitTransaction();
-        return result;
+        try
```
Fosol, line 80: This isn't a safe way to handle this.  A concurrency error means the process that is asking for the update has stale data.  They should be getting the error returned to them so that they figure out what to do.  They will need to get the current data, apply their changes, and then save again.

Fosol (changes requested):

api/net/Areas/Admin/Controllers/IngestController.cs
```diff
@@ -165,6 +165,29 @@ public async Task<IActionResult> UpdateIngestEnabledStatusAsync(int id, bool new
...
+    /// <returns></returns>
+    [HttpPut("{id}/reset")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(IngestModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Ingest" })]
+    public async Task<IActionResult> UpdateIngestResetAsync(int id)
+    {
+        var model = _service.FindById(id) ?? throw new NoContentException();
+        model.State = new Entities.IngestState(model)
```
kylgarmor, line 181: you will blow away the existing `State` if you do this.  Might be better to check if the ingest has a non null `State`, and if so, just update the `State.FailedAttempts = 0`, else new up a `State` like you are doing already

api/net/Areas/Admin/Controllers/IngestController.cs
```diff
@@ -165,6 +165,29 @@ public async Task<IActionResult> UpdateIngestEnabledStatusAsync(int id, bool new
         return new JsonResult(new IngestModel(model, _serializerOptions));
     }
 
+    /// <summary>
+    /// Update ingest with the specified 'id'.
+    /// </summary>
+    /// <param name="id"></param>
+    /// <returns></returns>
+    [HttpPut("{id}/reset")]
```
kylgarmor, line 173: nit picking here, but could we change out the name `reset` here for `resetfailures`.  Contextually that's all we are resetting, not the whole Ingest.  This would apply to the method name and also to the `useApiAdminIngests`

kylgarmor (changes requested):

kylgarmor (changes requested):
ready to approve but waiting for update to tno-core 0.1.35

Fosol (approved):
Does the the `enable/disable` checkbox work correctly too?  Or does it lose the ingest configuration when it's clicked?

fbarreta:
> Does the the `enable/disable` checkbox work correctly too? Or does it lose the ingest configuration when it's clicked?

Yes, and the settings are intact after the toggle.

kylgarmor (approved):

PR #1591, opened 2024-03-06
Tno-2375

kylgarmor:
image

- show icon on row hover
- when icon is clicked, copy relevant title to clipboard
- animate icon to show something happened

### DIFF
--- a/app/editor/src/features/admin/event-of-the-day/hooks/useColumns.tsx
+++ b/app/editor/src/features/admin/event-of-the-day/hooks/useColumns.tsx
@@ -1,5 +1,6 @@
 import { Topic } from 'features/content';
 import React from 'react';
+import { FaRegClipboard } from 'react-icons/fa';
 import { Link } from 'react-router-dom';
 import Select from 'react-select';
 import {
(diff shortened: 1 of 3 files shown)


app/editor/src/features/admin/event-of-the-day/hooks/useColumns.tsx
```diff
@@ -86,20 +87,38 @@ export const useColumns = (
     }
   };
 
+  const handleCopyTitle = (event: any, cell: any) => {
```
Fosol, line 90: Should this be a `useCallback` to stop re-renders?

Fosol (approved):

app/editor/src/features/admin/event-of-the-day/hooks/useColumns.tsx
kylgarmor: I'm not sure it would trigger a re-render as there's no change in state from anything happening in the method.  If I changed it to a `useCallback` it would have no dependencies.  That's the determination isnt it?

Fosol: I don't think that's entirely correct.  I think a function will result in a new function each time it is processed.  However, there is a trade off of placing everything into memory.

PR #1598, opened 2024-03-08
fix: #tno-2384 - fix issues with series add/search

kylgarmor:
- migrated content that comes in with a new Series value will be added as `isOther`
- these items are now hidden in the Subscriber Advanced Search page
  - all Series that are isOther OR isDisabled
  - all Media Sources that are set to isDisabled
  - all Contributors that are set to isDisabled


Fosol (approved):
Thanks for cleaning up your others too

PR #1600, opened 2024-03-08
TNO-2398 Add media types on source update and insert

fbarreta:
Add Search Mapping Media types while inserting and updating Sources.

### DIFF
--- a/libs/net/dal/Services/SourceService.cs
+++ b/libs/net/dal/Services/SourceService.cs
@@ -120,6 +120,14 @@ public class SourceService : BaseService<Source, int>, ISourceService
     public override Source AddAndSave(Source entity)
     {
         entity.AddToContext(this.Context);
+        entity.MediaTypeSearchMappingsManyToMany?.ForEach(m =>
+            {
+                m.MediaType = this.Context.MediaTypes.FirstOrDefault(x => x.Id == m.MediaTypeId);
+                if (m.MediaType != null)
+                {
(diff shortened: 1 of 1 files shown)


libs/net/dal/Services/SourceService.cs
```diff
@@ -145,6 +153,13 @@ public Source UpdateAndSave(Source entity, bool updateChildren = false)
     {
         var original = FindById(entity.Id) ?? throw new NoContentException("Entity does not exist");
         this.Context.UpdateContext(original, entity, updateChildren);
+        original.MediaTypeSearchMappingsManyToMany.Clear();
```
Fosol, line 156: This is too noisy.  We have example of how to only delete the values that should be removed.

Fosol (changes requested):

libs/net/dal/Services/SourceService.cs
```diff
@@ -145,6 +154,20 @@ public Source UpdateAndSave(Source entity, bool updateChildren = false)
     {
         var original = FindById(entity.Id) ?? throw new NoContentException("Entity does not exist");
         this.Context.UpdateContext(original, entity, updateChildren);
+        var originalMedias = original.MediaTypeSearchMappingsManyToMany.ToArray();
```
Fosol, line 157: Make sure this doesn't need to be before line 156.

Fosol (approved):

Fosol (approved):

PR #1604, opened 2024-03-08
[Ready] TNO-2378: Save user's view options for content list views

jdtoombs:
In this PR:

- set view options and group options will now persist
- stores under `user.preferences`


app/subscriber/src/components/content-list/ViewOptions.tsx
```diff
@@ -1,13 +1,52 @@
...
+  const [{ userInfo }, store] = useAppStore();
+  const api = useUsers();
+
+  /** Save the user's preferences for view/grouping options under preferences */
+  const savePreferences = async () => {
+    if (userInfo) {
+      const user = {
+        ...userInfo,
+        preferences: { ...userInfo.preferences, viewOptions, groupBy },
+        isSystemAccount: false,
```
jdtoombs, line 22: I notice we hardcoded these values in `MyMinisterSettings` as well, should I add an option to `findUser` in the subscriber app so we can avoid potentially overriding these values?

jdtoombs: Not sure if this would pose further implications security wise, but they are currently required fields when updating the user.

Fosol: We shouldn't let a subscriber update these fields.  As for the model, it's a generic one we're using across the API.  You could create an appropriate model for the subscriber endpoint.

jdtoombs: I'll make a model!

jdtoombs:
@kylgarmor - this is the one with the npm changes :)

Fosol (commented):
Does this require any changes to the API endpoint and model?

jdtoombs:
> Does this require any changes to the API endpoint and model?

@Fosol - looks like we are good, response just contains target items!

Fosol:
@jdtoombs I've published `tno-core:0.1.37`.  Please update your PR and rebase with the latest changes.

jdtoombs:
> @jdtoombs I've published `tno-core:0.1.37`. Please update your PR and rebase with the latest changes.

@Fosol - updated / thanks!

Fosol (approved):

PR #1612, opened 2024-03-12
Tno 2370 add list option on media type

fbarreta:
Add a enum on MediaType entity allowing to modify filter functionalities according to this setting.
Also leaves open a possibility to modify the enum in the future to add more possibilities.


libs/net/entities/MediaType.cs
```diff
@@ -18,6 +18,12 @@ public class MediaType : BaseType<int>
     [Column("auto_transcribe")]
     public bool AutoTranscribe { get; set; }
 
+    /// <summary>
+    /// get/set - The type of entity that should be listed with this media type
+    /// </summary>
+    [Column("list_option")]
+    public ListOption ListOption { get; set; }
+
```
kylgarmor, line 26: - Just wondering if we need to have this enum set for all media types now, or whether it really only needs to be set in particular cases?  If it's only particular media types, then I would suggest tweaking the enum to the first enum value be `Default = 0` and then bump the other two values up by one.  Then this new column can be set to 0 by default.
- Should we have a migration script to set the initial value for all "known media types"  Or, if we follow my idea above, just the ones that need changing

libs/net/entities/ListOption.cs
```diff
@@ -0,0 +1,17 @@
+namespace TNO.Entities;
+
+/// <summary>
+/// Provides value types which control what users can enter for data.
+/// </summary>
+public enum ListOption
+{
+    /// <summary>
```
kylgarmor, line 8: check comment below about a `default` value

libs/net/dal/Migrations/20240311192020_1.0.130.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
+            migrationBuilder.AddColumn<int>(
+                name: "list_option",
+                table: "media_type",
+                type: "integer",
+                nullable: false,
+                defaultValue: 0);
```
kylgarmor, line 20: default value of 0 means the default behavior will be to display `Sources` ?

jdtoombs:
I have a PR with a migration coming so I'll wait for this to get in :stuck_out_tongue:  - you beat me!

libs/net/dal/Migrations/20240311192020_1.0.130.cs
Fosol: Should this be nullable?  Should it allow for both?  I can't recall how deep this rabbit hole went.

libs/net/entities/ListOption.cs
Fosol: Would nullable be a better option in this case?  Or should we stick to a default option which is essentially NA.

libs/npm/core/src/hooks/api/constants/ListOption.ts
```diff
@@ -0,0 +1,4 @@
+export enum ListOption {
+  Source = 0,
```
Fosol, line 2: Enum values are serialized as string values for the most part.  You'll need to create an additional enum `ListOptionName`

libs/npm/core/src/hooks/api/interfaces/IMediaTypeModel.ts
```diff
@@ -1,6 +1,8 @@
+import { ListOption } from '../constants';
 import { IMediaTypeSettingsModel, ISortableModel } from '.';
 
 export interface IMediaTypeModel extends ISortableModel<number> {
   autoTranscribe: boolean;
+  listOption: ListOption;
```
Fosol, line 6: This should be `ListOptionName`

libs/net/entities/MediaType.cs
fbarreta: That's gonna be the case, if we keep the Source as a default, it will keep the actual data as it is in the moment of the migration. and them in my opinion, the change to other List Option should be done by the user.

libs/net/entities/ListOption.cs
fbarreta: The default behavior we have now is that every media type, list sources. So I wouldn't like to make it nullable, since when I will handle the code to load Source or Program (or another future option if that's the case) I won't have to deal with a null state. )That's why I would choose source as default just to keep the compatibility with the actual behavior after running a migration.

Fosol: If it works I'm in favour.

libs/net/dal/Migrations/20240311192020_1.0.130.cs
fbarreta: Like mentioned above, Source is the default right now, and keeps the compatibility with the actual functionality without needing user intervention. But is the user adds a new media or update a existing one, the default value doesn't matter anymore, since it's under his control.

libs/npm/core/src/hooks/api/interfaces/IMediaTypeModel.ts
fbarreta: Fix incoming.

Fosol:
@fbarreta I've published `tno-core:0.1.40` and rebased your branch.  Please test before we merge.

Fosol (changes requested):
You need to correctly rollback your database before you fetch the latest.  Follow these steps exactly.
1) Do not fetch or rebase your branch
2) cd libs/net
3) make db-rollback n=1.0.129
4) make db-remove
5) commit your changes to your local branch
6) fetch and rebase your branch with latest
7) resolve merge conflicts (do not destroy the other 1.0.130 db migration)
8) make db-add n=1.0.132
9) make db-update
10) Test to ensure everything is working
11) Review the __EFMigrationsHistory table to ensure it does not have a duplicate 1.0.130 migration.  If it does you'll probably have to refresh your DB entirely and test again

Fosol (approved):

PR #1619, opened 2024-03-14
accessing localhost:40001 results in a 'page not found'

kkwangsir:
Since everyone is migrating using the existing machines, there should be no issues following the guidelines. However, I cleaned up images, docker cache and rebuilt the project. I discovered 2 problems, one of which can be fixed. 

After I executed 
`make setup` 
`make up n=database`
`make up n=keycloak`

and then accessed `localhost:40001`, 

it redirected to `http://localhost:40001/auth/`, showing a `page not found` error. 

image

Referring to the discussion in this thread [link](https://keycloak.discourse.group/t/keycloak-17-0-0-results-in-page-not-found-for-admin-url/13682), I modified the Docker file and then I could normally enter the Keycloak admin.

### DIFF
--- a/auth/keycloak/Dockerfile
+++ b/auth/keycloak/Dockerfile
@@ -6,6 +6,4 @@ EXPOSE 8080
 COPY ./config/*.json /tmp/
 
 # ENTRYPOINT ["tail", "-f", "/dev/null"]
-ENTRYPOINT ["/opt/keycloak/bin/kc.sh", "start-dev","--import-realm"]
-
-# ENTRYPOINT ["/opt/keycloak/bin/kc.sh", "start-dev","--import-realm","--http-relative-path=/auth"]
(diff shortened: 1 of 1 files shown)


Fosol:
This issue may be related to a cached redirect on your browser.  Can you confirm this is required if you clear local cache, or try a different browser?
The old version of Keycloak had the `/auth/...` path.  The new version does not.

jdtoombs:
> This issue may be related to a cached redirect on your browser. Can you confirm this is required if you clear local cache, or try a different browser? The old version of Keycloak had the `/auth/...` path. The new version does not.

@Fosol @kkwangsir  -  this could be an issue with the keylcoak auth server url in a `.env` file still containing `/auth` (from the old one) - I had a similar issue on my Subscriber app

kkwangsir:
It is a browser cache issue, cleaning or using another browser could solve it.

PR #1620, opened 2024-03-15
Add script for updating secret keys

kkwangsir:
Since we changed the setup process and switched to a new Keycloak image, some Keycloak account secrets need to be manually changed. This script can automatically locate the tno directory from its address, so you can execute it from anywhere and the result will be the same. It helps you update secrets in 
```
/api/net/.env, 
/tools/css-api/.env, 
/services/net/**/.env.

```
You can find secrets from `tno-service-account` in keycloak admin.

e.q.

`./tools/scripts/kc-key-update.sh xxxxxxx{secret}
`

### DIFF
--- a/tools/scripts/kc-key-update.sh
+++ b/tools/scripts/kc-key-update.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# Check if a command-line argument was provided
+if [ $# -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


jdtoombs:
Nice! I like this

Fosol (approved):

PR #1626, opened 2024-03-19
TNO-1457 Add paper totals

Fosol:
The Papers page now keeps a tally of the totals for each 'Action'.

> Please STOP using constants that are mapped to magic strings.  Use the Settings to configure primary key mappings so that we don't get frustrating regression issues and complex changes when a name is modified on an action or any other list.

## Summary

- Update Editor
- Update Subscriber
- Publish tno-core:0.1.42
- Fixed several implementation issues with the Papers page
- Refactored Action filters
- API will send out signal to hub when content is updated as part of a group action

## Papers Page

image


api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -419,6 +419,8 @@ public async Task<IActionResult> UpdateContentAsync(ContentListModel model)
         {
             foreach (var content in update)
             {
+                await _kafkaMessenger.SendMessageAsync(_kafkaHubOptions.HubTopic, new KafkaHubMessage(HubEvent.SendAll, new KafkaInvocationMessage(MessageTarget.ContentUpdated, new[] { new ContentMessageModel(content) })));
```
Fosol, line 422: We now send a message to users that content has been updated

api/net/Areas/Services/Controllers/ContentController.cs
```diff
@@ -419,8 +419,12 @@ public async Task<IActionResult> AttachQuotes([FromRoute] long id, [FromBody] IE
         if (!quotes.Any()) return new JsonResult(new ContentModel(content, _serializerOptions));
 
         _quoteService.Attach(Array.ConvertAll(quotes.ToArray(), item => (Quote)item));
-        
+
         content = _contentService.FindById(id);
+        if (content == null) return new JsonResult(new { Error = "Content does not exist" })
```
Fosol, line 424: Please handle nullable objects.  The linter always tells you about these if you use it.

app/editor/src/features/content/constants/ShowOnlyValues.ts
Fosol, line 1: This whole enum needs to be destroyed.  Stop using constants this way.

app/editor/src/features/content/papers/Papers.tsx
```diff
@@ -41,138 +41,126 @@ export interface IPapersProps extends React.HTMLAttributes<HTMLDivElement> {}
...
+            return newPage;
+          });
+        }
+      })
+      .catch(() => {});
   });
 
+  // Stores the current page
   React.useEffect(() => {
-    if (!isProcessingMessages && !delayProcessing && contentUpdatesQueue?.length) {
```
Fosol, line 95: I ripped out all of this queue stuff, as it was complicated the simple update process.  The only thing it probably was solving was sometimes more than single request is made due to updates.  This can lead to the content being stale.  I haven't seen the error occur, but I think it can.

app/editor/src/features/content/papers/constants/defaultTotals.ts
```diff
@@ -0,0 +1,11 @@
+import { ITotalsInfo } from '../interfaces';
+
+export const defaultTotals: ITotalsInfo = {
+  topStories: 0,
+  commentary: 0,
+  featuredStories: 0,
+  published: 0,
+  publishedTopStories: 0,
```
Fosol, line 8: I'm not using these presently, but it may be useful to notice if content hasn't been published.

app/editor/src/store/hooks/lookup/useSettings.ts
```diff
@@ -0,0 +1,68 @@
...
+  featuredStoryActionId?: number;
+  alertActionId?: number;
+}
+
+/**
+ * Provides a helper to centralize configuration setting values.
+ * @param validate Whether to validate the settings configurations.
+ * @returns Configuration setting values.
+ */
+export const useSettings = (validate?: boolean) => {
```
Fosol, line 20: This is the most important change.  It centralizes the logic to get specific setting values.  We need to start using this hook instead of writing the same code in multiple locations.

app/subscriber/src/store/hooks/lookup/useSettings.ts
```diff
@@ -0,0 +1,68 @@
...
+  featuredStoryActionId?: number;
+  alertActionId?: number;
+}
+
+/**
+ * Provides a helper to centralize configuration setting values.
+ * @param validate Whether to validate the settings configurations.
+ * @returns Configuration setting values.
+ */
+export const useSettings = (validate?: boolean) => {
```
Fosol, line 20: I don't like having two copies of this, one for Editor and one for Subscriber.

PR #1627, opened 2024-03-20
TNO-2295 Report instance link in email

Fosol:
The link at the bottom of a report should go to the environment the report was generated and it should load the report or report instance.

image

### DIFF
--- a/libs/net/template/Models/BaseTemplateModel.cs
+++ b/libs/net/template/Models/BaseTemplateModel.cs
@@ -11,6 +11,16 @@ public abstract class BaseTemplateModel<T> : RazorEngineTemplateBase
 {
     #region Properties
     /// <summary>
+    /// get/set - The primary key to the report.
+    /// </summary>
+    public int ReportId { get; set; }
+
+    /// <summary>
+    /// get/set - The primary key to the report instance.
+    /// </summary>
+    public long? ReportInstanceId { get; set; }
+
+    /// <summary>
     /// get/set - The main model.
     /// </summary>
     public T Content { get; set; }
--- a/services/net/reporting/ReportingManager.cs
+++ b/services/net/reporting/ReportingManager.cs
@@ -374,47 +374,51 @@ public class ReportingManager : ServiceManager<ReportingOptions>
             }
         }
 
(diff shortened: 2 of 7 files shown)


libs/net/template/Models/BaseTemplateModel.cs
```diff
@@ -10,6 +10,16 @@ namespace TNO.TemplateEngine.Models;
 public abstract class BaseTemplateModel<T> : RazorEngineTemplateBase
 {
     #region Properties
+    /// <summary>
+    /// get/set - The primary key to the report.
+    /// </summary>
+    public int ReportId { get; set; }
```
Fosol, line 16: Including these primary keys provides a way for reports to include a link to themselves.

services/net/reporting/ReportingManager.cs
```diff
@@ -374,47 +374,51 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
             }
         }
 
+        // Save the report instance.
+        // Group content by the section name.
+        var instance = new ReportInstance(
```
Fosol, line 379: Need to generate the report instance before generating the body because the template now needs the instance primary key for a link.

services/net/reporting/ReportingManager.cs
```diff
@@ -447,22 +453,20 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
...
-                model.Id = instanceModel.Id;
-                model.Version = instanceModel.Version;
-                model.Content.ForEach(ric => ric.InstanceId = model.Id);
-                model.Subject = subject;
-                model.Body = fullTextFormatBody;
                 if (request.SendToSubscribers)
-                    instance.SentOn = instance.Status == ReportStatus.Accepted ? DateTime.UtcNow : null;
-                if (model.PublishedOn == null) model.PublishedOn = DateTime.UtcNow;
-                await this.Api.UpdateReportInstanceAsync(model);
+                    instanceModel.SentOn = instanceModel.Status == ReportStatus.Accepted ? DateTime.UtcNow : null;
```
Fosol, line 467: The original code looks like it has a few bugs in it.  I hope this update corrects them and doesn't introduce regression issues.

PR #1628, opened 2024-03-20
Tno 2370 populate media sub groups according to flag

fbarreta:
- Set Media type filter tree according to the new settings at the editor.
- Fixed the filter using the media type instead a array of sources.
- Filter by Source or Program when clicking on the entity.


libs/net/models/Areas/Editor/Models/MediaType/MediaType.cs
Fosol, line 1: File name should be the same as class name.

Fosol (changes requested):

libs/net/models/Areas/Subscriber/Models/Series/SeriesModel.cs
Fosol, line 1: Same issue here

fbarreta: I think this one is OK.

Fosol (approved):

PR #1635, opened 2024-03-22
TNO-1944 Fix Editor report admin

Fosol:
Fixing issue that was resulting in report instances not saving their content.  

> Note reports that don't include content from their prior report will be empty until new content is received.  We need a better way to inform the user of this, as it just looks broken otherwise.

## Summary

- Updated Editor app
- Fixed Reporting Service

### DIFF
--- a/app/editor/src/features/admin/reports/ReportFormInstances.tsx
+++ b/app/editor/src/features/admin/reports/ReportFormInstances.tsx
@@ -22,7 +22,7 @@ export const ReportFormInstance: React.FC = () => {
 
   React.useEffect(() => {
     findInstancesForReportId(values.id).then((instances) => {
-      setInstances(instances);
+      setInstances(instances ?? []);
     });
     // Only load the report instances when clicking on the tab.
     // eslint-disable-next-line react-hooks/exhaustive-deps
@@ -84,6 +84,7 @@ export const ReportFormInstance: React.FC = () => {
           onPreview: handlePreview,
         })}
         showActive={false}
+        showPaging={false}
       />
       <Modal
         headerText="Confirm Removal"
--- a/app/editor/src/features/admin/reports/components/ReportSectionContent.tsx
+++ b/app/editor/src/features/admin/reports/components/ReportSectionContent.tsx
@@ -62,6 +62,10 @@ export const ReportSectionContent = ({ index }: IReportSectionContentProps) => {
       <Col>
         <label>Options</label>
         <FormikCheckbox
+          name={`sections.${index}.settings.hideEmpty`}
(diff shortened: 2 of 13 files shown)


app/editor/src/features/admin/reports/ReportFormInstances.tsx
```diff
@@ -22,7 +22,7 @@ export const ReportFormInstance: React.FC = () => {
 
   React.useEffect(() => {
     findInstancesForReportId(values.id).then((instances) => {
-      setInstances(instances);
+      setInstances(instances ?? []);
```
Fosol, line 25: This was resulting in null values when the API had an OOM error for some reason.

app/editor/src/features/admin/reports/components/ReportSectionContent.tsx
```diff
@@ -61,6 +61,10 @@ export const ReportSectionContent = ({ index }: IReportSectionContentProps) => {
       />
       <Col>
         <label>Options</label>
+        <FormikCheckbox
```
Fosol, line 64: Not sure why this was missing.

libs/net/dal/Services/ReportInstanceService.cs
```diff
@@ -51,7 +53,7 @@ public IEnumerable<ReportInstance> FindInstancesForReportId(int reportId, int? o
         if (ownerId.HasValue)
             query = query.Where(ri => ri.OwnerId == ownerId);
 
-        query = query.OrderByDescending(ri => ri.Id);
+        query = query.OrderByDescending(ri => ri.Id).Skip(skip).Take(take);
```
Fosol, line 56: Allow us to page through instances.  This was for some reason resulting in an OOM issue when returning hundreds of records.  We will never need to go back through history that far.  At some point we should clean these out.

services/net/reporting/ReportingManager.cs
```diff
@@ -413,10 +413,12 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
                 this.Logger.LogWarning($"Report [{report.Name}] {request.GenerateInstance}has malformed content. It will be generated, but may not match expectations.");
 
             // Update instance
-            instance.Subject = subject;
-            instance.Body = fullTextFormatBody;
-            instance.ContentManyToMany.AddRange(reportInstanceContents);
-            instance.ContentManyToMany.ForEach(ric => ric.InstanceId = instanceModel.Id);
+            instanceModel.Subject = subject;
```
Fosol, line 416: Fixed bug introduced by me a week or so ago.

PR #1636, opened 2024-03-22
TNO-2402 Schedule validations

fbarreta:
Added validation on report schedule with the following rules:

- The times fields must respect the 24 hour format
- The time fields are required when the schedule is enabled.


Fosol (commented):
If they disable the schedule and it has an invalid time, does it throw an error when saving?  It should still show a validation error before saving.

fbarreta:
> If they disable the schedule and it has an invalid time, does it throw an error when saving? It should still show a validation error before saving.

Yes ... that's the otherwise clause:
image

Fosol (approved):

PR #1643, opened 2024-03-25
Fix Indexing Service

Fosol:
Fixing Indexing Service so that it deletes content from the published index when the content has been unpublished.  Not sure why this was working before and not now?

### DIFF
--- a/services/net/indexing/IndexingManager.cs
+++ b/services/net/indexing/IndexingManager.cs
@@ -390,7 +390,7 @@ public class IndexingManager : ServiceManager<IndexingOptions>
             content = await this.Api.UpdateContentAsync(content) ?? throw new InvalidOperationException($"Content failed to update. ID:{content.Id}");
         }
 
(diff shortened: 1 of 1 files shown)


services/net/indexing/IndexingManager.cs
```diff
@@ -390,7 +390,7 @@ private async Task<ContentModel> UnpublishContentAsync(ContentModel content)
             content = await this.Api.UpdateContentAsync(content) ?? throw new InvalidOperationException($"Content failed to update. ID:{content.Id}");
         }
 
-        var document = new DeleteRequest<ContentModel>(content, this.Options.PublishedIndex, content.Uid);
+        var document = new DeleteRequest<ContentModel>(content, this.Options.PublishedIndex, content.Id);
```
Fosol, line 393: I think this is why.  I won't know until I can test in Openshift.  Locally it all appears to work.

PR #1655, opened 2024-03-28
TNO-2435 Fix Reports

Fosol:
Fixed several bugs related to reports.

## Summary

- Fixed saving report with duplicate filter
- Fixed report section 'Remove Duplicate Content' functionality

### DIFF
--- a/libs/net/dal/Services/ReportService.cs
+++ b/libs/net/dal/Services/ReportService.cs
@@ -284,6 +284,10 @@ public class ReportService : BaseService<Report, int>, IReportService
             {
                 original.Sections.Add(updatedSection);
                 this.Context.AddRange(updatedSection.ChartTemplatesManyToMany);
+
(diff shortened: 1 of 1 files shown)


libs/net/dal/Services/ReportService.cs
```diff
@@ -595,23 +599,24 @@ public IEnumerable<ReportInstance> GetLatestInstances(int id, int? ownerId = nul
                 var defaultIndex = searchUnpublished ? _elasticOptions.UnpublishedIndex : _elasticOptions.PublishedIndex;
 
                 var content = await _elasticClient.SearchAsync<API.Areas.Services.Models.Content.ContentModel>(defaultIndex, query);
+                var contentHits = content.Hits.Hits.ToArray();
```
Fosol, line 602: I'm not certain why this was resulting in failures, but this change resolve the 'Remove Duplicate Content' bug.

PR #1656, opened 2024-03-28
TNO-2433 Fix notification template

Fosol:
Updated notification template to show the Byline instead of the Published Time.

## Summary

- Add DB migration 1.0.135


libs/net/dal/Migrations/1.0.135/Up/PostUp/00-NotificationTemplate.sql
```diff
@@ -0,0 +1,60 @@
...
+}
+<div>@sourceCode (@Content.Source?.Name)</div>
+@if (!string.IsNullOrEmpty(@Content.Series?.Name))
+{
+  <div>@Content.Series.Name</div>
+}
+<div>@Content.PublishedOn?.AddHours(@utcOffset).ToString("dd-MMM-yyyy")</div>
+@if (!string.IsNullOrEmpty(@Content.Byline))
+{
+  <div>@Content.Byline</div>
```
Fosol, line 32: Display byline not the published time

PR #1660, opened 2024-04-01
fix: #tno-2480 - folders not showing content in UI

kylgarmor:
- added overload for getFolder method in FolderService which brings back folder content
- wired up new includeContent parameter in the FolderControllers
- wired up the new parameter in the NPM core package
- tweaked UI layout for folder content


Fosol (approved):
Nice to see this

PR #1672, opened 2024-04-03
TNO-2420: Add Transcript to Advanced section (Editor)

jdtoombs:
This ticket was to add Transcript, Byline, and Summary. All three were already present; however, Transcript was pointed to `story` instead of `body`


app/editor/src/features/content/constants/AdvancedSearchKeys.ts
```diff
@@ -13,4 +13,5 @@ export enum AdvancedSearchKeys {
   PublishedOn = 'publishedOn',
   CreatedOn = 'createdOn',
   UpdatedOn = 'updatedOn',
+  Body = 'body',
```
Fosol, line 16: Did you confirm that this will result in searching the body of the story?

jdtoombs: Yes I have!

Fosol (approved):

PR #1673, opened 2024-04-03
fix: #tno-2411 - fix sort order on 365 day aggregate filter

kylgarmor:
- Proactive grouping is now first


Fosol:
Remember to manually apply the db migration.

PR #1675, opened 2024-04-03
TNO 2248 new tab for reports preview

kkwangsir:
In editor | admin | report

To create a new tab after clicking "generate preview" to display the report without any parent styles.

### DIFF
--- a/app/editor/src/features/admin/reports/utils/openPreviewInNewTab.ts
+++ b/app/editor/src/features/admin/reports/utils/openPreviewInNewTab.ts
@@ -2,16 +2,54 @@ import { IReportResultModel } from 'tno-core';
 
 export const openPreviewInNewTab = (previewContent: IReportResultModel) => {
   const previewHTML = `
-      <html>
-      <head>
-        <title>Report Preview</title>
-      </head>
-      <body>
-        <h1> ${previewContent.subject}</h1>
+    <html>
+    <head>
+      <title>Report Preview: ${previewContent.subject}</title>
+      <style>
+        body { font-family: Arial, sans-serif; margin: 20px; }
+        .header { 
+          background-color: #f3f3f3; 
+          padding: 10px; 
+          text-align: center; 
+          position: relative;
+        }
+        .content { 
+          margin-top: 20px;
+          position: relative;
+        }
+        .close-btn {
+          display: inline-block; 
+          padding: 5px 15px; 
+          background-color:transparent; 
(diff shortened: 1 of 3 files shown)


app/editor/src/features/admin/reports/utils/openPreviewInNewTab.ts
```diff
@@ -0,0 +1,19 @@
...
+      <head>
+        <title>Report Preview</title>
+      </head>
+      <body>
+        <h1> ${previewContent.subject}</h1>
+        <div>${previewContent.body}</div>
+      </body>
+      </html>
+    `;
+
```
kylgarmor, line 15: This is a good idea, but I think it needs to be fleshed out a bit with something other than just the subject and the body.  Once this new window or tab is opened, the context of what it is will be lost.  I would suggest a bit more markup at the top of the preview page that informs that it's a preview for a specific report, and also some sort of action (button) that can be taken to close the window/tab that is clearly tied to "Close Report Preview" or something.

Fosol (commented):
Am I understanding this implementation correctly?  Instead of two columns it will now open a new tab?  If so, that won't suite the design.
Is this to resolve the issue that occurs when images are pushing the width out?  If so we can edit the report template.  The template will need to work in Outlook too.

app/editor/src/features/admin/reports/utils/openPreviewInNewTab.ts
```diff
@@ -0,0 +1,23 @@
+import { IReportResultModel } from 'tno-core';
+
+export const openPreviewInNewTab = (previewContent: IReportResultModel) => {
+  const previewHTML = `
+      <html>
+      <head>
+        <title>Report Preview</title>
+      </head>
+      <body>
+        <h1> ${previewContent.subject}</h1>
```
kylgarmor, line 10: maybe just prefix this element with "Report Preview: " like this?
`<h1>Report Preview: ${previewContent.subject}</h1>`
You could do the same thing with the `<title>` as well.

kkwangsir:
> Am I understanding this implementation correctly? Instead of two columns it will now open a new tab? If so, that won't suite the design. Is this to resolve the issue that occurs when images are pushing the width out? If so we can edit the report template. The template will need to work in Outlook too.

This is another issue TNO-2248, Bobbi wants this preview to open a separate browser tab. and be completely isolated.

Fosol (approved):
Nice

PR #1676, opened 2024-04-04
fix: #tno-2474 - stop log file spam

kylgarmor:
- increased default timeout for corenlp
- increased max memory for corenlp in Kustomize


Fosol (approved):
The document you send is simply the body text correct?  What is the current limitation?

PR #1678, opened 2024-04-05
TNO-2455 Fix Commentary

Fosol:
There appeared to be an odd state bug where the Commentary wasn't receiving the correct values.  I still don't understand how it was occurring, but I've moved the state to redux instead of the local hook.  Locally it appears to be working.

## Commentary

image

### DIFF
--- a/app/subscriber/src/features/commentary/Commentary.tsx
+++ b/app/subscriber/src/features/commentary/Commentary.tsx
@@ -45,14 +45,16 @@ export const Commentary: React.FC = () => {
           }),
         ),
         false,
-      ).then((res) => {
-        setCommentary(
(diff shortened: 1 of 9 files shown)


app/subscriber/src/features/commentary/Commentary.tsx
```diff
@@ -45,14 +45,16 @@ export const Commentary: React.FC = () => {
...
+      )
+        .then((res) => {
+          setCommentary(
+            res.hits.hits.map((r) => {
+              const content = r._source as IContentModel;
+              return castToSearchResult(content);
+            }),
+          );
+        })
+        .catch(() => {});
```
Fosol, line 57: All promises need a catch

PR #1680, opened 2024-04-05
TNO-2404 Fix papers sort

Fosol:
The default sort for the Editor Papers should be the `source.sortOrder` then the `page`.  Additionally, if an Editor changes the `source.sortOrder` it will resort the content in the UI to handle this scenario.

Ideally we would reindex the Source in Elasticsearch to resolve this issue, but I've created a separate story for this.

### DIFF
--- a/app/editor/src/features/content/papers/Papers.tsx
+++ b/app/editor/src/features/content/papers/Papers.tsx
@@ -27,11 +27,12 @@ import { IContentListAdvancedFilter, IContentListFilter } from '../interfaces';
 import { defaultPage } from '../list-view/constants';
 import { queryToFilter, queryToFilterAdvanced } from '../list-view/utils';
 import { ReportActions } from './components';
-import { defaultPaperFilter, defaultTotals } from './constants';
-import { useColumns, usePaperSources } from './hooks';
+import { defaultPaperFilter, defaultSort, defaultTotals } from './constants';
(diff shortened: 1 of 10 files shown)


app/editor/src/features/content/papers/Papers.tsx
```diff
@@ -67,36 +69,6 @@ const Papers: React.FC<IPapersProps> = (props) => {
 
   const selectedIds = selected.map((i) => i.id.toString());
 
-  const calcTotals = React.useCallback(
```
Fosol, line 70: Extracted to separate util function.

app/editor/src/features/content/papers/Papers.tsx
```diff
@@ -167,26 +139,33 @@ const Papers: React.FC<IPapersProps> = (props) => {
...
+          toFilter({
+            ...filter,
+          }),
+          true,
+        );
         const items = results.hits.hits
           .filter((h) => !!h._source)
-          .map((h) => castContentToSearchResult(h._source! as IContentModel));
+          .map((h) => h._source! as IContentModel)
+          .sort(sortContent(filter.sort))
```
Fosol, line 151: We now apply the sort after in the UI too, so that we handle when Elasticsearch doesn't have the latest source.sortOrder.

PR #1682, opened 2024-04-05
fix: #tno-2474 - stop log file spam

kylgarmor:
- added config to control when to extract quotes on Index actions
- changed exception handling to reduce noise
- added content id to logging to track which content is missed due to error

### DIFF
--- a/openshift/kustomize/core-nlp/base/deploy.yaml
+++ b/openshift/kustomize/core-nlp/base/deploy.yaml
@@ -30,6 +30,15 @@ spec:
     type: Rolling
   triggers:
     - type: ConfigChange
+    - type: ImageChange
+      imageChangeParams:
+        automatic: true
+        containerNames:
+          - corenlp
+        from:
+          kind: ImageStreamTag
+          namespace: 9b301c-tools
+          name: corenlp:dev
   template:
     metadata:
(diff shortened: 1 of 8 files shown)


openshift/kustomize/core-nlp/base/deploy.yaml
```diff
@@ -30,6 +30,15 @@ spec:
...
     - type: ConfigChange
+    - type: ImageChange
+      imageChangeParams:
+        automatic: true
+        containerNames:
+          - corenlp
+        from:
+          kind: ImageStreamTag
+          namespace: 9b301c-tools
+          name: corenlp:dev
```
Fosol, line 41: You will need an overlay to handle deploying to TEST and PROD.

Fosol (approved):

openshift/kustomize/core-nlp/base/deploy.yaml
kylgarmor: actually that wasnt something I meant to commit.  It would only be relevant if we introduced a build step for our customer Docker image which I dont see as neccesary.  We dont do it for Kowl or NGINX either.

Fosol (approved):

PR #1687, opened 2024-04-08
Tno 2434 new workorder call

fbarreta:
A new call to workorder was created to avoid the Having and exists clauses that were affecting the performance.
Tests in dev using the DB visualizer reduced the time from 45 seconds to less than 1 second.

Need to test in dev using the UI now.


Fosol (approved):
Looking forward to seeing this in action.

PR #1689, opened 2024-04-08
fix: #tno-1891 - add new content to the content list

kylgarmor:
- if an editor creates branch new content, add it to the content list view ignoring any filters


kylgarmor (commented):
- not sure if it's worth it to identify these new unfiltered content pieces in the list with a highlight or icon or something.  food for thought

Fosol (approved):

PR #1690, opened 2024-04-08
TNO-2434 Remove contentId check from configuration

fbarreta:
There's no need to check the contentId into configuration, since we have it on joined content.
It fixes the error when the contentId param is passed.

### DIFF
--- a/libs/net/dal/Services/WorkOrderService.cs
+++ b/libs/net/dal/Services/WorkOrderService.cs
@@ -209,7 +209,7 @@ public class WorkOrderService : BaseService<WorkOrder, long>, IWorkOrderService
             query = query.Where(c => filter.Status.Contains(c.Status));
 
         if (filter.ContentId.HasValue)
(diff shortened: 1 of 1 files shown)


Fosol (approved):

libs/net/dal/Services/WorkOrderService.cs
```diff
@@ -209,7 +209,7 @@ select new
             query = query.Where(c => filter.Status.Contains(c.Status));
 
         if (filter.ContentId.HasValue)
-            query = query.Where(c => c.ContentId == filter.ContentId || EF.Functions.JsonContains(c.Configuration, $"{{\"contentId\":{filter.ContentId}}}"));
+            query = query.Where(c => c.ContentId == filter.ContentId);
```
Fosol, line 212: This was added at one point because of historical data not including the `ContentId` property.  Instead it referenced the content in the configuration JSON object.  I suspect removing it will only impact older data.

Fosol (approved):

PR #1691, opened 2024-04-08
TNO-2471 Add Toggle component

Fosol:
Created a new Toggle component for the report section admin panels.  We can now enforce report sections only have a single data source.

## Report Data Sources

image


app/subscriber/src/features/my-reports/edit/settings/ReportEditDataSourcesForm.tsx
```diff
@@ -14,6 +16,23 @@ import * as styled from './styled';
 
 export const ReportEditDataSourcesForm = () => {
   const { values, setFieldValue } = useReportEditContext();
+  const [{ myFolders }, { findMyFolders }] = useFolders();
```
Fosol, line 19: Moved up the component tree so that it would only fire once.

PR #1706, opened 2024-04-12
Tno 2506 & 2505 remove "add to folder"

kkwangsir:
-- have props not display "add to folder"
-- refresh myfolders if folder contents length changed


Fosol (approved):
Thanks

PR #1715, opened 2024-04-17
Tno 2490 better toast 

kkwangsir:
image

Added a default icon to the TNO link component, adjusted the style to separate the elements slightly, and added an underline. Hoping to make it clear to users that this is a URL link.


Fosol (approved):
Thanks

PR #1716, opened 2024-04-17
feat: #tno-2517 - switch Event of the day report to use Topic Scores

kylgarmor:
- instead of using just count of articles, the report uses the value of the Topic.Score to define percentages


libs/net/template/ReportExtensions.cs
```diff
@@ -372,5 +372,25 @@ public static string GetColorFromName(string targetValue, string[] knownValues,
         }
         else return string.Empty;
     }
+    
+    /// <summary>
+    /// returns a the sum of all child Aggregations
+    /// </summary>
+    /// <param name="aggregations"></param>
+    /// <returns></returns>
+    public static double GetSumOfAggregations(Dictionary<string, AggregationRootModel> aggregations) {
```
Fosol, line 381: At some point we're going to have to come up with a way to include all the extension methods in a help file for template builders.

PR #1719, opened 2024-04-17
TNO-2067 Add Media Type Column on Source List

fbarreta:
Added Media Type Column to source list view.
tno-core update required.

image


Fosol (commented):
I'll publish tno-core once Jonny's PR can be merged.

Fosol (approved):

PR #1727, opened 2024-04-19
fix: #tno-2559 - several fixes for EoD report

kylgarmor:
- fixed twitchy cursor in Topic Name field when editing
- added Preview and Send buttons to bottom of EoD page
- Removed limit on number of Topics shown in Daily view chart


app/editor/src/features/admin/topics/hooks/useColumns.tsx
```diff
@@ -48,7 +48,19 @@ export const useColumns = (
                   : cell.original.name
               }
               onBlur={async () => await handleBlur(cell)}
-              onChange={async (e: any) => await handleChange(e, cell)}
+              onChange={async (e: any) => {
```
Fosol, line 51: This is one of the common issues with the virtual DOM and React state.  The solution is normally to maintain local state within the component until it's complete before applying to the context state.

Fosol: Your solution probably works okay, but I wonder if there is weird lag behaviour, or issues when typing fast.

Fosol (approved):

PR #1731, opened 2024-04-22
fix: #tno-2405 - sections now use the folder as data source

kylgarmor:
- fixed issue with 24hr sections not using the folder as soruce of truth
- updated template to be able to switch between aggregate or query to render data
- updated folderservice to return topics on content
- added more helper methods specific to Event Of The Day / Aggregates


Fosol (approved):

Fosol:
Will it survive a rollback?

Fosol (approved):

PR #1742, opened 2024-04-24
TNO-2519 Fix preview report

Fosol:
Fixed issue related to the Report Preview extending past the edge of the page.

## My Reports

image


app/subscriber/src/features/my-reports/styled/MyReports.tsx
```diff
@@ -190,7 +189,19 @@ export const MyReports = styled.div`
   }
 
   .report-preview {
-    min-width: min-content;
+    position: relative;
+    max-width: 50%;
```
Fosol, line 193: I'm not certain this is the best way to do this, but it appears to work.

PR #1744, opened 2024-04-24
TNO-2540 Fix column width

fbarreta:
Set columns width to a minimum visible width when the window is smaller.

image
Regular view

image
Smaller window

### DIFF
--- a/app/editor/src/features/content/form/constants/timeLogColumns.tsx
+++ b/app/editor/src/features/content/form/constants/timeLogColumns.tsx
@@ -5,6 +5,7 @@ import { Column } from 'react-table';
 import { ITimeTrackingForm } from '../components/time-log/interfaces';
 import { IContentForm } from '../interfaces';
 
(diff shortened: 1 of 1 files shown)


app/editor/src/features/content/form/constants/timeLogColumns.tsx
```diff
@@ -5,31 +5,36 @@ import { Column } from 'react-table';
 import { ITimeTrackingForm } from '../components/time-log/interfaces';
 import { IContentForm } from '../interfaces';
 
+const colWidth = window.innerWidth > 1500 ? 140 : 100;
```
Fosol, line 8: Can we do this with a media query?

fbarreta: I don't think so, because the class is generated by the react-table component, and I could not intercept it (and yes that was my first approach).

PR #1746, opened 2024-04-24
TNO-2508 Fix report section sorting

Fosol:
Each section in a report that displays content will now sort based on the configured option.  Also added a way to change from ascending to descending.

## Summary

- Publish tno-core:0.1.53
- Updated Editor app
- Updated Subscriber app
- Fix report section sorting

## Subscriber Edit Report

image

## Editor Report Admin

image


app/editor/src/features/admin/reports/components/ReportSectionContent.tsx
```diff
@@ -156,6 +157,17 @@ export const ReportSectionContent = ({ index }: IReportSectionContentProps) => {
               }}
             />
           </Col>
+          <Checkbox
```
Fosol, line 160: We don't have a good toggle component for the Editor.  It would would take more time an effort to add one.  The checkbox does the trick.

libs/net/dal/Services/ReportService.cs
```diff
@@ -478,23 +478,51 @@ public async Task<ReportInstance> RegenerateReportInstanceSectionAsync(int id, i
...
     /// <returns>Ordered Content</returns>
     /// <exception cref="InvalidOperationException"></exception>
-    public static ReportInstanceContent[] OrderBySectionField(IEnumerable<ReportInstanceContent> content, string? sortBy)
+    public static ReportInstanceContent[] OrderBySectionField(IEnumerable<ReportInstanceContent> content, string? sortBy, string? direction)
     {
         // If you edit this function also edit the one in the ReportEngine.OrderBySectionField.
-        return sortBy switch
+        IEnumerable<ReportInstanceContent> results;
+
+        if (direction == "desc")
```
Fosol, line 489: Ugly implementation, but LINQ doesn't seem to provide an easy switch.

libs/net/models/Areas/Services/Models/Content/ContentModel.cs
```diff
@@ -349,5 +349,15 @@ public ContentModel(Entities.Content entity, JsonSerializerOptions options) : ba
 
         return entity;
     }
+
+    public Entities.Content ToContent()
+    {
+        var content = (Entities.Content)this;
+        content.Source = this.Source != null ? (Entities.Source)this.Source : null;
```
Fosol, line 356: The default converter can't include these or they can result in bad updates in the database.

libs/net/template/ReportEngine.cs
```diff
@@ -119,22 +119,50 @@ public class ReportEngine : IReportEngine
...
+    /// <param name="direction"></param>
     /// <returns>Ordered Content</returns>
-    public IEnumerable<ContentModel> OrderBySectionField(IEnumerable<ContentModel> content, string sortBy)
+    public IEnumerable<ContentModel> OrderBySectionField(IEnumerable<ContentModel> content, string sortBy, string? direction)
     {
         // If you edit this function, also edit the related function in ReportService.OrderBySectionField
-        return sortBy switch
+        IEnumerable<ContentModel> results;
+
+        if (direction == "desc")
```
Fosol, line 129: Ugly implementation, but LINQ doesn't seem to provide an easy switch.

PR #1748, opened 2024-04-25
TNO-2510: More filter by media fixes

jdtoombs:
- can now click the label of select all
- select all now works with the alphabet filter
- fix warning for uncontrolled to controlled input

One more issue with the alphabet filter (very weird edge case) - may leave for now as it is not common, and the filter maybe changing next sprint. Creating a ticket to capture this work. Spent too long on it.


app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
```diff
@@ -185,8 +188,8 @@ export const FilterMediaLanding: React.FC = () => {
                       if (!e.target.checked) {
                         storeFilter({
                           ...filter,
-                          sourceIds: undefined,
-                          seriesIds: undefined,
```
jdtoombs, line 189: Throws warning

PR #1750, opened 2024-04-25
TNO-2526, TNO-2542: My folder fixes

jdtoombs:
- change to context provider instead of prop drilling
- fix duplicate state issue (had two state variables for the active folder and only one was in sync)
- minor useEffect fixes 

With these updates, it should get rid of all errors when trying to save folders in any combination. Previously emptying the folder / trying to rename the folder would cause an error.

### DIFF
--- a/app/subscriber/src/features/my-folders/ConfigureFolder.tsx
+++ b/app/subscriber/src/features/my-folders/ConfigureFolder.tsx
@@ -23,16 +23,16 @@ import {
 } from 'tno-core';
 
 import { getFilterOptions } from './constants';
+import { useMyFolderContext } from './MyFolderContext';
 import { Schedule } from './Schedule';
 import * as styled from './styled';
 import { createSchedule } from './utils';
 
-export interface IConfigureFolderProps {
-  active?: IFolderModel;
-}
+export interface IConfigureFolderProps {}
 
-export const ConfigureFolder: React.FC<IConfigureFolderProps> = ({ active }) => {
+export const ConfigureFolder: React.FC<IConfigureFolderProps> = () => {
   const [{ myFilters }, { findMyFilters }] = useFilters();
+  const { activeFolder: active, setActiveFolder } = useMyFolderContext();
   const [, { findContentWithElasticsearch }] = useContent();
   const [{ myFolders }, { findMyFolders, getFolder, updateFolder, deleteFolder }] = useFolders();
   const { id } = useParams();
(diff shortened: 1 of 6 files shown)


app/subscriber/src/features/my-folders/ConfigureFolder.tsx
```diff
@@ -23,16 +23,16 @@ import {
...
 import { getFilterOptions } from './constants';
+import { useMyFolderContext } from './MyFolderContext';
 import { Schedule } from './Schedule';
 import * as styled from './styled';
 import { createSchedule } from './utils';
 
-export interface IConfigureFolderProps {
-  active?: IFolderModel;
-}
+export interface IConfigureFolderProps {}
```
Fosol, line 31: I've a PR that fixes issues related to this.  While I like using contexts for shared components, I don't think that will fix the issues.

Fosol: But I'm all for it if you found it has fixed an issue.

app/subscriber/src/features/my-folders/ConfigureFolder.tsx
```diff
@@ -108,47 +107,46 @@ export const ConfigureFolder: React.FC<IConfigureFolderProps> = ({ active }) =>
...
           label="Folder name:"
-          value={currentFolder?.name ?? ''}
-          onChange={(e) =>
-            setCurrentFolder({ ...currentFolder, name: e.target.value } as IFolderModel)
-          }
+          value={active?.name ?? ''}
+          onChange={(e) => setActiveFolder({ ...active, name: e.target.value } as IFolderModel)}
         />
         <h2>Folder automation settings</h2>
         <Row>
```
Fosol, line 152: @jdtoombs The checkbox on line 161 doesn't do anything.  It needs to be hooked up to enabling the schedule.

Fosol (changes requested):
Review my PR I've just posted.  You'll need to make some changes.  I also couldn't tell how your changes would fix the save functionality?

app/subscriber/src/features/my-folders/ConfigureFolder.tsx
jdtoombs: There was a place where we had another state variable that became out of sync and I replaced that with the context provider. Going to do some testing here though to see with the new changes

app/subscriber/src/features/my-folders/ConfigureFolder.tsx
```diff
@@ -41,7 +41,6 @@ export const ConfigureFolder: React.FC<IConfigureFolderProps> = ({ active }) =>
 
   const [activeFilter, setActiveFilter] = React.useState<IFilterModel>();
   const [actionName, setActionName] = React.useState<'delete' | 'empty'>();
-  const [currentFolder, setCurrentFolder] = React.useState<IFolderModel>();
```
jdtoombs, line 44: This was the state variable becoming out of sync, and we don't really need it with context provider

app/subscriber/src/features/my-folders/ConfigureFolder.tsx
jdtoombs: I've marked it in a separate comment

PR #1764, opened 2024-04-30
TNO-2528: Basic validation for saving search

jdtoombs:
When saving search it will no longer attempt to save and will throw a toast error when:

- it is a duplicate name
- the user has not set a name yet

PS. Just having some issues with the folders area... will hopefully get that up soon

### DIFF
--- a/app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
+++ b/app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
@@ -158,6 +158,7 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ onSearch }) => 
     storeSearchFilter,
     navigate,
     updateFilter,
+    myFilters,
   ]);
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
```diff
@@ -124,6 +124,14 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ onSearch }) =>
       : { ...defaultFilter, name: searchName, query, settings };
 
     if (!filter.id) {
+      if (!searchName) {
+        toast.error('Please enter a name for your search.');
+        return;
+      }
+      if (myFilters.some((f) => f.name === searchName)) {
+        toast.error('A filter with this name already exists.');
```
Fosol, line 132: I would prefer a proper validation error, but I'll take what I can get.

Fosol (approved):

PR #1768, opened 2024-05-01
TNO-2598 Fix Reporting Service

Fosol:
Removing invalid email so that reports and notifications will not fail.  

Also corrected DevOps configuration to enable email in TEST.

### DIFF
--- a/openshift/kustomize/services/extract-quotes/overlays/dev/config-map.yaml
+++ b/openshift/kustomize/services/extract-quotes/overlays/dev/config-map.yaml
@@ -18,4 +18,4 @@ data:
   MAX_FAIL_LIMIT: "5"
   TOPICS: index
   CHES_EMAIL_ENABLED: "true"
-  CHES_EMAIL_AUTHORIZED: "false"
+  CHES_EMAIL_AUTHORIZED: "true"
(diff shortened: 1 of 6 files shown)


openshift/kustomize/services/extract-quotes/overlays/dev/config-map.yaml
```diff
@@ -18,4 +18,4 @@ data:
   MAX_FAIL_LIMIT: "5"
   TOPICS: index
   CHES_EMAIL_ENABLED: "true"
-  CHES_EMAIL_AUTHORIZED: "false"
+  CHES_EMAIL_AUTHORIZED: "true"
```
Fosol, line 21: Mistakenly set this as false in a few places.

services/net/reporting/ReportingManager.cs
```diff
@@ -703,7 +703,7 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
         await HandleChesEmailOverrideAsync(request.RequestorId);
 
         var contexts = new List<EmailContextModel>();
-        if (!String.IsNullOrWhiteSpace(request.To))
+        if (!String.IsNullOrWhiteSpace(request.To) && request.To.IsValidEmail())
```
Fosol, line 706: The downside to this is it will ignore invalid email addresses and will hide a possible issue.

PR #1775, opened 2024-05-02
fix: #tno-2608 - filter out bad data from chart data

kylgarmor:
- also remove defunct section from report


libs/net/template/ReportExtensions.cs
```diff
@@ -268,12 +268,17 @@ public static string GetContentGroupByPropertyValue(this ContentModel content, s
...
+            .Where(a => 
+                a.Topics.Any() // must have at least ONE Topic set
+                && a.Topics.All(a => a.Name != "Not Applicable" && !string.IsNullOrEmpty(a.Name)) // Name must not "Not Applicable" OR be empty
+                && a.Topics.All(t => t.Score > 0) // must have a Topic Score > ZERO
+            )
             .GroupBy(g => g.GetContentGroupByPropertyValue("topicType"))
             .ToDictionary(
                 g => g.Key,
-                g => g.ToList().Sum(s => s.Topics.FirstOrDefault()?.Score ?? 0)
+                g => g.Sum(s => s.Topics.First().Score)
```
Fosol, line 281: Won't this be more breakable than `FirstOrDefault() ?? 0`?

Fosol (approved):

libs/net/template/ReportExtensions.cs
kylgarmor: The filtering logic above it should remove any instances of content that has no `Topic`

PR #1777, opened 2024-05-03
Fix advanced search

Fosol:
I have fixed the advanced search date logic.

image

### DIFF
--- a/app/subscriber/src/components/date-filter/DateFilter.tsx
+++ b/app/subscriber/src/components/date-filter/DateFilter.tsx
@@ -28,58 +28,36 @@ export const DateFilter: React.FC<IDateFilterProps> = ({
     setOpen(false);
   }, [filter.startDate, setOpen]);
 
-  // set initial DateFilter date
-  React.useEffect(() => {
-    if (!filter.startDate) {
-      const todaysDate = moment().toISOString();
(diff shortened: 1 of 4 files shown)


app/subscriber/src/components/date-filter/DateFilter.tsx
```diff
@@ -28,58 +28,36 @@ export const DateFilter: React.FC<IDateFilterProps> = ({
     setOpen(false);
   }, [filter.startDate, setOpen]);
 
-  // set initial DateFilter date
-  React.useEffect(() => {
```
Fosol, line 32: None of this logic is required and it was causing bugs

app/subscriber/src/features/search-page/components/advanced-search/components/sections/DateSection.tsx
```diff
@@ -16,28 +16,10 @@ export const DateSection: React.FC = () => {
     { storeSearchFilter: storeFilter },
   ] = useContent();
 
-  // disable quick picker when user selects a date on react-date-picker
-  const [disableQuickPick, setDisableQuickPick] = React.useState(
```
Fosol, line 20: Why do we need to disable the offset picker?  I removed all this logic to get rid of bugs

PR #1778, opened 2024-05-03
Tno 2367 filter by media group by othersource

fbarreta:
The results with no source were removed due not being possible to group by source.
Just added a check to add other source when source does not exist.


Fosol (approved):
`content.otherSource` will always be the same as the `content.source.code` value if there is a related `source`.  The difference is that `otherSource` supports the editor creating new unique sources that don't exist in the system.

PR #1781, opened 2024-05-03
TNO-2590 Add close icon to Manage Folder

fbarreta:
image

The close button hides the Manage Folder panel.

### DIFF
--- a/app/subscriber/src/features/manage-folder/ManageFolder.tsx
+++ b/app/subscriber/src/features/manage-folder/ManageFolder.tsx
@@ -1,3 +1,4 @@
+import { Action } from 'components/action';
 import { ContentList } from 'components/content-list';
 import { reorderDragItems } from 'components/content-list/utils';
 import { PageSection } from 'components/section';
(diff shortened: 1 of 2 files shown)


app/subscriber/src/features/manage-folder/ManageFolder.tsx
```diff
@@ -121,7 +123,22 @@ export const ManageFolder: React.FC = () => {
 
   return (
     <styled.ManageFolder>
-      <PageSection header={`Manage Folder: ${folder?.name}`}>
+      <PageSection
+        header={
+          <Col className="header-col">
+            <Row className="header-row">
+              <h1 className="title">{`Manage Folder: ${folder?.name}`}</h1>
+              <FaX
```
Fosol, line 131: I have a PR coming that has an updated `Action` component which supports a `variant` property.  I'll get you to use it instead of this custom solution.  That way we'll be using the same component throughout the app.

Fosol (changes requested):
Use the updated common component `<Action variant="close"/>`

Fosol (approved):

PR #1784, opened 2024-05-03
feat: #tno-2543 - add chart for Sections over time

kylgarmor:


Fosol (approved):
Thanks!

kylgarmor:
makes it possible to plot a chart like this:
image

Requires creation of a report which has a section for each line. This report would become the data source for the chart

PR #1785, opened 2024-05-03
TNO-2610 Fix edit report content

Fosol:
This is to resolve an issue when editing report content that loses data.


app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
```diff
@@ -51,13 +51,17 @@ export const ContentForm: React.FC<IContentFormProps> = ({
           rows={1}
           disabled={disabled}
           onChange={(e) => {
+            const summary = content.versions?.[userId]?.summary ?? content.summary;
```
Fosol, line 54: Need to extract the original values if required

PR #1793, opened 2024-05-07
TNO-2511: Ability to limit subscribers media types

jdtoombs:
`this.MediaTypes = user.MediaTypesManyToMany.Select(s => s.MediaTypeId).ToArray();` this is the line I needed. Started a fresh branch and everything clicked.

In this PR:

- on the editor app an admin can now select a user then choose which media types they are not allowed to see
- now when the user navigates to filter by media, or the advanced search the corresponding media types will not be present
- `tno-core 0.1.59`

The changes were made to the `useSubMediaGroup` hook, which applies it in all areas already.


api/net/Models/Auth/PrincipalModel.cs
```diff
@@ -86,6 +86,11 @@ public class PrincipalModel
     /// </summary>
     public AccountAuthState AuthState { get; set; }
 
+    /// <summary>
+    /// get/set - An array of media types not accessible to this user.
+    /// </summary>
+    public IEnumerable<int> MediaTypes { get; set; } = Array.Empty<int>();
```
Fosol, line 92: You should also add the `SourceTypes` while you're here.  The subscriber is also limited to which sources they're supposed to be allowed to see.

Fosol (approved):

api/net/Models/Auth/PrincipalModel.cs
jdtoombs: @Fosol - updated with sources!

Fosol (approved):

PR #1796, opened 2024-05-08
TNO-2566: Evening Overview Error

areyeslo:
- Filtering null values in FileReferences before querying for content.
- New migration for Weekday and Weekend Evening Overview Templates.


libs/net/dal/Migrations/1.0.160/Up/PostUp/00-WeekdayTemplateUpdate.sql
```diff
@@ -0,0 +1,119 @@
...
+        };
+        var summary = String.IsNullOrEmpty(item.Summary) ? "---" : item.Summary.Replace("\n", "<br/>");
+        <tr style="border-bottom: 1px solid #DED9DA;">
+          <td style="border-bottom: 1px solid #DED9DA; font-size:.9rem; padding:2px 2px 2px 10px;">@placement</td>
+          <td style="border-bottom: 1px solid #DED9DA; font-size:.9rem; padding:2px 2px 2px 10px;">@(time?.Length == 8 ? time.Substring(0, 5) : time)</td>
+          <td style="border-bottom: 1px solid #DED9DA; font-size:.9rem; padding:2px 2px 2px 10px;">
+            @if (item.ContentId.HasValue)
+            {
+              <a href="@($"{ViewContentUrl}{item.ContentId}")">@summary</a>
+              @if (item.Content.FileReferences != null && item.Content.FileReferences.Any() && item.Content.FileReferences.FirstOrDefault().ContentType.StartsWith("video/") )
```
Fosol, line 79: `.FirstOrDefault()` can result in a null value.  This will fail if there is a null value.

libs/net/dal/Migrations/1.0.160/Up/PostUp/01-WeekendTemplateUpdate.sql
```diff
@@ -0,0 +1,119 @@
...
+        };
+        var summary = String.IsNullOrEmpty(item.Summary) ? "---" : item.Summary.Replace("\n", "<br/>");
+        <tr style="border-bottom: 1px solid #DED9DA;">
+          <td style="border-bottom: 1px solid #DED9DA; font-size:.9rem; padding:2px 2px 2px 10px;">@placement</td>
+          <td style="border-bottom: 1px solid #DED9DA; font-size:.9rem; padding:2px 2px 2px 10px;">@(time?.Length == 8 ? time.Substring(0, 5) : time)</td>
+          <td style="border-bottom: 1px solid #DED9DA; font-size:.9rem; padding:2px 2px 2px 10px;">
+            @if (item.ContentId.HasValue)
+            {
+              <a href="@($"{ViewContentUrl}{item.ContentId}")">@summary</a>
+              @if (item.Content.FileReferences != null && item.Content.FileReferences.Any() && item.Content.FileReferences.FirstOrDefault().ContentType.StartsWith("video/") )
```
Fosol, line 79: `.FirstOrDefault()` can result in a null value.  This will fail if there is a null value.

Fosol (changes requested):

libs/net/dal/Migrations/1.0.160/Up/PostUp/00-WeekdayTemplateUpdate.sql
areyeslo: Added `item.Content.FileReferences.Any()` to check if there is any content and just in that case obtain the first record with contentType starting with 'video'.  If `item.Content.FileReferences.Any()` returns false then it does not proceed to the rest of the conditional if statement, so it would not execute `FirstOrDefault()`. 

Please let me know if I should check for another way to implement it.

Thank you!

libs/net/dal/Migrations/1.0.160/Up/PostUp/01-WeekendTemplateUpdate.sql
areyeslo: Added `item.Content.FileReferences.Any()` to check if there is any content and just in that case obtain the first record with contentType starting with 'video'.  If `item.Content.FileReferences.Any()` returns false then it does not proceed to the rest of the conditional if statement, so it would not execute `FirstOrDefault()`. 

Please let me know if I should check for another way to implement it.

Thank you!

libs/net/dal/Migrations/1.0.160/Up/PostUp/00-WeekdayTemplateUpdate.sql
Fosol: I believe this will work.  You can also do `item.Content.FileReferences.FirstOrDefault()?.ContentType.StartsWith("video/") == true`

Fosol (approved):

libs/net/dal/Migrations/1.0.160/Up/PostUp/00-WeekdayTemplateUpdate.sql
areyeslo: Thank you!

PR #1801, opened 2024-05-10
TNO-2576 Enhance Report Charts

Fosol:
We now have very customizable charts that can be configured independently within reports and sections.

The Custom chart allows the user to configure their report section with the following features.

- Chart type
- Size of chart
- Dataset grouping
- Dataset value (count or sentiment)
- Group dataset by
- Chart layout (stacked, flip XY, show values)
- Chart scale settings (min, max, step size, min bar length)
- Legend text titles
- Legend layout
- Custom colors

## Summary

- Add DB Migration 1.0.161
- Published tno-core:0.1.60
- Updated API
- Updated Editor app
- Updated Subscriber app
- Updated Reporting Service
- Added new `Custom` chart template
- Enhanced report configuration to include settings for charts

## Editor Chart Admin

image

## Editor Report Admin

image

## Subscriber Edit Report

image


libs/net/dal/Migrations/1.0.162/Up/PostUp/00-Chart.sql
```diff
@@ -0,0 +1,47 @@
+DO $$
+BEGIN
+
+INSERT INTO public."chart_template" (
```
Fosol, line 4: New awesome configurable chart

libs/net/models/Settings/ChartSectionSettingsModel.cs
```diff
@@ -25,29 +25,74 @@ public class ChartSectionSettingsModel
...
-    /// get/set - Which property to group by [contentType|otherSource|mediaType|series|byline].
+    /// get/set - Which property to group by [contentType|otherSource|mediaType|series|byline|sentiment|sentimentSimple].
     /// </summary>
-    public string GroupBy { get; set; } = "mediaType";
+    public string GroupBy { get; set; } = "otherSource";
+
+    /// <summary>
+    /// get/set - Which property to separate datasets by [contentType|otherSource|mediaType|series|byline|section].
+    /// </summary>
+    public string Dataset { get; set; } = "";
```
Fosol, line 35: Tons of new configuration options

libs/npm/core/src/hooks/api/utils/mergeChartSettings.ts
```diff
@@ -0,0 +1,163 @@
...
+import { defaultChartSectionSettings } from '../constants';
+
+/**
+ * Generates a new chart settings configuration based on the default, current and overridden values.
+ * @param defaultChartOptions Original chart options provided by the template. Only used if 'currentSettings' is undefined.
+ * @param currentSettings The current chart settings you want to merge.
+ * @param overrideSettings The chart settings to apply and override the current chart settings
+ * @returns New chart settings.
+ */
+export const mergeChartSettings = (
```
Fosol, line 11: We now configure the chart template options when creating the report.  This may break older reports.

PR #1802, opened 2024-05-10
Tno 2622 refactor of media filtering

fbarreta:
tno-core changes:

- Allow to send an empty array to an elastic search 2, so we can filter when there is NO elements

Media filter changes:

- Refactor the select all logic
- Implement the tno-core changes toggling the empty array parameters according to the search criterias (SourceIds or Series Ids)
- Added logic to start searching after loading with the default values.


app/subscriber/src/features/filter-media/FilterMediaLanding.tsx
```diff
@@ -106,14 +135,92 @@ export const FilterMediaLanding: React.FC = () => {
...
+      sourceIds,
+      seriesIds,
+      emptyArrayTerms,
+      mediaTypeIds: [mediaGroup.key],
+    });
+    setActiveFilter(mediaGroup);
+    setParentClicked(true);
+  };
+
+  const handleClickAll = (e: any) => {
```
Fosol, line 180: Should any of these helper functions become `React.useCallback(...)`?  Currently they would result in a new function every time the code is traversed.

libs/npm/core/src/utils/elastic/generateTerms.ts
```diff
@@ -4,9 +4,15 @@
  * @param values An array of values to search for.
  * @returns An Elasticsearch query.
  */
-export const generateTerms = (field: string, values?: any[]) => {
+export const generateTerms = (field: string, values?: any[], sendEmptyArray: boolean = false) => {
```
Fosol, line 7: What is this functionality for?

Fosol:
Let me know when you're ready and I'll publish `tno-core` for you?

libs/npm/core/src/utils/elastic/generateTerms.ts
fbarreta: When we send an empty array to a term, this functions suppress the array.
Ex: ids: [] should bring nothing, but since it's suppressed, brings all.
So I added this parameter because ins some cases I really need the array to be empty, because I don't want return for that call.
I added this parameter the keep back compatibility, because some parts of our application, uses this like:
Ex: ids: [] = brings all
When I need to be specific saying:
Ex: When my field is SourceIds, pass (not suppress) an empty array.
I can make it work, and the rest of the application, is not affected.

fbarreta:
> Let me know when you're ready and I'll publish `tno-core` for you?

Ready !

libs/npm/core/src/utils/elastic/generateTerms.ts
Fosol: `ids: []` shouldn't bring all if I'm understanding this correctly.  That's telling elastic that we only want content in the array.  If the array is empty, then it shouldn't bring any content.  I think it makes more sense that if it's supposed to bring all, that it should remove `ids` from the filter before sending it.

fbarreta: That exactly what this change is from ... but ... changing the default behavior  through the whole application, is not a good idea. A lot of calls does not expect this ...
Maybe in (a near) the future we can revisit all the calls, but a change for everything would blow everything up, and probably transform this in a scope creep.
But for the media types with this PR, id; [] brings empty.

libs/npm/core/src/utils/elastic/generateQueryValues.ts
```diff
@@ -25,11 +25,23 @@ export const generateQueryValues = (
   const values = [
     generatePublishedOnQuery(settings),
     generatePostedOnQuery(settings),
-    generateTerms('sourceId', settings.sourceIds),
-    generateTerms('mediaTypeId', settings.mediaTypeIds),
-    generateTerms('seriesId', settings.seriesIds),
-    generateTerms('contributorId', settings.contributorIds),
-    generateTerms('contentType', settings.contentTypes),
+    generateTerms('sourceId', settings.sourceIds, settings.emptyArrayTerms?.includes('sourceId')),
```
Fosol, line 28: Why can you not do something simple like this on line 68?

`].filter((v) => (v !== undefined && !Array.isArray(v)) || (Array.isArray(v) && v.length > 0));`

Fosol: I can push out a test to see how this line behaves in DEV.

fbarreta: Because this on line 16 of generateTerms.ts:
return fieldValues.length > 0
    ? {
        terms: { [field]: fieldValues },
      }
    : undefined;

Would detect an Zero length array, and remove it from the ES call.

fbarreta: > I can push out a test to see how this line behaves in DEV.

We can make it on Monday morning, no need to rush. I'm sure it will work in dev, but we can re think a better approach. Because it really seems a little complicated from outsiders.

Fosol: I simply don't know why we would pass an empty array to elastic

fbarreta: > I simply don't know why we would pass an empty array to elastic

Me neither. :(

Fosol (approved):

PR #1805, opened 2024-05-13
New popout on same page for A/V content

jdtoombs:
Before we only had this semi applied to the search results, but as this is desired across the site, I have removed the prop and made it the default.

In settings the admin may define what pieces of content will have this pop out option.

image


jdtoombs:
misunderstood ticket

Fosol (approved):

PR #1806, opened 2024-05-13
prevent the second round of content fetching.

kkwangsir:
1. Set the 'active' state initial to undefined.
2. get user filterPreference and set it.
3. update sate to 'all'  if we didn't get user preferences.


app/subscriber/src/components/media-type-filters/FilterOptions.tsx
```diff
@@ -109,12 +109,24 @@ export const FilterOptions: React.FC<IMediaTypeFiltersProps> = ({ filterStoreNam
 
   const getClassName = (type: FilterOptionTypes) => (type === active ? 'active' : 'inactive');
 
+  useEffect(() => {
```
Fosol, line 112: Can you provide more information on what this is solving?  Are you essentially waiting for the userinfo to be returned from the API?

You can modify the logic to only make a request when `userInfo` is truthy.  The `useApp()` hook has the `userInfo` state.  You can wait until the `userInfo` is defined before making any requests.

Fosol (approved):

PR #1807, opened 2024-05-14
TNO-2574: MMI Alert format

areyeslo:
- Added new migration 1.0.165 to add time in the template.

On Subscriptor share a content using `SHARE` button:
image

Then, you should receive an email with the time included in the date:
image

Ticket https://apps.itsm.gov.bc.ca/jira/browse/TNO-2574


libs/net/dal/Migrations/1.0.164/Down/PostDown/00-NotificationTemplate.sql
```diff
@@ -0,0 +1,56 @@
+DO $$
```
Fosol, line 1: Down is for rollback.  Sometimes we don't need any special steps to rollback because the change is not destructive.  Is this rollback required?  It doesn't appear to do anything.

Fosol (changes requested):
Can you also bump the DB migration to `1.0.165`?  Do the following steps exactly or it will cause issues for you.
1) Do NOT get the latest code changes
2) Go to `/libs/net`
3) `make db-rollback n=1.0.163`
4) `make db-remove`
5) Commit your changes to your local branch
6) Get the latest code and merge with yours
7) `make db-update`
8) `make db-add n=1.0.165`
9) `make db-update`
10) Test

Fosol (approved):

PR #1810, opened 2024-05-14
add icon and hover effect on user name TNO-2621

kkwangsir:
Add a dropdown icon and adjust the styles to create a hover effect when the mouse over it.

image


app/subscriber/src/components/user-profile/UserProfile.tsx
```diff
@@ -1,6 +1,7 @@
 import { ScreenSizes } from 'components/layout/constants';
+import React, { useState } from 'react';
```
jdtoombs, line 2: Just have to remove useState for the linter!

kkwangsir: Thanks. I found my linter was not working, I am going fix it.

Fosol (approved):

PR #1816, opened 2024-05-15
TNO-2272 Display only source regular name

fbarreta:
After a quick review with Scott, we are just displaying source's regular names.

### DIFF
--- a/app/subscriber/src/features/hooks/useSubMediaGroups.ts
+++ b/app/subscriber/src/features/hooks/useSubMediaGroups.ts
@@ -11,13 +11,6 @@ interface ILookup {
   options: IGroupOption[];
 }
 
-const getSourceName = (source: ISourceModel) => {
-  const name = source.shortName ? source.shortName : source.name;
-  const code =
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/hooks/useSubMediaGroups.ts
```diff
@@ -57,7 +50,7 @@ export const useSubMediaGroups = (
             l.options.push({
               id: s.id,
               listOption: ListOptionName.Source,
-              name: getSourceName(s),
+              name: s.name,
```
Fosol, line 53: Are you certain this is correct?  Normally Scott wants the Common Call (`shortName`) if it's available.

fbarreta: Not now. We reviewed that and I changed to the regular name on the fly, and he wanted like this.
Probably because we have the Common Call now at the groups description on the results side.

Fosol (approved):

PR #1822, opened 2024-05-18
TNO-2643 Fix save report

Fosol:
This should resolve the missing Corporate Calendars.  Also added a way to override the exclude content report options per section.  A section can now include prior report content, which is useful for charts and folders.

## Summary

- Publish tno-core:0.1.69
- Updated API
- Updated Editor
- Updated Subscriber
- Fixed bug where unpublished content did not receive a `postedOn` value

## Subscriber Report Section Options

image


api/net/Areas/Editor/Controllers/ContentController.cs
```diff
@@ -211,9 +211,7 @@ public async Task<IActionResult> AddAsync(ContentModel model)
         var user = _userService.FindByUsername(username) ?? throw new NotAuthorizedException($"User [{username}] does not exist");
         var newContent = (Content)model;
         newContent.OwnerId = user.Id;
-        if (!newContent.PostedOn.HasValue &&
-            (newContent.Status == ContentStatus.Publish ||
-            newContent.Status == ContentStatus.Published))
+        if (!newContent.PostedOn.HasValue)
```
Fosol, line 214: We now set the value when saved no matter what.  This fixes the bug where the report option to only include newly posted content would block unpublished content filters.

libs/net/dal/Services/ReportInstanceService.cs
```diff
@@ -120,11 +120,11 @@ public ReportInstance Update(ReportInstance entity, bool updateChildren = false)
...
-            var original = this.Context.ReportInstances.FirstOrDefault(ri => ri.Id == entity.Id) ?? throw new InvalidOperationException("Report instance does not exist");
-            var originalInstanceContents = this.Context.ReportInstanceContents
-                .AsNoTracking()
-                .Include(ic => ic.Content)
-                .Where(ric => ric.InstanceId == entity.Id).ToArray();
+            var original = this.Context.ReportInstances
+                .Include(ri => ri.ContentManyToMany)
+                .ThenInclude(ric => ric.Content)
+                .FirstOrDefault(ri => ri.Id == entity.Id) ?? throw new InvalidOperationException("Report instance does not exist");
+            var originalInstanceContents = original.ContentManyToMany.ToArray();
```
Fosol, line 127: Fixed update bug

PR #1825, opened 2024-05-21
TNO-2368: Update Subscriber NPM Packages

jdtoombs:
`eslint` will have to wait to be updated as v9 conflicts with `eslint-webpack-plugin`. This will be fixed soon as mentioned in https://github.com/webpack-contrib/eslint-webpack-plugin/issues/256

The linter will fail with the new version of  `typescript` and `prettier`, so these are additional packages that we will update when said issue is resolved. Additionally there is a minor update for `react` to apply when we upgrade `typescript`. Bit of a dependency nightmare but everything is pretty much "up-to-date".

Linter still works as before - will just be on the older version until `eslint-webpack-plugin` gets the patch.


jdtoombs:
Need to downgrade a few more eslint packages

jdtoombs:
Ugh I just had it working, trying to figure out what broke it

jdtoombs:
Subscriber is complete, need to test updated package against tno-editor before publishing

Fosol:
If we remove eslint we lose linting locally and in our Github Actions.  I don't think we should move forward without it.  We already have enough issues that occur with javascript code.

app/editor/src/App.tsx
```diff
@@ -11,10 +12,21 @@ import React from 'react';
 import { BrowserRouter } from 'react-router-dom';
 import { ToastContainer } from 'react-toastify';
 import { Tooltip } from 'react-tooltip';
+import { StyleSheetManager } from 'styled-components';
 import { createKeycloakInstance, Loading, Show, useKeycloakEventHandler } from 'tno-core';
 
 const appName = 'Media Monitoring Insights';
 
+// This implements the default behavior from styled-components v5
```
jdtoombs, line 20: Gets rid of console errors, and erroneously passing custom props to the DOM

jdtoombs:
Had some stale dep version issues - had to clear yarn cache... hence the 3000 changed files

Fosol (approved):
Thanks

PR #1828, opened 2024-05-22
TNO-2671: Saved changes in search && Create new search

areyeslo:
Allow the user to save changes in the current search when modifying the settings:
image

When user clicks on Save Changes, the updates are saved in the same record:
image

When user wants to create a series of searches based on a general search, the user can modify the search and click on Save image next to the search name. For instance, the user can create an individual search for each province:
image

The records in the database include the search series:
image

Story: https://apps.itsm.gov.bc.ca/jira/browse/TNO-2671


Fosol (approved):
Looks good

PR #1829, opened 2024-05-22
TNO-2673 Remove headline line breaks to generate configuration

fbarreta:
When a transcript is requested and there is a line break on headline, the Json parser throws an exception while generating the work order configuration field.
So just to generate the configuration, all line breaks are stripped from the headline.


Fosol (approved):
Thanks

PR #1830, opened 2024-05-22
Fix My Minster

Fosol:
The My Minister page had some serious implementation issues.

- It was making multiple updates to the user preferences
- It didn't handle invalid or deleted ministers
- It will now only save when you click 'Save'

image

### DIFF
--- a/app/subscriber/src/features/settings/MyMinisterSettings.tsx
+++ b/app/subscriber/src/features/settings/MyMinisterSettings.tsx
@@ -1,75 +1,59 @@
+import { Button } from 'components/button';
 import React from 'react';
 import { toast } from 'react-toastify';
-import { useApp, useLookup, useUsers } from 'store/hooks';
-import { useAppStore } from 'store/slices';
-import { Button, Checkbox, ISubscriberUserModel, IUserInfoModel, IUserModel, Row } from 'tno-core';
+import { useLookup, useUsers } from 'store/hooks';
+import { useProfileStore } from 'store/slices';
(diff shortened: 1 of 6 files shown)


app/subscriber/src/features/settings/MyMinisterSettings.tsx
```diff
@@ -1,75 +1,59 @@
...
+        await updateUser(user);
+        toast.success(`Your minister(s) have successfully been updated.`);
+      } catch {}
+    },
+    [profile, mergeValues, updateUser],
+  );
 
   React.useEffect(() => {
-    // check if any of the users previous selections are no longer active
-    if (userInfo && userInfo?.preferences?.myMinisters?.length > 0 && activeMinisters.length > 0) {
```
Fosol, line 44: This code is problematic and not needed.

app/subscriber/src/store/hooks/subscriber/useMinisters.ts
```diff
@@ -9,23 +9,24 @@ interface IMinisterController {
...
+  const [{ ministers }, store] = useLookupStore();
 
   const controller = React.useMemo(
     () => ({
       getMinisters: async () => {
         const response = await dispatch<IMinisterModel[]>('find-all-ministers', () =>
           api.getMinisters(),
         );
-        store.storeMyMinisters(response.data);
+        store.storeMinisters(response.data);
```
Fosol, line 24: The ministers are part of the lookups currently.  No need to store them in two places,

PR #1831, opened 2024-05-22
TNO-2652 Retrieve events and schedule on load folders

fbarreta:
At the initial load, events and schedule were not loaded.
Which implied on toggle without saving, that the events were not displayed on screen, leading the user to think that was not being saved at all.
This guarantees that no matter on how it's loaded/saved/added, the UI is always accurate to the data.

### DIFF
--- a/libs/net/dal/Services/FolderService.cs
+++ b/libs/net/dal/Services/FolderService.cs
@@ -51,6 +51,7 @@ public class FolderService : BaseService<Folder, int>, IFolderService
         return this.Context.Folders
             .Include(f => f.Owner)
             .Include(f => f.ContentManyToMany)
+            .Include(f => f.Events).ThenInclude(f => f.Schedule)
             .Where(f => f.OwnerId == userId)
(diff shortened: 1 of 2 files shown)


libs/net/dal/Services/FolderService.cs
```diff
@@ -51,6 +51,7 @@ public IEnumerable<Folder> FindMyFolders(int userId)
         return this.Context.Folders
             .Include(f => f.Owner)
             .Include(f => f.ContentManyToMany)
+            .Include(f => f.Events).ThenInclude(f => f.Schedule)
```
Fosol, line 54: The problem with this approach is now every time we ask for content in a folder it will return events and schedules.  Which are not required 90% of the time.

Fosol (changes requested):

libs/net/dal/Services/FolderService.cs
Fosol: I suppose since it's just for FindMyFolders it'll be okay.

Fosol (approved):

PR #1834, opened 2024-05-23
Tno 2676 Incorrect Report Next Scheduled Send Date

kkwangsir:
The issue with an "invalid date" has two reason:

Since we choose schedule generate report but doesn't auto send,  the variable `sends` must be empty. We haven't handled this situation yet.

on other hand , due to various reasons, our datetime composition may get errors, resulting in an invalid date.
I use moment-timezone to avoid it.

Therefore, when there's no automatic next report schedule, I had a new text message.

image


Fosol (approved):
Thanks

PR #1840, opened 2024-05-24
TNO-266: Update content list 

jdtoombs:
Debugging weird console error when using `alt-text` on an `` tag that also has a tooltip. But code should stay the same.

In this PR:

- new "pop-out button"
- new inline play button
- new transcript symbol
- new report pin
- specific attributes based on media types, also responsive for mobile
- several styling changes
image


app/subscriber/public/assets/report_pin.svg
Fosol, line 1: Why include this when we already have `<FaThumbtack/>`?

app/subscriber/src/components/content-list/Attributes.tsx
```diff
@@ -0,0 +1,64 @@
...
+  margin?: boolean;
+}
+export const Attributes: React.FC<IAttributesProps> = ({
+  item,
+  showDate,
+  showTime,
+  viewOptions,
+  margin,
+}) => {
+  const noByLine = ['Talk Radio', 'News Radio', 'Events'];
```
Fosol, line 28: This will break as soon as an admin changes the names of the media types.

app/subscriber/src/components/content-list/ContentReportPin.tsx
```diff
@@ -34,11 +34,16 @@ export const ContentReportPin = ({ contentId }: IContentReportPinProps) => {
     );
   }, [contentId, myReports, reportContent]);
 
-  if (!reports.length) return <></>;
+  if (!reports.length) return <div className="icon-placeholder" />;
 
   return (
     <styled.ContentReportPin className="content-report-pin">
-      <FaThumbtack data-tooltip-id={`report-pin-${contentId}`} />
+      <img
```
Fosol, line 41: Why are we switching to an image instead of the icon?

app/subscriber/src/components/content-list/ContentRow.tsx
```diff
@@ -58,65 +66,85 @@ export const ContentRow: React.FC<IContentRowProps> = ({
...
-        </Show>
-        <Show visible={!!activeStream && item.id === activeStream.id}>
-          <FaEyeSlash
-            className="eye-slash"
-            onClick={(e) => {
-              e.stopPropagation();
-              setActiveStream({ source: '', id: 0 });
-              setActiveFileReference(undefined);
-            }}
+        {!!width && width > 768 ? (
```
Fosol, line 79: why can't we use media queries for this stuff?

app/subscriber/src/components/content-list/ContentReportPin.tsx
jdtoombs: SVG provided by Bobbi

app/subscriber/src/components/content-list/ContentRow.tsx
jdtoombs: Mainly because I want to control the DOM and not just hide/show via CSS. Lower down there is another conditional check that adds an additional `Row` when viewing in mobile

app/subscriber/src/components/content-list/Attributes.tsx
jdtoombs: Will rewire with settings, forgot to change this

app/subscriber/src/components/content-list/ContentReportPin.tsx
jdtoombs: (prob doesn't matter if I leave it as you had it)

app/subscriber/src/components/content-list/ContentRow.tsx
jdtoombs: I could rewire with media query, I just find this logic easier to follow as nothing is hidden in the CSS

jdtoombs: Will rewrite to make consistent

app/subscriber/public/assets/report_pin.svg
jdtoombs: Just to match Bobbi's new mockup with the given SVGs

app/subscriber/src/components/content-list/ContentRow.tsx
Fosol: You don't need to use media query if you're hiding elements.  I'm fine with it.  I was just asking.

PR #1849, opened 2024-05-26
TNO-2639: Update core + few more package updates

jdtoombs:
Was able to get these working by leaving `eslint` related packages completely untouched. From what I've read `eslint v9` is causing a lot of headache for people and not worth touching right now.

Also gave another stab at updating to `React 18` - same issue with infinite loop. Just an earlier error message to stop it thanks to a `keycloak-js` update.

**`CRA` does not play well with the newest version of TypeScript. Wondering if we will have to migrate to something like `Vite` at one point**


app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { useApp } from 'store/hooks';
 import { Col, ContentTypeName, IContentModel, Loading, Show, TextArea, Wysiwyg } from 'tno-core';
 
-export interface IContentFormProps extends React.HTMLAttributes<HTMLDivElement> {
+export interface IContentFormProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'content'> {
```
jdtoombs, line 5: Guessing this is due to `react-dom` updates. `content` is now used

Fosol (approved):

PR #1850, opened 2024-05-26
make search bold work, and fix formatSearch function

kkwangsir:
image
Because the default search did not enable the function `formatSearch` of bold keyword , I applied it to the relevant text and modified the logic that may encounter null value errors


Fosol (commented):
Nice work.  Just need to rebase and fix branch history.

PR #1855, opened 2024-05-27
TNO-2667,2690: Update grouped by date display, and fix filter option sync issue

jdtoombs:
Small changes to update the grouped by date display, and additionally fix a bug that causes the api to fetch before the user's filter preference is grabbed. 

I added the active filter to a context provider for the landing page to use, this way it will not fetch until the active filter is set based on the userPrefs, which should eliminate the race condition that was present when using `userInfo` as the ready state.

Also some padding fixes for the search page.

image

### DIFF
--- a/app/subscriber/src/components/content-list/utils/groupContent.ts
+++ b/app/subscriber/src/components/content-list/utils/groupContent.ts
@@ -39,38 +39,35 @@ export const groupContent = (groupBy: IGroupByState, content: IContentSearchResu
   const grouped = content
     .sort(sortFunc(firstSort))
     .sort(sortFunc(secondSort))
-    .reduce(
-      (acc, item) => {
-        let key = item.source?.name; // default to source
-        if (!item?.source?.name) {
-          key = item.otherSource;
-        }
-        switch (groupBy) {
-          case 'time':
-            const date = new Date(item.publishedOn);
-            const hours = date.getHours().toString().padStart(2, '0');
-            const minutes = date.getMinutes().toString().padStart(2, '0');
-            key = `${moment(date).format('DD-MMM-YYYY')} ${`(${hours}:${minutes})`}`;
-            break;
-          case 'source':
-            if (!item?.source?.name) {
-              key = item.otherSource;
-            } else {
-              const shortName =
-                item.source.shortName && item.source.shortName !== ''
-                  ? ` (${item.source.shortName})`
-                  : '';
-              key = `${item.source.name}${shortName}`;
-            }
-            break;
-        }
-        if (!acc[key]) {
-          acc[key] = [];
-        }
-        acc[key].push(item);
-        return acc;
-      },
-      {} as Record<string, IContentSearchResult[]>,
-    );
+    .reduce((acc, item) => {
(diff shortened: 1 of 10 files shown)


app/subscriber/src/components/content-list/utils/groupContent.ts
```diff
@@ -49,7 +49,7 @@ export const groupContent = (groupBy: IGroupByState, content: IContentSearchResu
           const date = new Date(item.publishedOn);
           const hours = date.getHours().toString().padStart(2, '0');
           const minutes = date.getMinutes().toString().padStart(2, '0');
-          key = `${moment(date).format('DD/MM/YY')} (${hours}:${minutes})`;
+          key = `${moment(date).format('DD-MMM-YYYY')} ${`(${hours}:${minutes})`}`;
```
Fosol, line 52: Why not `format('DD-MMM-YYYY hh:mm')`?

app/subscriber/src/features/home/Home.tsx
```diff
@@ -45,7 +46,7 @@ export const Home: React.FC = () => {
   React.useEffect(() => {
     // stops invalid requests before filter is synced with date
     // wait for userinfo incase applying previously viewed filter
-    if (!!featuredStoryActionId && !!userInfo) {
+    if (!!featuredStoryActionId && !!active) {
```
Fosol, line 49: userInfo would ensure the user has been activated.  Is there anything related to user preferences in this request?

jdtoombs: `active` relies on `userInfo` in `FilterOptions.tsx`, the previous fix was working locally, but on dev there was some sort of weird race condition. Active will become truthy once `userInfo` is parsed and the active filter is set

app/subscriber/src/components/content-list/utils/groupContent.ts
jdtoombs: Just in case we eventually add an option to remove time, or if certain content types don't need time when grouped by content. I know it was that way at one point so I just left it easy to change!

app/subscriber/src/features/home/Home.tsx
jdtoombs: I could keep userInfo in there in the check though, just had to add active to get rid of that weird behaviour I was observing earlier

jdtoombs: After reading this again, I think what would be better is adding `hasProcessedInitialPreference` to a context provider, rather than the string state. Will update

Fosol: Just make sure you're not adding a secondary "user info has been loaded", as it'll make things inconsistent and more difficult to maintain.
`userInfo` when truthy is all we need to know to determine whether the user has logged in, we have activated them, and we've pulled back the latest info for them.

jdtoombs: i.e) Filter has selected the appropriate className to mark as active, and has the state synced up to the user preferences

Fosol (approved):

PR #1863, opened 2024-05-28
TNO-2641 Add chart font config

Fosol:
Added additional configuration options to control the font size of chart text and legends.

## Summary

- Published tno-core:0.1.78
- Updated Editor app
- Updated Subscriber app
- Updated API

## Editor Chart Template Admin

image

## Editor Report Section Admin

image

## Subscriber Report Template Admin

image


openshift/kustomize/postgres/crunchy/base/deploy.yaml
```diff
@@ -48,6 +48,13 @@ spec:
           requests:
             storage: 7Gi
         storageClassName: netapp-block-standard
+      # walVolumeClaimSpec:
```
Fosol, line 51: Tried this but it didn't work.  It immediately filled up every WAL volume.

PR #1872, opened 2024-05-29
TNO-2655 New welcome screen

fbarreta:
image
New screen was created with a more detailed description.
It also shows the loading spinner after 5 seconds to keep the user "entertained".


Fosol (approved):
Thanks

PR #1878, opened 2024-05-30
Fix DB migration 1.0.173

Fosol:
Before merging/rebasing this code into your branch make sure you rollback.

`make db-rollback n=1.0.172`

Then you can merge/rebase the this code.

`make db-update`


libs/net/dal/Migrations/20240529201121_1.0.173.cs
```diff
@@ -15,7 +15,7 @@ protected override void Up(MigrationBuilder migrationBuilder)
             migrationBuilder.RenameColumn(
                 name: "Format",
                 table: "user_report",
-                newName: "format");
+                newName: "format2");
```
Fosol, line 18: Could not change case with SQL.

PR #1879, opened 2024-05-30
TNO-2651Update report instance content inline string

kkwangsir:
image

I've got confirmation, so I don't need to sort through AV content, which simplifies the issue. I just need to create a standalone display for AV content, with the format: Source - Program (Series) - Type - DateTime.

### DIFF
--- a/app/subscriber/src/features/my-reports/utils/reportInstanceContentInlineDisplay.ts
+++ b/app/subscriber/src/features/my-reports/utils/reportInstanceContentInlineDisplay.ts
@@ -1,0 +1,26 @@
+import moment from 'moment';
+import { ContentTypeName } from 'tno-core';
+
+import { IReportInstanceContentForm } from '../interfaces';
+
+export const reportInstanceContentInlineDisplay = (row: IReportInstanceContentForm) => {
+  if (!row.content) return '';
+  if (row.content.contentType === ContentTypeName.AudioVideo) {
+    // AV Content format
+    console.log('xxx1', row.content.series);
+    return (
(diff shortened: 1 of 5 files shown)


app/subscriber/src/features/my-reports/utils/reportInstanceContentInlineDisplay.ts
```diff
@@ -0,0 +1,26 @@
+import moment from 'moment';
+import { ContentTypeName } from 'tno-core';
+
+import { IReportInstanceContentForm } from '../interfaces';
+
+export const reportInstanceContentInlineDisplay = (row: IReportInstanceContentForm) => {
```
Fosol, line 6: Make this a component instead of a function since it is returning output.

app/subscriber/src/features/my-reports/utils/reportInstanceContentInlineDisplay.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (!row.content) return '';
+  if (row.content.contentType === ContentTypeName.AudioVideo) {
+    // AV Content format
+    console.log('xxx1', row.content.series);
```
Fosol, line 10: Remove

Fosol (changes requested):

app/subscriber/src/features/my-reports/utils/reportInstanceContentInlineDisplay.ts
kkwangsir: I remember our lint flagged the log, but anyway, I will ensure next time it won't be in the code.

Fosol (approved):
Nice

PR #1887, opened 2024-05-31
TNO-2694 Add exclude empty values

Fosol:
Report charts can now exclude empty values when the dataset or group by options result in content that has an empty value.

## Summary

- Publish tno-core:0.1.82
- Updated Editor app
- Updated Subscriber app
- Updated API
- Updated Reporting service

## Subscriber Exclude Empty Values

image


app/editor/src/features/admin/reports/components/ReportOptions.tsx
```diff
@@ -0,0 +1,52 @@
+import { Col, FormikCheckbox, Row } from 'tno-core';
+
+export const ReportOptions = () => {
```
Fosol, line 3: Moved these options into a separate component to make it easier to maintain.

libs/net/template/ReportExtensions.cs
```diff
@@ -558,12 +558,14 @@ public static bool IsPrivate(this ContentModel content, ReportEngineContentModel
...
 
+        var excludeEmptyValues = settings?.ExcludeEmptyValues ?? false;
+
         var groups = groupBy switch
         {
             "mediaType" => content.GroupBy(c => c.MediaType?.Name ?? "Other").OrderBy(group => group.Key),
             "contentType" => content.GroupBy(c => c.ContentType.ToString()).OrderBy(group => group.Key),
-            "byline" => content.GroupBy(c => String.IsNullOrWhiteSpace(c.Byline) ? "Unknown" : c.Byline).OrderBy(group => group.Key),
-            "series" => content.GroupBy(c => c.SeriesId?.ToString() ?? "NA").OrderBy(group => group.Key),
+            "byline" => content.GroupBy(c => String.IsNullOrWhiteSpace(c.Byline) ? "Unknown" : c.Byline).Where((g) => excludeEmptyValues ? g.Key != "Unknown" : true).OrderBy(group => group.Key),
```
Fosol, line 567: It only works for byline and series currently.  Most of the other options are required values.

PR #1891, opened 2024-05-31
Fix ResizeObserver loop bug

Fosol:
This is a workaround to fix the ResizeObserver infinite loop bug introduced by react-tooltip and content on the page.

image


app/subscriber/src/features/utils/fixTooltipResizeObserver.ts
```diff
@@ -0,0 +1,27 @@
...
+      const ctx = self;
+      tid && clearTimeout(tid);
+      tid = setTimeout(() => {
+        callback.apply(ctx, args);
+      }, delay);
+    };
+  };
+
+  const _ = (window as any).ResizeObserver;
+  (window as any).ResizeObserver = class ResizeObserver extends _ {
```
Fosol, line 21: Essentially override the behaviour to force a debounce.

PR #1893, opened 2024-06-01
TNO-2728=MMI-Advanced Search Problems

areyeslo:
Summary stores the transcription for Content Type Audio/Video and sometimes summary is different to body. We need to show the summary and body when they are different.

- Remove HTML tags, square brackets and brake lines before comparison
- Check if the difference in length is greater than 10%
- If summary and body are different then show both of them:

image

Story: https://apps.itsm.gov.bc.ca/jira/browse/TNO-2728

### DIFF
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -195,11 +195,30 @@ export const ViewContent: React.FC<IViewContentProps> = ({ setActiveContent }) =
       wo.userNotifications?.some((un) => un.userId === profile?.id),
   );
 
+  //Remove HTML tags, square brackets and line breaks before comparison.
+  const cleanString = (str: string | undefined) => str?.replace(/<[^>]*>?|\[|\]|\n/gm, '').trim();
+
+  //Difference ratio
+  const threshold = 0.1;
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -195,11 +195,30 @@ export const ViewContent: React.FC<IViewContentProps> = ({ setActiveContent }) =
...
   );
 
+  //Remove HTML tags, square brackets and line breaks before comparison.
+  const cleanString = (str: string | undefined) => str?.replace(/<[^>]*>?|\[|\]|\n/gm, '').trim();
+
+  //Difference ratio
+  const threshold = 0.1;
+
+  //Return true if difference between length of str1 & str2 is greater than 10%
+  const isDifferent = (str1: string | undefined, str2: string | undefined) => {
```
Fosol, line 205: When a function is not memoized `useCallback(...)` it can result in a lot of extra renders.  It's generally always best to extract a utility function if it doesn't require access to local values, or add it to a `useCallback(...)` if it does require local values.

Fosol (changes requested):

Fosol (approved):

PR #1904, opened 2024-06-05
TNO-2734 subscriber modal update styling

nehalaggarwal-bcgov:
TNO-2734 changed modal source to components/modal instead of tno-core in subscriber for most files.
Also change in gen-env-files.sh file to remove the extra ".

### DIFF
--- a/tools/scripts/gen-env-files.sh
+++ b/tools/scripts/gen-env-files.sh
@@ -775,7 +775,7 @@ Service__ApiUrl=http://host.docker.internal:$portApi/api
 
 # Elastic__Url=host.docker.internal:$portElastic
 ELASTIC_USERNAME=$elasticUser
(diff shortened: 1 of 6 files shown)


Fosol:
You'll need to fix the build issues before this can be merged.

tools/scripts/gen-env-files.sh
```diff
@@ -775,7 +775,7 @@ Service__ApiUrl=http://host.docker.internal:$portApi/api
 
 # Elastic__Url=host.docker.internal:$portElastic
 ELASTIC_USERNAME=$elasticUser
-ELASTIC_PASSWORD=$password" >> ./services/net/folder-collection/.env
```
Fosol, line 778: This will break the script

Fosol (changes requested):

Fosol (approved):

PR #1905, opened 2024-06-05
TNO-2744: Small enhancements again

jdtoombs:
- added a loading indicator to the appropriate section to give the user some feedback when the request is in process 
- removed unnecessary dom clutter
- slight styling changes
- active report selected will now have icon next to it even when not hovering

### DIFF
--- a/app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
+++ b/app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
@@ -114,24 +114,24 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content, onCl
                   <FaFileExport className="report-icon" />
                   {report.name}
                 </Row>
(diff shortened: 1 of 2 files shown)


app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -20,6 +20,7 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content, onCl
   const [{ requests }] = useApp();
   const [activeReport, setActiveReport] = React.useState<IReportModel>();
   const [reportId, setReportId] = React.useState<number | null>(null);
+  const [inProgress, setInProgress] = React.useState({ sectionName: '', value: false });
```
jdtoombs, line 23: So we can display a loading indicator next to the section that the user is adding content to

Fosol (approved):

PR #1912, opened 2024-06-06
TNO-2734 subscriber modal update styling

nehalaggarwal-bcgov:
TNO-2734 Changes to ShareMenu and Modal files to add enableConfirm.

### DIFF
--- a/app/subscriber/src/components/modal/Modal.tsx
+++ b/app/subscriber/src/components/modal/Modal.tsx
@@ -27,6 +27,8 @@ export interface IModalProps {
   hasHeight?: boolean;
   /** Form is submitting and buttons should be disabled. */
   isSubmitting?: boolean;
+  /** Checking condition for enabling Confirm button*/
+  enableConfirm?: boolean;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


app/subscriber/src/components/modal/Modal.tsx
```diff
@@ -44,6 +46,7 @@ export const Modal: React.FC<IModalProps> = ({
   customButtons,
   type,
   hasHeight,
+  enableConfirm,
```
Fosol, line 49: This property won't do anything.  How should it be implemented below?

Fosol (changes requested):

app/subscriber/src/components/modal/Modal.tsx
nehalaggarwal-bcgov: It is being used in the ShareMenu.tsx file in the Share Content Modal as: enableConfirm={content.length > 0 && (user?.colleague !== undefined || emailAddress !== '')}

Fosol: Without any logic below, it doesn't make sense to have this property.  Check the Modal in tno-core to see how it was implemented.

fbarreta (approved):

Fosol (approved):

PR #1914, opened 2024-06-06
Tno 2657 prevent sending empty report

kkwangsir:
Check empty Front Page Images report before send, and make toast messages.


Fosol (approved):

app/editor/src/features/content/papers/PaperToolbar.tsx
Fosol, line 1: I would have though the preview will have some HTML still, but no content?

kkwangsir: image

yes, I copied part of  preview details.

`{
    "subject": "\n",
    "body": "\n",
`

Fosol: This doesn't look correct to me.  Make certain your report template is valid.

kkwangsir: image
Yes, You are right. I found I only have three. Should I create Front Pages report manually or import from somewhere?

Fosol: You can probably copy it from the TEST environment manually.  Not sure why that report doesn't exist in our local setup.  You may want to fix that issue to.

kkwangsir: Thanks. I copied it into my machine. and change the setting default front page report id,

Fosol: The issue I see here is that the report that is generated will still have HTML, it simply will not have any content.  Which will make the below code fail.

kkwangsir:
image
image
I spend some time to reviewing the backend code and and made modifications to parse the report preview data, which includes content results. I make a few tests using both empty and populated reports, and the system appears to be functioning correctly, even when handling reports that contain HTML but lack actual content

app/editor/src/features/content/papers/PaperToolbar.tsx
```diff
@@ -41,18 +42,76 @@ export interface IPaperToolbarProps {
...
+  const [isLoading, setIsLoading] = React.useState(false);
+
   React.useEffect(() => {
     setMediaTypeOptions([new OptionItem<number>('Any', 0), ...mediaTypeOptions]);
   }, [mediaTypeOptions]);
 
+  async function checkReportContent(reportID: number) {
+    try {
+      setIsLoading(true);
+      const preview = await previewReport(reportID);
```
Fosol, line 62: Can you check if the report has the message in it "There is no content in this report."?

kkwangsir: OK. No problem.

Fosol (approved):

PR #1926, opened 2024-06-10
TNO-2710 No reason to display otherSource at this point. confuses the client.

kkwangsir:
When the user clicks on the option to turn off the display of the source, the extension looks for the otherSource, which is basically the same code as the source.code. This can be confusing.

### DIFF
--- a/libs/net/template/ReportExtensions.cs
+++ b/libs/net/template/ReportExtensions.cs
@@ -367,9 +367,13 @@ public static class ReportExtensions
             ? $"{content.Source?.ShortName}"
             : "";
 
-        var source = (context.Settings.Headline.ShowSource
-                ? content?.Source?.Name
-                : "");
+        var source = "";
+        if (context.Settings.Headline.ShowSource)
+        {
+            source = !string.IsNullOrEmpty(content?.Source?.Name)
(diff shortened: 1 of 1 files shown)


libs/net/template/ReportExtensions.cs
```diff
@@ -370,12 +370,6 @@ public static string GetSource(this ContentModel content, ReportEngineContentMod
         var source = (context.Settings.Headline.ShowSource
                 ? content?.Source?.Name
                 : "");
-        // source is empty or empty string, use the other source
-        if (String.IsNullOrWhiteSpace(source))
```
Fosol, line 374: The problem here is that `OtherSource` should be displayed when content does not have a `Source`.  The reason `OtherSource` exists is because editors can add content that doesn't not belong to any of the currently configured sources.

Fosol: Wouldn't we then do the following?
When `Show Source == false` do not show `Content.Source.Name` or `Content.OtherSource`.
When `Show Source == true` show `Content.Source?.Name ?? Content.OtherSource`.

Fosol (approved):

PR #1944, opened 2024-06-13
TNO-2720: Allow for impersonation / updating of my ministers 

jdtoombs:
In this PR:

- users who are able, may now impersonate another user and update their my minister settings
- when viewing My Ministers you will see it as the impersonated user
- same with My Minister settings, you may see their selection and update it for them


app/subscriber/src/features/impersonation/Impersonation.tsx
```diff
@@ -95,7 +95,7 @@ export const Impersonation = () => {
             <Col flex="1">Username</Col>
             <Col flex="1">Email</Col>
             <Col flex="1">First Name</Col>
-            <Col flex="1">Last Name</Col>
+            <Col flex="1">Last Name</Col>Impersonate
```
Fosol, line 98: Was this intentional?

jdtoombs: Good catch, I have no idea how that happened... removed

Fosol (approved):

PR #1956, opened 2024-06-18
TNO-2757 Wildcard implementation

fbarreta:
The method simple_query_string doesn't support wildcards, but instead of switching it all, I made a check on search criteria to call the query_string method. Just because the query_string is less performant than the simple_query_string.

The only wildcards supported by query_string are ? and *

So we will call the method on demand.

image
With wildcards

image
Without wildcards


Fosol (commented):
I have published `tno-core:0.1.91`

Fosol (approved):

PR #1957, opened 2024-06-19
TNO-2720: Impersonation for my minster fix

jdtoombs:
`updateUsers` had a `storeMyProfile` which was overwriting the `profile.preferences.impersonate`, this will fix that allowing the conditionals to be correct in a `useEffect` after updating a user. 

I am not sure why it was working locally before this change...

Key changes:

- update user now takes an optional prop to check if it is being used in an impersonation or not so the user's profile will not get overwritten by the wrong information

### DIFF
--- a/app/subscriber/src/store/hooks/subscriber/useUsers.ts
+++ b/app/subscriber/src/store/hooks/subscriber/useUsers.ts
@@ -42,7 +42,16 @@ export const useUsers = (): IUserController => {
         return response.data;
       },
     }),
-    [dispatch, getUser, findUsers, storeMyProfile, userInfo, storeUserInfo, updateUser],
+    [
+      dispatch,
+      getUser,
+      findUsers,
+      storeMyProfile,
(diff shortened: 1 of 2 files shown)


app/subscriber/src/store/hooks/subscriber/useUsers.ts
```diff
@@ -33,11 +33,11 @@ export const useUsers = (): IUserController => {
         const response = await dispatch('find-users', () => findUsers(filter));
         return response.data;
       },
-      updateUser: async (model: ISubscriberUserModel) => {
+      updateUser: async (model: ISubscriberUserModel, impersonate?: boolean) => {
         const response = await dispatch<ISubscriberUserModel>('update-user', () =>
           updateUser(model),
         );
-        storeMyProfile(response.data);
```
jdtoombs, line 40: This was causing the issue, updated most of the store logic except for this guy

Fosol (approved):

PR #1964, opened 2024-06-20
Tno 2758 display old content on search

fbarreta:
image
Date range

image
Offset

Implemented 7 day older search, for offset cases, and also for date range cases.


Fosol (commented):
I find this implementation difficult to understand.  Can you explain the flow?  
The way I thought it would work was if the redux store had prior results, it would display them.  If not, it would make a request for content and then store the results in the redux store.

fbarreta:
> I find this implementation difficult to understand. Can you explain the flow? The way I thought it would work was if the redux store had prior results, it would display them. If not, it would make a request for content and then store the results in the redux store.

Sure. It's implemented the way it was on FilterMedia. (and it was a little bit hard to implement, but let's go ...)

Offset:
I make a call for "Today" so the previous behavior would request a offset of 1 day.
The actual changes add 7 days to the offset, so in reality, it brings the results of the last 8 days.
Having it's results, we make a filter for "Today, using the offset number, so we only display the results that have Date = Today (currDateResults array), and the previous results are stored at the prevDateResults array.
If there are no results on currDateResults we display  the previous results bellow grouped by day.
If the offset is bigger than 1 (48 hours = 2, 1 month = 30, and so on ...) it keeps adding 7 days, and just change the date rage to use in the filter.

Date Rage:
This was already implemented, I just needed to make some adjustments to not step into the offset call boundaries.
But it works the same way, but just subtracts 7 days from initial date, allowing to get 7 days older content, and uses the same principle to display the current (currDateResults) and the previous (prevDateResults) results.

Fosol:
Ah, I was thinking of a different story.  This story is for showing older content if there is no content for the current request.  Thanks!

Fosol (approved):

PR #1967, opened 2024-06-20
TNO-2786: Advanced Search including columnist name

areyeslo:
- Added columnist name to search in query to elasticsearch when the name is checked.

Story: https://apps.itsm.gov.bc.ca/jira/browse/TNO-2786


areyeslo:
Implemented different logic. This implementation included columnist name in the search terms, but this was getting undesired content.

PR #1969, opened 2024-06-21
TNO-2781 Add National Observer

Fosol:
Updated File Monitor Service to handle other XML feeds.  Corrected implementation issues with XPath.

This will need an update to the Ingest configuration in DEV, TEST, PROD.

### DIFF
--- a/app/editor/src/features/admin/ingests/configurations/Newspaper.tsx
+++ b/app/editor/src/features/admin/ingests/configurations/Newspaper.tsx
@@ -80,15 +80,15 @@ export const Newspaper: React.FC = (props) => {
     <styled.IngestType>
       <ImportContent />
       <Row>
-        <Col flex="1 1 0">
+        <Col flex="1">
           <FormikCheckbox label="Escape Content" name="configuration.escapeContent" />
           <FormikCheckbox label="Add Parent" name="configuration.addParent" />
           <FormikCheckbox label="Fix Blacks Newsgroup XML" name="configuration.fixBlacksXml" />
         </Col>
-        <Col flex="1 1 0"></Col>
+        <Col flex="1"></Col>
       </Row>
       <Row>
-        <Col flex="1 1 0">
+        <Col flex="1">
           <FormikSelect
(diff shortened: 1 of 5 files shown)


app/editor/src/features/admin/ingests/configurations/Newspaper.tsx
```diff
@@ -148,15 +148,33 @@ export const Newspaper: React.FC = (props) => {
         entry.
       </p>
       <Row>
-        <Col flex="1 1 0">
+        <Row flex="1">
```
Fosol, line 151: Namespaces are required for some XML parsing

libs/net/models/Extensions/IngestModelExtensions.cs
```diff
@@ -141,8 +141,8 @@ public static T GetConfigurationValue<T>(this IngestModel ingest, string key, T
                     }
             },
             JsonValueKind.True or JsonValueKind.False => (T)Convert.ChangeType($"{element.GetBoolean()}", typeof(T)),
-            JsonValueKind.Object => throw new NotImplementedException(),
-            JsonValueKind.Array => throw new NotImplementedException(),
+            JsonValueKind.Object => (T)Convert.ChangeType($"{element.GetRawText()}", typeof(T)),
+            JsonValueKind.Array => (T)Convert.ChangeType($"{element.GetRawText()}", typeof(T)),
```
Fosol, line 145: Returns basic string.

services/net/content/ContentManager.cs
```diff
@@ -550,6 +550,7 @@ private async Task<ContentModel> CopyFileWithSSHAsync(DataLocationModel dataLoca
         var password = connection.GetConfigurationValue("password");
         var keyFileName = connection.GetConfigurationValue("keyFileName");
         var path = connection.GetConfigurationValue("path");
+        var port = connection.GetConfigurationValue<int?>("port");
```
Fosol, line 553: We now correctly handle port values.

PR #1972, opened 2024-06-24
TNO-2617 Automatically set Contributor based on Byline

laidaoyu:
- updated content manager to find matched contributor by the ingested content Byline field
- updated Editor Contributor Model: added Aliases and added convert to Contributor method


Fosol (approved):
Looks good

PR #1977, opened 2024-06-25
TNO-2768 Fix distribution list

Fosol:
Reduce chance of duplicate emails being used for emails.


services/net/contentmigration/appsettings.json
```diff
@@ -93,6 +93,8 @@
         "The Northern View": "NV",
         "The Tri-City News": "TCN",
         "Victoria Times-Colonist": "TC",
+        "Victoria Times Colonist": "TC",
+        "Times Colonist": "TC",
```
Fosol, line 97: I hope this will resolve the issue where content from TNO is not getting put into the correct source.

PR #1978, opened 2024-06-26
TNO-2812 Fix front page images

Fosol:
Fix issue that was not loading front page images for prior dates.


app/subscriber/src/features/todays-front-pages/TodaysFrontPages.tsx
```diff
@@ -17,20 +16,27 @@ export const TodaysFrontPages: React.FC = () => {
     },
     { findContentWithElasticsearch, storeFrontPageFilter: storeFilter },
   ] = useContent();
-  const [, { getFilter }] = useFilters();
-  const { frontpageFilterId } = useSettings(true);
+  const { frontPageImageMediaTypeId } = useSettings(true);
```
Fosol, line 19: We cannot use the front page images filter because it is specific for the current date.  Instead we'll generate our own with the foreign key to the correct media type.

PR #1982, opened 2024-06-26
Disable transcription for Events / Update press gallery 

jdtoombs:
In this PR:
- new setting that allows you to put the id's for media types that you want to disable transcription on
- ensure attributes for content display the same as rest of site on press gallery


Fosol (changes requested):
I don't think any of this was required based on the comment I added in the story.  All events will currently be in the "Media Availability" source.  You can already disable these through the Source Admin.

image

If you want to provide the functionality to disable transcripts based on Media Type, then you should do this the same way we do it for Sources.

PR #1983, opened 2024-06-26
My Report UI change

kkwangsir:
The new Report UI modifications are almost complete. However, regarding the request for subscription, I will handle it separately in another pull request.
The main changes include the new section for `subscribers` under the setting, as well as the addition of `preview & send` and a `history` section.
There have also been various minor and major modifications.

A. `Subscribers`
image
B. `preview & send`
image

C. `History`
image


app/subscriber/src/features/my-reports/edit/settings/styled/ReportEditSubscribersForm.tsx
```diff
@@ -0,0 +1,46 @@
+import styled from 'styled-components';
+
+export const ReportEditSubscribersForm = styled.div`
+  display: flex;
+  flex-direction: column;
+  margin: 1rem;
+  gap: 1rem;
+  .select-column {
+    width: 5rem !important;
```
Fosol, line 9: Is there a way we can do this without using `!important`?  Generally this only needed when we're using a 3rd party component that wasn't written well.

Fosol (commented):
Great work.  Looking forward to seeing these improvements.  Only issue I saw was the css comment I added.

Fosol (approved):

PR #1984, opened 2024-06-26
Use split query

Fosol:
Improve the performance of the query so that it doesn't stop working.


services/net/contentmigration/ContentMigrationAction.cs
```diff
@@ -93,7 +93,8 @@ public class ContentMigrationAction : IngestAction<ContentMigrationOptions>
             .Where(predicate)
             .OrderBy(ni => ni.UpdatedOn) // oldest first
             .ThenBy(ni => ni.ItemTime) // oldest first
-            .ThenBy(ni => ni.RSN);
+            .ThenBy(ni => ni.RSN)
+            .AsSplitQuery();
```
Fosol, line 97: This is recommended for performance, however it can result in data issues.

services/net/contentmigration/sources/oracle/Entities/BaseNewsItem.cs
```diff
@@ -27,25 +27,25 @@ public abstract class BaseNewsItem
     /// <summary>
     /// get/set.
     /// </summary>
-    [Column("SOURCE")]
+    [Column("SOURCE", TypeName = "VARCHAR2")]
```
Fosol, line 30: Recommended by some to improve performance.

PR #1988, opened 2024-06-26
TNO-2764 adding loading indicator for slow network issues

nehalaggarwal-bcgov:
TNO-2764 adding loading indicator for Top Stories and Press Gallery initially, using the same loading component from Evening Overview.


Fosol (approved):

Fosol:
@nehalaggarwal-bcgov you'll need to resolve the failed check.

PR #1997, opened 2024-06-28
TNO-2786: Search for byline using contributor name.

areyeslo:
- Adding  `should` (OR) in the Elasticsearch query to include the contributor name to be included in  any of the fields: headline, byline, summary, body:

image


libs/npm/core/src/hooks/api/interfaces/IContentModel.ts
```diff
@@ -37,6 +37,7 @@ export interface IContentModel extends IAuditColumnsModel {
   seriesId?: number;
   series?: ISeriesModel;
   contributorId?: number;
+  contributorName?: string;
```
Fosol, line 40: We don't want this here.  It's already in the `contributor.name`

libs/npm/core/src/utils/elastic/generateQueryValues.ts
```diff
@@ -28,13 +29,16 @@ export const generateQueryValues = (
     generateTerms('sourceId', settings.sourceIds),
     generateTerms('mediaTypeId', settings.mediaTypeIds),
     generateTerms('seriesId', settings.seriesIds),
-    generateTerms('contributorId', settings.contributorIds),
+    // Following line replaced by generateShouldQuery:
+    // generateTerms('contributorId', settings.contributorIds),
```
Fosol, line 33: Do we actually want to remove this?  When searching for `contributorId` I would think this would still be the right thing.

areyeslo: If the contributorId is not in content.contributor_id, we would not get any result. For instance, if we have John Doe as byline but it is not included in content.contributor_id, we would not get the record, it is handled as `AND` statement. Should we handle as 'OR' as we are doing with byline?

Fosol: If the user is making a search for content with a single `contributorId` then it should be required.  If they are searching for multiple then it would be a `should` (or).

areyeslo: At the moment the line `generateTerms('contributorId', settings.contributorIds),` is generating a list of IDs:
```
{
                    "terms": {
                        "contributorId": [
                            6,
                            8,
                            9
                        ]
                    }
                },
```
I would need to add the logic you mention.

libs/npm/core/src/utils/elastic/generateQueryShouldValues.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Generates an Elasticsearch query based on specified 'query'.
+ * @param settings Form values that will be used to configure the elasticsearch query.
+ * @param query Original query object.
+ * @returns Elasticsearch query JSON.
+ */
+export const generateQueryShouldValues = (
+  settings: Omit<IFilterSettingsModel, 'size' | 'searchUnpublished'>,
+): any[] => {
+  const matches = generateMatch('byline', settings.contributorNames);
```
Fosol, line 12: This isn't going to work correctly.  I reviewed this morning.

The `content.byline` and `content.contributor.name` is not always going to be the same.  We are adding code to apply the contributor based on the byline, but this isn't the inverse also.  If an editor selects the contributor we don't apply it to the byline, nor do we currently want to.

If an Advanced Search selects a contributor you will need to search the `content.byline` for the name, and the `content.contributorId` for the selected contributor.  If they only select one contributor it must find them and will be an AND.  If they select multiple, then it will be an OR/SHOULD.

Also when the user has selected the Byline checkbox to search for keywords,  you will need to search the `content.contributor.name` for any of the keywords.

Fosol (changes requested):

libs/npm/core/src/utils/elastic/generateQueryShouldValues.ts
areyeslo: I changed the logic to cover your comments. Please take a look. Thank you!

app/editor/package.json
```diff
@@ -59,7 +59,7 @@
     "redux-logger": "3.0.6",
     "styled-components": "6.1.11",
     "stylis": "4.3.2",
-    "tno-core": "0.1.100"
```
Fosol, line 62: I'll need to publish a new package for this.

Fosol (approved):

PR #2002, opened 2024-07-03
Report preview fix

kkwangsir:
Summary:

1. Hide some buttons in bottom action bar for 'Preview&Send' and 'History'

image

2. Init Preview Report and always display the Preview Report in the right panel..
3. Considering report updateOn, tracking this data value with hook, so the behavior is regenerate report only when report got update. This could reduce times of request from user, save server resources


areyeslo (approved):
LGTM

PR #2003, opened 2024-07-03
TNO-2867 Avoid undefined of publishedOn, tno-core update

kkwangsir:
This change requires to bump tno-core 

Modify the logic of URL generation, if `publishedOn` is null or undefined, we shouldn't append any query string, and the old way gives an undefined follow by ` includeContent=true`

image


jdtoombs (approved):

jdtoombs:
@kkwangsir - Hey, I published `tno-core 0.1.97`, you'll just have to update your package.json's in the editor and subscriber with the new version number!

Fosol (approved):

Fosol (changes requested):
Update with latest tno-core version published by Jonny

Fosol:
Going to close so that I can create a new PR and get this out now.

PR #2010, opened 2024-07-08
TNO-2879 Fix do not send report

Fosol:
The Do Not Send option for reports only handle one logic path, but would not work for auto reports.

### DIFF
--- a/services/net/reporting/ReportingManager.cs
+++ b/services/net/reporting/ReportingManager.cs
@@ -492,14 +492,14 @@ public class ReportingManager : ServiceManager<ReportingOptions>
             var responseModel = new ReportEmailResponseModel();
             try
             {
-                if ((linkOnlyFormatTo.Any() || linkOnlyFormatCC.Any() || linkOnlyFormatBCC.Any()) && String.IsNullOrEmpty(request.To))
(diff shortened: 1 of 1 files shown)


services/net/reporting/ReportingManager.cs
```diff
@@ -492,14 +492,14 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
             var responseModel = new ReportEmailResponseModel();
             try
             {
-                if ((linkOnlyFormatTo.Any() || linkOnlyFormatCC.Any() || linkOnlyFormatBCC.Any()) && String.IsNullOrEmpty(request.To))
```
Fosol, line 495: Required to not send email out for an auto report.

services/net/reporting/ReportingManager.cs
```diff
@@ -629,14 +629,14 @@ private async Task GenerateReportAsync(ReportRequestModel request, API.Areas.Ser
             var responseModel = new ReportEmailResponseModel();
             try
             {
-                if (request.SendToSubscribers && (linkOnlyFormatTo.Any() || linkOnlyFormatCC.Any() || linkOnlyFormatBCC.Any()) && !report.Settings.DoNotSendEmail)
+                if (!report.Settings.DoNotSendEmail && request.SendToSubscribers && (linkOnlyFormatTo.Any() || linkOnlyFormatCC.Any() || linkOnlyFormatBCC.Any()))
```
Fosol, line 632: Move to front to speed up conditional check.

PR #2011, opened 2024-07-09
 TNO-2821 use filter option all as page's default filter.

kkwangsir:
As the ticket request, we use the All option as the default filter when entering a list page. 
image


areyeslo (approved):
LGTM

PR #2012, opened 2024-07-09
TNO-2764 duplicate summaries partial fix

nehalaggarwal-bcgov:
TNO-2764 duplicate summaries partial fix just shows summary if it exists.


areyeslo (commented):
Can we remove the comments or Do we need them?

app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -32,7 +32,7 @@ import { ToneValue } from './ToneValue';
 import { isWorkOrderStatus } from './utils';
 
 //Difference ratio
-const threshold = 0.1;
+//const threshold = 0.1;
```
jdtoombs, line 35: Looks good - Quick question, do we still need the commented out code? Here and below?

nehalaggarwal-bcgov: Yes, I will be working on these later on to check if the summary is different from transcript.

areyeslo (approved):

app/subscriber/src/features/content/view-content/ViewContent.tsx
areyeslo: You'll need to remove them in the following PR.

PR #2014, opened 2024-07-09
TNO-2772: More user feedback for no results

jdtoombs:
Very small change to add a toast when no results come back, similar to when there are too many.

image


areyeslo (approved):
LGTM

PR #2015, opened 2024-07-09
TNO-2816: Paper sort order not working

areyeslo:
Having back the use of `naturalSortValue` in Papers.

image


jdtoombs (approved):
looks good!

PR #2016, opened 2024-07-10
TNO-2690: Filter options not correctly remembering / applying when navigating back to site

jdtoombs:
Little fix for a the bug that doesn't apply the user's preferred filter on the content list views.


areyeslo (approved):
LGTM

PR #2017, opened 2024-07-10
Improve user admin

Fosol:
Indirect users will now require a username.  This is to support accounts Scott uses to manage related reports.  Ownership of various objects has been updated to only allow Direct or Indirect users.  Impersonation has been updated to only display Direct or Indirect users.  The API now supports searching for an array of account types.  Note the change to the user search will now return system accounts, which may not be ideal in some cases.

## Summary

- Publish tno-core:0.1.98
- Updated Editor app
- Updated Subscriber app
- Updated API


libs/npm/core/scripts/publish.sh
```diff
@@ -10,7 +10,7 @@ package_name=$(node -p "require('./package.json').name")
 package_version=$(node -p "require('./package.json').version")
 
 info=$()
-current_version=$(yarn npm info tno-core | grep -oP '(?<="customfield_11500": ")[^"]*' -)
+current_version=$(yarn npm info tno-core version)
```
Fosol, line 13: Fixed script so that it works on Macos

app/subscriber/src/components/action/styled/Action.tsx
```diff
@@ -52,5 +52,9 @@ export const Action = styled.div<IActionProps>`
     min-width: ${(props) => props.size};
     color: ${(props) =>
       !props.disabled ? props.theme.css.iconPrimaryColor : props.theme.css.iconGrayColor};
+
+    &:hover {
```
Fosol, line 56: Minor styling fix for hover on a disabled component.

PR #2028, opened 2024-07-14
Add Transfer User Account

Fosol:
Added the new feature to transfer or copy a user account.  This is a functional Pull Request, it does not implement the UI as designed by Bobbi.  The UI can be updated when we have time.

A Direct user now has an Account Mirroring section.  You can select the user account you want to mirror and choose between "transfer" or "copy".  Transfer assigns the ownership of the user account objects (reports, notifications, filters, folders).  Copy makes new copies and assigns them to the user.

## Summary

- Publish tno-core:0.1.100
- Updated API
- Updated Editor app
- Fixed several bugs
- Resolved Mac OS issues

## Account Mirroring

image


libs/npm/core/src/components/formik/wysiwyg/FormikContentWysiwyg.tsx
```diff
@@ -1,7 +1,5 @@
 import 'react-quill/dist/quill.snow.css';
 
-import _ from 'lodash';
```
Fosol, line 3: Fixed lint issue

PR #2029, opened 2024-07-15
Anonymous view report page

Fosol:
The link in the report email will now allow any user anonymously to view the report.  This isn't an ideal solution, but will work for us temporarily.  There isn't any PIA issues.

image

### DIFF
--- a/api/net/Areas/Subscriber/Controllers/ReportInstanceController.cs
+++ b/api/net/Areas/Subscriber/Controllers/ReportInstanceController.cs
@@ -1,6 +1,7 @@
 using System.Net;
 using System.Net.Mime;
 using System.Text.Json;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
 using Swashbuckle.AspNetCore.Annotations;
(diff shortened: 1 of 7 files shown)


api/net/Areas/Subscriber/Controllers/ReportInstanceController.cs
```diff
@@ -173,17 +174,18 @@ public IActionResult Delete(ReportInstanceModel model)
     /// <param name="regenerate"></param>
     /// <returns></returns>
     [HttpPost("{id}/view")]
+    [AllowAnonymous]
```
Fosol, line 177: Allow anonymous access to this endpoint.

app/subscriber/src/features/router/AppRouter.tsx
```diff
@@ -20,26 +20,47 @@ import { Claim, InternalServerError, NotFound } from 'tno-core';
...
-    if (!authenticated && !window.location.pathname.startsWith('/login'))
+    if (
+      !authenticated &&
+      !window.location.pathname.startsWith('/login') &&
+      !window.location.pathname.startsWith('/report/instances')
+    )
       navigate(`/login?redirectTo=${window.location.pathname}`);
   }, [authenticated, navigate]);
 
+  if (!authenticated)
```
Fosol, line 43: Need different routes for unauthenticated users.

PR #2030, opened 2024-07-15
Fix notifications

Fosol:
There was a logic bug that would result in missing notifications when there was more than one matching notification configured.


services/net/notification/NotificationManager.cs
```diff
@@ -360,6 +364,21 @@ private async Task ProcessNotificationAsync(ConsumeResult<string, NotificationRe
         }
     }
 
+    /// <summary>
+    /// The alert has been sent, remove the flag from the content so that future alerts can be sent.
+    /// </summary>
+    /// <param name="content"></param>
+    private async Task ResetAlertAsync(API.Areas.Services.Models.Content.ContentModel content)
```
Fosol, line 371: This was changing the content before other notifications were validated.

PR #2031, opened 2024-07-16
Fix auto convert

Fosol:
Auto conversion of .mov files would incorrectly interrupt the next snippet.  This fix should resolve that issue.  Also updated .NET packages to resolve security issue.

## Summary

- Update .Net packages in all projects
- Updated Editor app

### DIFF
--- a/api/net/Areas/Editor/Models/Storage/DirectoryModel.cs
+++ b/api/net/Areas/Editor/Models/Storage/DirectoryModel.cs
@@ -83,7 +83,7 @@ public class DirectoryModel
     /// </summary>
     /// <param name="files"></param>
     /// <param name="isLocal"></param>
-    public DirectoryModel(IEnumerable<Renci.SshNet.Sftp.SftpFile> files, bool isLocal = false)
+    public DirectoryModel(IEnumerable<Renci.SshNet.Sftp.ISftpFile> files, bool isLocal = false)
     {
         var result = new List<ItemModel>();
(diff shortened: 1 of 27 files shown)


api/net/Areas/Editor/Models/Storage/DirectoryModel.cs
```diff
@@ -83,7 +83,7 @@ public DirectoryModel(IEnumerable<string> files, bool isLocal = false)
     /// </summary>
     /// <param name="files"></param>
     /// <param name="isLocal"></param>
-    public DirectoryModel(IEnumerable<Renci.SshNet.Sftp.SftpFile> files, bool isLocal = false)
+    public DirectoryModel(IEnumerable<Renci.SshNet.Sftp.ISftpFile> files, bool isLocal = false)
```
Fosol, line 86: Fixed issue with upgraded library.

app/editor/src/features/content/form/hooks/useContentForm.ts
```diff
@@ -199,6 +199,8 @@ export const useContentForm = ({
   React.useEffect(() => {
     if (!!id && +id > 0) {
       fetchContent(+id);
+    } else if (!id || +id === 0) {
+      setForm((values) => ({ ...defaultFormValues(values.contentType) }));
```
Fosol, line 203: Need to reset the form values when going to the next snippet action.

PR #2034, opened 2024-07-16
TNO-2764 duplicate summaries transcript

nehalaggarwal-bcgov:
TNO-2764 added function isDifferent to check if the summary and transcript are the same.


Fosol (approved):
Thanks!

PR #2035, opened 2024-07-16
TNO-2878: Filter by media type is still not returning content based on Type?

areyeslo:
Currently, when searching by Media Type field in subscriber, there is a mapping with source. As a result, if a content is created including Media Type `Online` and source `BC 1`, there is no content returned when querying by Media Type `Online` because sourceId is not in `source_media_type_search_mapping` table.

The user is expecting to obtain content for the Media Type selected even though the source is not mapped to the mediaType. There are some special MediaTypes that keep the original mapping: CP Wire, Events, Corporate Calendar, AV Archive

Story: https://citz-gdx.atlassian.net/browse/TNO-2878

## Summary

- Publish tno-core:01.1.107
- Add DB migration 1.1.1


app/subscriber/src/features/hooks/useSubMediaGroups.ts
```diff
@@ -26,6 +26,16 @@ export const useSubMediaGroups = (
 } => {
   const [subMediaGroups, setSubMediaGroups] = useState<ISubMediaGroupItem[]>([]);
   const [{ userInfo }] = useApp();
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const MediaTypeAllSources = [
```
Fosol, line 30: This will break once the media types are changed with the Editor admin.  Can we get this list from the lookups?

areyeslo: We are creating mediaTypes in DB using  [01-MediaType.sql migration](https://github.com/bcgov/tno/blob/dev/libs/net/dal/Migrations/1.0.114/Up/PostUp/01-MediaType.sql) and creating mapping using [00-MediaTypeSearchMapping.sql](https://github.com/bcgov/tno/blob/dev/libs/net/dal/Migrations/1.0.95/Up/PostUp/00-MediaTypeSearchMapping.sql) 

Then, we are retrieving all mediaTypes and sources using [useLookup](https://github.com/bcgov/tno/blob/dev/app/subscriber/src/features/filter-media/FilterMediaLanding.tsx#L22). This call gives us all records for mediaTypes and sources. Then, [useSubmBediaGroups](https://github.com/bcgov/tno/blob/dev/app/subscriber/src/features/filter-media/FilterMediaLanding.tsx#L23) hook takes this as input to assign which sources are mapped to specific MediaType.

The only way I see the mediaType can be modified is through another migration as I can see. I checked in the editor side and we do not have the option to modify a mediaType. 

Not sure what I am missing.

Fosol: https://dev.editor.mmi.gov.bc.ca/admin/media-types

Fosol: Also each source can be mapped to a media types through the source admin page. https://dev.editor.mmi.gov.bc.ca/admin/sources/261

Fosol: And the Series can be mapped to media types through the show/program admin page.  https://dev.editor.mmi.gov.bc.ca/admin/programs/788

areyeslo: Added the Settings configuration for MediaTypes that do not follow the mapping and query just by Media Type.  

Ready for your review.

app/subscriber/src/features/filter-media/FilterMedia.tsx
```diff
@@ -4,8 +4,8 @@ import { DateFilter } from 'components/date-filter';
 import { ContentListActionBar } from 'components/tool-bar';
 import { IContentSearchResult } from 'features/utils/interfaces';
 import React from 'react';
-import { useContent } from 'store/hooks';
-import { generateQuery, IContentModel, Loading, Show } from 'tno-core';
+import { useContent, useLookup } from 'store/hooks';
```
Fosol, line 7: Do not use the `useLookup` for settings.  Use the `useSettings` hook so that we have consistent behaviour in all places.

Fosol (changes requested):

app/subscriber/src/features/filter-media/FilterMedia.tsx
areyeslo: Thank you! I am ready for your review :)

Fosol: The PR still is not using the correct hook.

areyeslo: sorry about that, somehow I lost my changes. I am applying again. I will let you know. Thank you!

app/subscriber/src/features/filter-media/FilterMedia.tsx
```diff
@@ -4,7 +4,7 @@ import { DateFilter } from 'components/date-filter';
 import { ContentListActionBar } from 'components/tool-bar';
 import { IContentSearchResult } from 'features/utils/interfaces';
 import React from 'react';
-import { useContent } from 'store/hooks';
+import { useContent, useSettings } from 'store/hooks';
```
areyeslo, line 7: Change to useSettings. Added to settingsSlice the new field mediaTypesIdsAllSources to be saved in the state.

Fosol (commented):
I had to publish the correct DB migration.

Fosol (approved):

PR #2041, opened 2024-07-17
TNO-2885 Fix missing content in report

Fosol:
The report `PublishedOn` date was incorrectly being converted to local time when creating a filter to ensure only newly posted content was included in the next report.

I've also added the ability to pass a reference to several components.  Currently it doesn't do anything, but the goal was to limit the event handler related to the content form page.  I have also made it so that the shortcuts continue to perform their default behaviour on the page.  I did this because it was extremely annoying not being able to refresh or copy anything.

### DIFF
--- a/app/editor/src/features/content/form/ContentActions.tsx
+++ b/app/editor/src/features/content/form/ContentActions.tsx
@@ -40,206 +40,201 @@ export interface IContentActionCheckbox {
   value: boolean;
 }
 
-export const ContentActions: React.FC<IContentActionsProps> = ({
-  name = 'actions',
-  init,
-  filter,
-  addRowOn,
-}) => {
-  const { values, setFieldValue } = useFormikContext<IContentForm>();
-  const [{ actions, holidays }] = useLookup();
-  const { featuredStoryActionId, commentaryActionId, alertActionId } = useSettings();
-  const { field } = useNamespace(name);
+export const ContentActions = React.forwardRef<HTMLDivElement, IContentActionsProps>(
+  ({ name = 'actions', init, filter, addRowOn }, ref) => {
+    const { values, setFieldValue } = useFormikContext<IContentForm>();
+    const [{ actions, holidays }] = useLookup();
+    const { featuredStoryActionId, commentaryActionId, alertActionId } = useSettings();
+    const { field } = useNamespace(name);
 
-  const formActions: IContentActionModel[] = getIn(values, name, []);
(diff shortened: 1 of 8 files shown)


libs/net/dal/Extensions/JsonDocumentExtensions.cs
```diff
@@ -58,7 +58,7 @@ public static JsonDocument IncludeOnlyLatestPosted(this JsonDocument query, Date
         var json = JsonNode.Parse(query.ToJson())?.AsObject();
         if (json == null || previousInstancePublishedOn == null) return query;
 
-        var jIncludeOnlyLatestPosted = JsonNode.Parse($"{{ \"range\": {{ \"postedOn\": {{ \"gte\": \"{previousInstancePublishedOn.Value.ToLocalTime():yyyy-MM-ddTHH:mm:ss}\", \"time_zone\": \"US/Pacific\" }} }} }}");
+        var jIncludeOnlyLatestPosted = JsonNode.Parse($"{{ \"range\": {{ \"postedOn\": {{ \"gte\": \"{previousInstancePublishedOn.Value:yyyy-MM-ddTHH:mm:ssZ}\", \"time_zone\": \"US/Pacific\" }} }} }}");
```
Fosol, line 61: This was causing the bug with missing content

services/net/reporting/ReportingManager.cs
```diff
@@ -350,7 +350,7 @@ private async Task<(API.Areas.Services.Models.ReportInstance.ReportInstanceModel
 
         // If we don't clear the prior report we must append new content to it.
         var currentSectionContent = new Dictionary<string, ReportSectionModel>();
-        if (reportInstanceModel != null && !report.Settings.Content.ClearOnStartNewReport)
+        if (reportInstanceModel != null && !reportInstanceModel.SentOn.HasValue && !report.Settings.Content.ClearOnStartNewReport)
```
Fosol, line 353: I'm not 100% certain but I think this should only currently include the prior content if it wasn't sent.

app/editor/src/features/content/form/ContentActions.tsx
```diff
@@ -40,206 +40,201 @@ export interface IContentActionCheckbox {
...
+              break;
+          }
         }
-      };
+      },
+      [formActions, setFieldValue, field],
+    );
 
-      if (e.ctrlKey) {
-        e.stopImmediatePropagation();
```
Fosol, line 142: Removed these two lines so that it would still behave correctly.

PR #2043, opened 2024-07-18
TNO-2801: Adding content to folder/reports does not uncheck them

jdtoombs:
Pretty simple bug, the `onClear` was not being passed down to the problematic pages. Home page worked as expected because it had it.


app/subscriber/src/features/filter-media/FilterMedia.tsx
```diff
@@ -101,6 +101,7 @@ export const FilterMedia: React.FC<IFilterMediaProps> = ({ loaded }) => {
       <ContentListActionBar
```
Fosol, line 101: Maybe the ContentListActionBar should require the property to be provided?

jdtoombs: Ah never mind, we use that component when viewing singular pieces of content too, which won't have an `onClear`

Fosol (approved):

PR #2049, opened 2024-07-18
Fix reports

Fosol:
Fixing issue that occurs when attempting to update a report with a new chart.


libs/net/models/Areas/Admin/Report/ReportModel.cs
```diff
@@ -180,13 +180,7 @@ public Entities.Report ToEntity(JsonSerializerOptions options, bool includeDepen
             };
             section.ChartTemplatesManyToMany.AddRange(modelSection.ChartTemplates.OrderBy(ct => ct.SortOrder).Select(ct => new Entities.ReportSectionChartTemplate(modelSection.Id, ct.Id, ct.SortOrder)
             {
-                ChartTemplate = new Entities.ChartTemplate(ct.Id, ct.Name, ct.Template)
```
Fosol, line 183: Having the chart here resulted in an attempt to create a new one each time.

PR #2056, opened 2024-07-22
TNO-2869,TNO-2819: prevent override transcript, and match contributor and byline.

kkwangsir:
summary:

1. consider the` isApproved` from `props.values` disable request transcript if the transcript is approved.
2. Automatically select the Contributor, Contributor, and Byline input that match.

image


Fosol (changes requested):
This is not going to solve the issue.  The Editor can override the transcript to their hearts content.  What we don't want is for a subscriber requesting a transcript and then the Transcription Service overwriting the Editor's work.

kkwangsir:
Added `IsApproved` check in backend. 
In the subscriber application, it has been handled to disable the `Request Transcript` button for any contents have `isApproved' . 

```
        <Show
          visible={
            isAV &&
            !content?.source?.disableTranscribe &&
            !content.isApproved &&
            !isTranscriptRequestor
          }
        >

```
Also add the toast pop up message when subscriber request a content with a approved transcription. 
image

Fosol (approved):

PR #2060, opened 2024-07-23
Change Authentication

Fosol:
We're moving away from the Shared Keycloak Realm to a Custom Keycloak Realm with Azure Entra IDP.  This required refactoring all the CSS authentication/authorization.

We can now support guest email addresses through the Keycloak users.  Ideally we could use guest accounts within Azure, but they appear to be limited to Microsoft accounts.

Also need to update all the service configurations in Openshift to get this to work.

> This will require manually refreshing your local Keycloak setup and configuration.

## Summary

- Migration to Keycloak Custom Realm
- Add Azure Entra IDP
- Refactor local Keycloak
- Publish tno-core:0.1.112


api/net/Areas/Admin/Controllers/UserController.cs
```diff
@@ -128,7 +132,14 @@ public IActionResult Add(UserModel model)
     [SwaggerOperation(Tags = new[] { "User" })]
     public async Task<IActionResult> UpdateAsync(UserModel model)
     {
-        await _cssHelper.UpdateUserRolesAsync(model.Key, model.Roles.ToArray());
+        if (!Guid.TryParse(model.Key, out Guid key)) throw new InvalidOperationException($"User key must be a guid: {model.Key}");
+
+        var kUser = await _keycloakService.GetUserAsync(key);
```
Fosol, line 137: Only update if the user exists in keycloak.

app/subscriber/src/components/content-list/ContentListContext.tsx
```diff
@@ -13,7 +13,7 @@ export interface IContentListProviderProps {
   children: ReactNode;
 }
 export const ContentListProvider: React.FC<IContentListProviderProps> = ({ children }) => {
-  const [, { authenticated }] = useApp();
+  const [, { initialized }] = useApp();
```
Fosol, line 16: This won't make a request if the user is not authorized yet.

app/subscriber/src/features/access-request/AccessRequest.tsx
```diff
@@ -36,27 +32,36 @@ export const AccessRequest: React.FC = () => {
...
       <Show
         visible={
-          userInfo?.status !== UserStatusName.Requested &&
-          userInfo?.status !== UserStatusName.Denied
+          userInfo &&
+          [UserStatusName.Preapproved, UserStatusName.Approved].includes(userInfo.status)
         }
       >
-        <Col>
+        <Col alignContent="center">
```
Fosol, line 39: Fix display issues for new users.

docker-compose.yml
```diff
@@ -142,28 +142,6 @@ services:
     networks:
       - tno
 
-  ####################### Azure Storage #######################
-  azure-storage:
```
Fosol, line 146: Not needed

docker-compose.yml
```diff
@@ -348,32 +326,6 @@ services:
     networks:
       - tno
 
-  ####################### API #######################
-  css:
```
Fosol, line 352: Not needed

libs/net/keycloak/Models/ResourceAccessModel.cs
```diff
@@ -17,13 +17,13 @@ public class ResourceAccessModel
     /// <summary>
     /// get/set - Realm management claims.
     /// </summary>
-    [JsonPropertyName("tno-app")]
+    [JsonPropertyName("mmi-app")]
```
Fosol, line 20: I'd love to come up with a way where these were not hardcoded.  It would have made this PR easier.

Fosol:
I still need to reconfigure all the services.

PR #2072, opened 2024-07-25
TNO-2857 Fixed logo issue and others

laidaoyu:
- fixed the logo issue
- updated login text message
- fixed sorting capitalized string issue (for TNO-2866)


Fosol (commented):
Can you fix the MobileLogin.tsx component?  If it's needed, then please extract any of the shared logic with the BrowserLogin.tsx component so that we don't get regression issues when we make changes to our login options.

Fosol (approved):

Fosol:
I've extracted the IDP buttons out to a separate component.  I don't think I did a great job, but it's at least a way to reduce duplication.  I've noticed the components are not very responsive...  Or at least they aren't done very well.

PR #2079, opened 2024-07-25
TNO-2796: Update My Minister View

jdtoombs:
The My Minister table has now been updated to `ContentList` and the `ContentRow` will display the minister mention tags when they are present (same behavior as before but on the new component).

@nehalaggarwal-bcgov - Just in case you wanted to see how this is done! 

image


app/subscriber/src/features/my-minister/MyMinister.tsx
```diff
@@ -229,19 +209,10 @@ export const MyMinister: React.FC = () => {
...
-          selectedRowIds={selectedIds}
-          isMulti
-          groupBy={(item) => item.original.ministerName ?? ''}
-          onRowClick={(e: any) => {
-            navigate(`/view/my-minister/${e.original.id}`);
-          }}
-          data={filteredContent}
-          pageButtons={5}
-          showPaging={false}
+        <ContentList
```
Fosol, line 212: cheer!

Fosol (approved):

PR #2085, opened 2024-07-25
CSS From Session w/ Bobbi

jdtoombs:
Changes to:

- settings css (minor)
- my products css

image

### DIFF
--- a/app/subscriber/src/features/my-products/MyProducts.tsx
+++ b/app/subscriber/src/features/my-products/MyProducts.tsx
@@ -4,7 +4,7 @@ import React from 'react';
 import { FaEnvelope, FaUserPlus } from 'react-icons/fa6';
 import { toast } from 'react-toastify';
 import { useApp, useProducts } from 'store/hooks';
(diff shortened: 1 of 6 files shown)


app/subscriber/src/features/my-products/MyProducts.tsx
```diff
@@ -98,10 +98,10 @@ export const MyProducts: React.FC = () => {
                 })}
             </div>
           </Show>
-          <Row className="page-section-title">
-            <FaUserPlus className="page-section-icon" /> Available products
-          </Row>
-          <p>
+          <h2 className="product-section-title">
+            <FaUserPlus /> <span>Available products</span>
```
jdtoombs, line 102: I've been guilty of this too, but we should wrap text in a div or span and not just leave it "hanging". Poses risk for whitespace styling errors

jdtoombs: ex)
Before the code was `<FaUserPlus /> Available Products` and another area was `<Icon />Test`. Could not figure out why the first had a `margin-right`, but in the end it was whitespace in the code.

Fosol (approved):

PR #2089, opened 2024-07-26
TNO-2907 Add status to disable send

fbarreta:
I've found out when it was a resend, the this.Api.UpdateReportInstanceAsync method was not called, not generating a new SignalR message, and keeping the status of the instance locked on "Submitted".
As far as I could see, the is no problem on running the update on resend, and having the "Last Sent", and Status up to date.
Testing it locally, it didn't generated multiple instances, and the data integrity seem to be OK.

### DIFF
--- a/app/subscriber/src/features/my-reports/edit/view/ReportViewForm.tsx
+++ b/app/subscriber/src/features/my-reports/edit/view/ReportViewForm.tsx
@@ -153,7 +153,10 @@ export const ReportViewForm: React.FC = () => {
           <Row alignItems="flex-start" className="preview-send-details-row">
             <Button
               disabled={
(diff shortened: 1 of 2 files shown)


Fosol:
I don't think we want to block the ability to resend.   There is technically wrong with sending a report again, although subscribers would receive multiple copies.

fbarreta:
> I don't think we want to block the ability to resend. There is technically wrong with sending a report again, although subscribers would receive multiple copies.

The main issue was on the status not being updated on resend, and then getting locked. The change on the service will do it.

Fosol (changes requested):

fbarreta:
Changes on service addressed this error.
This won't be necessary anymore.

PR #2090, opened 2024-07-26
Fix reporting

Fosol:
Failed reports will now keep trying over and over by sending a message to Kafka.  I will need to put a limiter on this at some point, but want to test before doing this.

A failed report which is partially successful will pick up where it left off, by publishing a message to Kafka for the report instance instead.

I also discovered that the prior implementation to restart services after failures was not implemented correctly, as such only a few services actually do this.  The new code added for the reports will need to be applied to other services later.

### DIFF
--- a/libs/net/kafka/KafkaListener`.cs
+++ b/libs/net/kafka/KafkaListener`.cs
@@ -283,7 +283,7 @@ public class KafkaListener<TKey, TValue> : IKafkaListener<TKey, TValue>, IDispos
     /// <param name="partitions"></param>
     public void Resume(IEnumerable<TopicPartition> partitions)
     {
-        if (this.IsPaused)
+        if (this.IsPaused || !this.IsConsuming)
         {
             _logger.LogDebug("Resuming consumption: {topics}", String.Join(", ", partitions.Select(p => p.Topic).Distinct()));
             this.IsPaused = false;
--- a/libs/net/services/Config/ServiceOptions.cs
+++ b/libs/net/services/Config/ServiceOptions.cs
@@ -25,9 +25,19 @@ public class ServiceOptions
     public int RetryDelayMS { get; set; } = 5000;
 
     /// <summary>
-    /// get/set - Default millisecond delay between process cycle.  This stops run-away threads (default: 30 second).
+    /// get/set - Default millisecond delay between process cycle.  This stops run-away threads (default: 30 seconds).
     /// </summary>
     public int DefaultDelayMS { get; set; } = 30000;
+
+    /// <summary>
+    /// get/set - Whether to restart a service after a critical sequence of failures.
+    /// </summary>
+    public bool AutoRestartAfterFailure { get; set; } = true;
+
+    /// <summary>
+    /// get/set - Number of millisecond delay before restarting service after a critical sequence of failures (default: 30 seconds).
+    /// </summary>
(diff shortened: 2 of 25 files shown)


libs/net/kafka/KafkaListener`.cs
```diff
@@ -283,7 +283,7 @@ public void Resume()
     /// <param name="partitions"></param>
     public void Resume(IEnumerable<TopicPartition> partitions)
     {
-        if (this.IsPaused)
+        if (this.IsPaused || !this.IsConsuming)
```
Fosol, line 286: Without this Kafka never appears to get going again after a failure in the service.

libs/net/services/Config/ServiceOptions.cs
```diff
@@ -25,10 +25,20 @@ public class ServiceOptions
...
     /// <summary>
-    /// get/set - Default millisecond delay between process cycle.  This stops run-away threads (default: 30 second).
+    /// get/set - Default millisecond delay between process cycle.  This stops run-away threads (default: 30 seconds).
     /// </summary>
     public int DefaultDelayMS { get; set; } = 30000;
 
+    /// <summary>
+    /// get/set - Whether to restart a service after a critical sequence of failures.
+    /// </summary>
+    public bool AutoRestartAfterFailure { get; set; } = true;
```
Fosol, line 35: These config options should have been implemented a long time ago when we added this functionality to a few of the services.

libs/net/services/ServiceStatus.cs
```diff
@@ -27,4 +27,12 @@ public enum ServiceStatus
...
     /// </summary>
     Sleeping = 4,
+    /// <summary>
+    /// The service has a failure and must stop.
+    /// </summary>
+    RequestFailed = 5,
+    /// <summary>
+    /// The service has failed.
+    /// </summary>
+    Failed = 6,
```
Fosol, line 37: Now we'll know if it failed or was put to sleep.

services/net/reporting/ReportingManager.cs
```diff
@@ -97,20 +97,32 @@ public override async Task RunAsync()
...
                 // The service is stopping or has stopped, consume should stop too.
                 this.Listener.Stop();
+                _cancelToken?.Cancel();
+            }
+            else if (this.State.Status == ServiceStatus.Failed)
+            {
+                this.Logger.LogDebug("The service has failed: '{Status}'", this.State.Status);
+                if (this.Options.AutoRestartAfterFailure)
+                {
+                    await Task.Delay(this.Options.RetryAfterFailedDelayMS);
```
Fosol, line 115: Now it will restart after a sequence of failures.

services/net/reporting/ReportingManager.cs
```diff
@@ -232,27 +244,40 @@ private async Task HandleMessageAsync(ConsumeResult<string, ReportRequestModel>
...
+            await this.SendEmailAsync("Failed to handle message", ex);
             ListenerErrorHandler(this, new ErrorEventArgs(ex));
+
+            await this.Api.SendMessageAsync(result.Message.Value);
         }
         finally
         {
+            // Inform Kafka this message is completed.
+            // We do this regardless of whether it failed or succeeded.
+            this.Listener.Commit(result);
```
Fosol, line 278: Because we post a failures back to Kafka we can commit all messages.

PR #2109, opened 2024-07-30
Update Reports to support failures

Fosol:
The reporting service now has the ability to send emails out to individuals.  If a failure occurs it will pick up where it left off.

## Summary

- Add DB migration 1.2.0
- Updated Editor app
- Updated API
- Updated Reporting Service

### DIFF
--- a/api/net/Areas/Admin/Controllers/ReportInstanceController.cs
+++ b/api/net/Areas/Admin/Controllers/ReportInstanceController.cs
@@ -146,7 +146,7 @@ public class ReportInstanceController : ControllerBase
         var username = User.GetUsername() ?? throw new NotAuthorizedException("Username is missing");
         var user = _userService.FindByUsername(username) ?? throw new NotAuthorizedException($"User [{username}] does not exist");
 
-        instance.Status = ReportStatus.Submitted;
+        instance.Status = instance.Status == Entities.ReportStatus.Pending ? Entities.ReportStatus.Submitted : instance.Status;
         instance = _reportInstanceService.UpdateAndSave(instance);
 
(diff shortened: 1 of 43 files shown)


api/net/Areas/Admin/Controllers/ReportInstanceController.cs
```diff
@@ -146,7 +146,7 @@ public async Task<IActionResult> Publish(int id)
         var username = User.GetUsername() ?? throw new NotAuthorizedException("Username is missing");
         var user = _userService.FindByUsername(username) ?? throw new NotAuthorizedException($"User [{username}] does not exist");
 
-        instance.Status = ReportStatus.Submitted;
+        instance.Status = instance.Status == Entities.ReportStatus.Pending ? Entities.ReportStatus.Submitted : instance.Status;
```
Fosol, line 149: Don't change the status here unless it was pending.  Provides better outcome when resending.

app/editor/src/features/admin/users/UserFormDistribution.tsx
```diff
@@ -77,6 +77,10 @@ export const UserFormDistribution: React.FC = () => {
     <div className="form-container">
       <Section className="frm-in">
         <label>Account Information</label>
+        <p className="info">
```
Fosol, line 80: Temporary until I fix it up tomorrow.

libs/net/dal/Migrations/1.2.0/Up/PostUp/00-Users.sql
```diff
@@ -0,0 +1,34 @@
+DO
+$BODY$
+
+DECLARE
+	tempRow record;
+	tempItem json;
+
+BEGIN
+
+	FOR tempRow IN
```
Fosol, line 10: Move the JSON reference to the users to an actual table with foreign keys.

PR #2119, opened 2024-08-01
TNO-2846 Help page videos

fbarreta:
I don't know exactly how the mapping would work, so I wonder if the mountPath is correct on the config below:
 - name: subscriber-public-storage
   mountPath: /public/videos

Rather than that, this is the Help Page Layout:
image


Fosol (changes requested):
Do not add any of the binaries to the repo.

openshift/kustomize/app/subscriber/base/deploy.yaml
```diff
@@ -61,6 +61,8 @@ spec:
             - name: subscriber-keycloak
               mountPath: /usr/share/nginx/html/keycloak.json
               subPath: keycloak.json
+            - name: subscriber-public-storage
```
Fosol, line 64: You need to have a volume identified in the `volumes` section above.

openshift/kustomize/app/subscriber/base/deploy.yaml
```diff
@@ -61,6 +61,8 @@ spec:
...
+              mountPath: /public/videos
```
Fosol, line 65: I doubt this path is correct.  The path must be to the location the app is installed.  You'll need to review the Dockerfile.

Fosol (changes requested):

Fosol (approved):

PR #2125, opened 2024-08-02
TNO-2777 report sticky buttons

nehalaggarwal-bcgov:
Added code to make the buttons stick to the bottom.


Fosol (changes requested):
Something appears wrong with your branch history

laidaoyu:
The changes Nehal made look ok to me. I am not sure if the git history would look right after merging to dev. I will let Jeremy to decide.

PR #2128, opened 2024-08-02
TNO-2777 code to stick buttons to bottom

nehalaggarwal-bcgov:
TNO-2777 added code to stick buttons to bottom on my reports


app/subscriber/src/features/my-reports/edit/styled/ReportEditForm.tsx
```diff
@@ -4,6 +4,7 @@ export const ReportEditForm = styled.div`
   display: flex;
   flex-direction: column;
   overflow-x: clip;
+  min-height: 89vh;
```
Fosol, line 7: this won't sticky to the bottom.  It'll only work for one specific screen size.

Fosol (approved):

PR #2137, opened 2024-08-06
Fix report dashboard

Fosol:
An update to the report dashboard to help make it work in production.  We still need to change the endpoint to only return the necessary data for each report, then a separate request for the subscribers (this will be in a separate PR).  Hopefully this PR will get it working in production.

## Summary

- Publish tno-core:0.1.128
- Update API
- Update Editor app

## Report Dashboard

image


app/editor/src/features/admin/reports/dashboard/ReportDashboard.tsx
```diff
@@ -1,32 +1,71 @@
...
+    fetchReports(filter).catch(() => {});
+    // Only when filter changes.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [filter]);
 
   return (
     <styled.ReportDashboard>
       <h1>Report Dashboard</h1>
+      <Row justifyContent="center" gap="1rem;">
```
Fosol, line 50: A couple basic filters

PR #2140, opened 2024-08-07
Revamp log-in page

jdtoombs:
In this PR:

- simplified code
- got rid of excess 
- new lightweight modal component, maybe eventually we can replace with this everywhere (uses html's dialog)
- now mobile responsive 

Updated:
image

Pasted image 20240807004704
Pasted image 20240807004824
Pasted image 20240807004846


app/subscriber/src/components/lightweight-modal/styled/LightweightModal.tsx
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components';
+
+export const LightweightModal = styled.dialog`
```
jdtoombs, line 3: Using [<dialog> which is native to HTML 5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog), super nice simple way for modals.

Fosol (approved):
Thanks

PR #2146, opened 2024-08-08
Fix highlight keywords

Fosol:
Updated the code to extract keywords so that they can be highlighted in a report.  This resolves compound phrases.

### DIFF
--- a/libs/net/elastic/Extensions/StringExtensions.cs
+++ b/libs/net/elastic/Extensions/StringExtensions.cs
@@ -9,14 +9,15 @@ namespace TNO.Elastic;
 public static class StringExtensions
 {
     #region Variables
-    static readonly Regex RangeRegex = new Regex(@":([\[\{])");
-    static readonly Regex BoostOrProximityRegex = new Regex(@"[\^\~][0-9]*$");
-    static readonly Regex WildCardRegex = new Regex(@"[\*\?]");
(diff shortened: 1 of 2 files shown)


libs/net/elastic/Extensions/StringExtensions.cs
```diff
@@ -75,16 +78,16 @@ public static IEnumerable<string> ExtractKeywords(this string search, string que
                 else if (endOfPhrase > -1)
                 {
                     // Reset and add these words to a phrase.
-                    phrase.Append(cleanToken);
-                    keywords.Add(phrase.ToString());
+                    phrase.Append($" {cleanToken}");
```
Fosol, line 81: Fix phrases

libs/net/elastic/Extensions/StringExtensions.cs
```diff
@@ -61,11 +63,12 @@ public static IEnumerable<string> ExtractKeywords(this string search, string que
                     cleanToken = cleanToken.Replace("*", "").Replace("?", "");
 
                 // TODO: This doesn't handle escaped characters.
-                // TODO: Handle grouping.
-                cleanToken = cleanToken.Replace("(", "").Replace(")", "").Replace("\"", "");
+                cleanToken = cleanToken.Replace("(", "").Replace(")", "");
```
Fosol, line 66: Fix grouping

PR #2147, opened 2024-08-08
TNO-2896: Update commentary block to use a simplified list view

jdtoombs:
In this PR:

- add a new prop called `simpleView` to the `ContentList` component. This will hide attributes, and the checkbox. 
- small styling changes 
- adds media buttons to the commentary block 

image

### DIFF
--- a/app/subscriber/src/features/commentary/Commentary.tsx
+++ b/app/subscriber/src/features/commentary/Commentary.tsx
@@ -1,13 +1,13 @@
 import { Action } from 'components/action';
+import { ContentList } from 'components/content-list';
 import { PageSection } from 'components/section';
-import { Sentiment } from 'components/sentiment';
 import { useActionFilters } from 'features/search-page/hooks';
 import { filterFormat } from 'features/search-page/utils';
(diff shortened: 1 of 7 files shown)


app/subscriber/src/features/commentary/Commentary.tsx
```diff
@@ -112,19 +111,7 @@ export const Commentary: React.FC = () => {
...
-                <Sentiment value={x.tonePools?.length ? x.tonePools[0].value : 0} />
-                <DetermineContentIcon contentType={x.contentType} />
-                <div className="headline" onClick={() => navigateAndScroll(`/view/${x.id}`)}>
-                  {x.headline}
-                </div>
-              </Row>
-            );
-          })}
-        </div>
+        <ContentList selected={[]} onContentSelected={noop} content={commentary} simpleView />
```
jdtoombs, line 114: I tried making selected, and onContentSelected optional props; however that resulted in many errors. For now passing an empty array and noop. Will create a ticket to make this more flexible.

Fosol (approved):

PR #2155, opened 2024-08-09
TNO-924 fixed multiple calls to pageview snowplow

nehalaggarwal-bcgov:
changes to files to call trackpageview once per page

### DIFF
--- a/app/subscriber/src/features/router/AppRouter.tsx
+++ b/app/subscriber/src/features/router/AppRouter.tsx
@@ -31,6 +31,7 @@ export const AppRouter: React.FC = () => {
   const navigate = useNavigate();
 
   React.useEffect(() => {
+    window.snowplow('trackPageView');
     // There is a race condition, when keycloak is ready state related to user claims will not be.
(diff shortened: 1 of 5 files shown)


app/subscriber/src/features/router/AppRouter.tsx
```diff
@@ -31,6 +31,7 @@ export const AppRouter: React.FC = () => {
   const navigate = useNavigate();
 
   React.useEffect(() => {
+    window.snowplow('trackPageView');
```
Fosol, line 34: I am doubtful this will work as it will fire every time the navigate function is changed.  But we can give it a try.

nehalaggarwal-bcgov: Isn't the navigate function going to change only when the page view changes? When else will it be fired?

Fosol: It 'might' change, but you would have to test it.  It's a function, so unless it's maintaining state it should rarely change.

PR #2162, opened 2024-08-12
TNO-2859 Fix find by date method

fbarreta:
Looks like this is a long time bug, and was causing a lot of confusion.

The find by date method was never returning data, since the date comparison was wrong never bringing the actual AV Overview, every time it was only loaded with the template data.
Since the template was loaded, the search methods were never called, bringing invalid data from previous states, showing wrong data to the user when switching between dates.

Once this is merged, I'll make some final tests in dev, but I believe, data wise it's the definitive fix.


Fosol (approved):
Thanks!

PR #2165, opened 2024-08-13
TNO-2945: When  a schedule is enabled, at least one day must be selected for running

jdtoombs:
Just adding some validation for reports.

Previously, you could enable a schedule, set a start time and save. Now you must select at least one day before saving.

image


app/subscriber/src/features/my-reports/edit/settings/template/ReportSchedule.tsx
```diff
@@ -44,8 +43,9 @@ export const ReportSchedule: React.FC<IReportScheduleProps> = ({ index, label })
           </p>
         </Col>
       </Col>
-      <Col className="frm-in">
+      <Col className="frm-in run-on">
         <label>Run On</label>
+        <div role="alert">{getIn(errors, `events.${index}.runOnWeekDays`)}</div>
```
jdtoombs, line 48: Placing the error here, rather than having it appear 7 times under each checkbox.

Fosol (approved):

PR #2187, opened 2024-08-16
Fix Notification Service

Fosol:
Users should not receive duplicate notifications.

### DIFF
--- a/app/subscriber/src/features/content/view-content/ViewContent.tsx
+++ b/app/subscriber/src/features/content/view-content/ViewContent.tsx
@@ -385,7 +385,8 @@ export const ViewContent: React.FC<IViewContentProps> = ({ setActiveContent, act
               isAV &&
               !content?.source?.disableTranscribe &&
               !content.isApproved &&
-              !isTranscriptRequestor
+              !isTranscriptRequestor &&
+              !!content?.fileReferences.length
             }
           >
             <Button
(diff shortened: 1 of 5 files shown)


app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -385,7 +385,8 @@ export const ViewContent: React.FC<IViewContentProps> = ({ setActiveContent, act
               isAV &&
               !content?.source?.disableTranscribe &&
               !content.isApproved &&
-              !isTranscriptRequestor
+              !isTranscriptRequestor &&
+              !!content?.fileReferences.length
```
Fosol, line 389: Don't allow asking for a transcript if there is no file.

services/net/notification/NotificationManager.cs
```diff
@@ -411,7 +407,8 @@ private async Task ResetAlertAsync(API.Areas.Services.Models.Content.ContentMode
         subscribers.AddRange(content.UserNotifications.Where(cun => cun.User != null && cun.IsSubscribed && cun.User.AccountType != UserAccountType.Distribution)
             .Select(cun => new API.Areas.Services.Models.Notification.UserModel(cun.User!)));
 
-        return subscribers;
+        // Remove duplicate subscribers.  This can occur if a user is both subscribed to the content and the notification.
+        return subscribers.GroupBy(s => s.Id).Select(s => s.First());
```
Fosol, line 411: Only return unique users.

PR #2193, opened 2024-08-19
TNO-2915 refresh button consistency

nehalaggarwal-bcgov:
TNO-2915 changed Action to Refresh Button component for reusability


app/editor/src/features/admin/event-of-the-day/styled/EventOfTheDayList.tsx
```diff
@@ -95,4 +95,15 @@ export const EventOfTheDayList = styled.div`
     transition: all 1s;
     transform: rotateY(180deg);
   }
+
+  .icon-refresh {
```
Fosol, line 99: You shouldn't need this if the new component `RefreshButton` has been styled.

nehalaggarwal-bcgov: RefreshButton is only added to Subscriber. And there will be a separate story for Editor site.

Fosol (approved):

PR #2194, opened 2024-08-20
Evening Overview Bug Fixes

Fosol:
This will need to be tested more in PROD due to the lack of content to review.  

I have cleaned up the useEffects and useCallback statements to only refresh when a value changes.  There is still a lot of noise when the summary text is edited due to the section item being updated and the suggested value popup.

Clips should now be displayed correctly.

## Editor Evening Overview

image


Fosol (commented):
I removed some functionality that I couldn't figure out.  There was some weird logic going on with the summary, the time, a regex, and selecting a clip option.

PR #2196, opened 2024-08-20
TNO-2949 Prevent sending report with no subscribers 

kkwangsir:
disable sending button in editor and subscriber 
prevent any sending with no subscribers in reporting service, and set status to failed.

image

### DIFF
--- a/app/subscriber/src/features/my-reports/ReportPreview.tsx
+++ b/app/subscriber/src/features/my-reports/ReportPreview.tsx
@@ -43,6 +43,7 @@ export const ReportPreview = ({ report, onFetch, onClose }: IReportPreviewProps)
   const [isSubmitting, setIsSubmitting] = React.useState(false);
 
   const instance = report?.instances.length ? report.instances[0] : undefined;
+  const hasSubscribers = report?.subscribers?.some((s) => s.isSubscribed === true) ?? false;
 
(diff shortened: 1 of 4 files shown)


app/subscriber/src/features/my-reports/ReportPreview.tsx
```diff
@@ -194,10 +195,18 @@ export const ReportPreview = ({ report, onFetch, onClose }: IReportPreviewProps)
...
+              title={
+                !hasSubscribers
+                  ? 'There are no subscribers for this report. Add subscribers to enable sending.'
+                  : ''
+              }
+            >
               Send
               <FaPaperPlane />
-            </Button>
+            </Button>{' '}
```
Fosol, line 209: Do you need this space?  Wouldn't it be better to use css to accomplish this?

Fosol (commented):
Can you confirm that after a failure due to not having subscribers, that you can then correct it by adding subscribers and then manually sending.

kkwangsir:
> Can you confirm that after a failure due to not having subscribers, that you can then correct it by adding subscribers and then manually sending.
yes. I have tested this case. will tweak the page first.

Fosol (approved):

kkwangsir:
image

PR #2205, opened 2024-08-22
TNO-2858 loading indicator editor

nehalaggarwal-bcgov:
Added loading indicator to Papers tab, added to ContentListView as well but need to test in dev.


app/editor/src/features/admin/ministers/MinisterList.tsx
```diff
@@ -16,14 +16,18 @@ const MinisterList: React.FC = () => {
   const [{ ministers }, api] = useMinisters();
 
   const [items, setItems] = React.useState<IMinisterModel[]>([]);
+  const [isLoading, setIsLoading] = React.useState(false);
```
Fosol, line 19: Take a look at the `ReportCard.tsx` component.  It used the `useApp()` hook state property `requests` to determine if it's loading.  Then you don't need to manage another state variable everywhere.

Fosol (commented):
You may be interested in using the `useApp()` hook.  It has a state property called `requests` which we use to keep track of active requests made to the API.  You can check if there is an active request on the page to determine if the loading indicator should be visible.

An example of it used when any request is made is `ContentReferenceList.tsx`.

An example of it used when a specific request is made is `ReportCard.tsx

app/editor/src/features/admin/ministers/MinisterList.tsx
```diff
@@ -16,14 +17,18 @@ const MinisterList: React.FC = () => {
   const [{ ministers }, api] = useMinisters();
 
   const [items, setItems] = React.useState<IMinisterModel[]>([]);
+  const [{ requests }] = useApp();
 
   React.useEffect(() => {
+    // setIsLoading(true);
```
Fosol, line 23: No need to keep these commented out lines.

libs/npm/core/src/components/grid/Grid.tsx
```diff
@@ -93,6 +97,9 @@ export const Grid = <T,>({
           })}
         </div>
         <div className="grid-table">
+          <Show visible={isLoading}>
```
Fosol, line 100: Have you tested this to determine if changing the `isLoading` property in the parent component makes this component re-render.  React state doesn't detect property changes.

app/editor/src/features/content/papers/Papers.tsx
```diff
@@ -68,7 +68,7 @@ const Papers: React.FC<IPapersProps> = (props) => {
   const [, setCurrentItems] = useLocalStorage('currentContent', {} as IContentSearchResult[]);
   const [currentItemId, setCurrentItemId] = useLocalStorage('currentContentItemId', -1);
 
-  const [, setIsLoading] = React.useState(false);
+  const [isLoading, setIsLoading] = React.useState(false);
```
Fosol, line 71: Instead of managing a separate state value you can `useApp` and check if there is an active request for content.

app/editor/src/features/content/list-view/ContentListView.tsx
```diff
@@ -206,78 +206,75 @@ const ContentListView: React.FC = () => {
...
-              ],
-            };
-
-            const response = await findWorkOrders(woFilter);
-            workOrders = response.data;
-            filter.contentIds = workOrders.items
-              .filter((wo) => !!wo.contentId)
-              .map((wo) => wo.contentId!);
-          }
+        setIsLoading(true);
```
Fosol, line 209: Instead of managing a separate state value you can `useApp` and check if there is an active request for content.

libs/npm/core/src/components/box/Box.tsx
```diff
@@ -16,7 +16,7 @@ export const Box: React.FC<IBoxProps> = ({ isLoading, className, children, ...re
     <styled.Box className={`box${className ? ` ${className}` : ''}`} {...rest}>
       {isLoading && (
```
Fosol, line 17: Validate that changing the parent component will result in the child state being updated.

libs/npm/core/src/components/grid/Grid.tsx
nehalaggarwal-bcgov: I tested in the local environment and it is showing the loading indicator on the editor content page.

Fosol: On the initial load it may.  What if you change the value from false to true again?  Depending on various scenarios properties do not cause a render when they change.

app/editor/src/features/content/papers/Papers.tsx
nehalaggarwal-bcgov: I can do it in another PR.

app/editor/src/features/content/list-view/ContentListView.tsx
nehalaggarwal-bcgov: I can do it in another PR.

libs/npm/core/src/components/box/Box.tsx
nehalaggarwal-bcgov: I just changed the loading component here. The rest of the logic should work as before.

libs/npm/core/src/components/grid/Grid.tsx
nehalaggarwal-bcgov: Seems to be working when changing the filters as well.

Fosol (changes requested):
I have published `tno-core:0.1.39` and updated the components to use the `useApp` logic.  This makes it less lines of code to maintain for loading.

Testing indicates the loading overlay does not align or expand correctly.

image

Fosol:
You will need to reset your branch, as I needed to rebase and force push the latest changes.

app/editor/tno-core-0.1.134.tgz
Fosol, line 1: remove

app/editor/tno-core-0.1.138.tgz
Fosol, line 1: remove

app/subscriber/tno-core-0.1.134.tgz
Fosol, line 1: remove

app/subscriber/tno-core-0.1.138.tgz
Fosol, line 1: remove

Fosol (changes requested):

Fosol (approved):

PR #2206, opened 2024-08-23
Fix no content in resent report

Fosol:
Fixed issue where a resent report would result in a no content message.  This appears to only have impacted the `Send a Test` email, but it is used by Scott more than realized.

### DIFF
--- a/services/net/reporting/ReportingManager.cs
+++ b/services/net/reporting/ReportingManager.cs
@@ -535,6 +535,17 @@ public class ReportingManager : ServiceManager<ReportingOptions>
     }
 
     /// <summary>
+    /// Determine if this request is to resend the report.
+    /// </summary>
+    /// <param name="request"></param>
+    /// <param name="instance"></param>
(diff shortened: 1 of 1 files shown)


services/net/reporting/ReportingManager.cs
```diff
@@ -534,6 +534,17 @@ private async Task GenerateAndSendReportAsync(ReportRequestModel request, API.Ar
         await SendReportAsync(request, report, instanceModel, sectionContent);
     }
 
+    /// <summary>
+    /// Determine if this request is to resend the report.
+    /// </summary>
+    /// <param name="request"></param>
+    /// <param name="instance"></param>
+    /// <returns></returns>
+    private bool IsResend(ReportRequestModel request, API.Areas.Services.Models.ReportInstance.ReportInstanceModel? instance)
```
Fosol, line 543: This logic allows for a report to be sent out without needing to fetch all the content and regenerating the output.

PR #2207, opened 2024-08-23
Fix bug 500+ return bug

jdtoombs:
When 500+ results were returned it would truncate - meaning that the currently selected date may not have the results present on the page even if there were some (as the previous 5 days could be taking those spots).

### DIFF
--- a/app/subscriber/src/features/todays-commentary/TodaysCommentary.tsx
+++ b/app/subscriber/src/features/todays-commentary/TodaysCommentary.tsx
@@ -47,7 +47,6 @@ export const TodaysCommentary: React.FC = () => {
     if (commentaryActionId) {
       let actionFilters = getActionFilters();
       const commentaryAction = actionFilters.find((a) => a.id === commentaryActionId);
-      setIsLoading(true);
       commentaryAction &&
         fetchResults({
           ...filter,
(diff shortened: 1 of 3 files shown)


app/subscriber/src/features/todays-commentary/TodaysCommentary.tsx
```diff
@@ -54,7 +54,7 @@ export const TodaysCommentary: React.FC = () => {
           actions: [commentaryAction],
           startDate: filter.startDate ?? moment().startOf('day').toISOString(),
           searchUnpublished: false,
-          size: 500,
```
Fosol, line 57: I think we still need a limiter on size.

app/subscriber/src/features/todays-commentary/TodaysCommentary.tsx
```diff
@@ -47,15 +47,15 @@ export const TodaysCommentary: React.FC = () => {
     if (commentaryActionId) {
       let actionFilters = getActionFilters();
       const commentaryAction = actionFilters.find((a) => a.id === commentaryActionId);
-      setIsLoading(true);
```
jdtoombs, line 50: handled in hook

Fosol (approved):

PR #2211, opened 2024-08-25
Improve report charts

Fosol:
We now have a report chart wizard to aid in the design of charts and the complicated configuration settings.  You can also generate random test data to see how the chart looks before you report even has content in it.

## Summary

- Add chart.js
- Add chart data generator
- Publish tno-core:0.1.135
- Updated API
- Updated Subscriber App

## Subscriber Report Chart Template Wizard

image


app/editor/src/features/admin/reports/utils/createReportSection.ts
```diff
@@ -27,8 +27,6 @@ export const createReportSection = (
           ? 'row'
           : 'column',
       groupBy: 'otherSource',
-      datasetColors: [],
```
Fosol, line 30: Not including them allows for using the default configured colours.

app/editor/src/features/content/list-view/components/tool-bar/filter/InputOption.tsx
```diff
@@ -52,7 +52,7 @@ export const InputOption: React.FC<any> = ({
       <input type="checkbox" id={children} defaultChecked={isSelected} />
       <styled.SelectLabel
         className="label"
-        onClick={() => document.getElementById(children)?.click()}
```
Fosol, line 55: I was trying to figure out what was throwing errors, but I don't think this was it.  I think the drag+drop library is broken.

app/subscriber/package.json
```diff
@@ -18,6 +18,8 @@
     "@react-keycloak/web": "3.4.0",
     "@reduxjs/toolkit": "2.2.5",
     "axios": "1.7.2",
+    "chart.js": "4.4.4",
```
Fosol, line 21: Add charts to the subscriber app so that we can dynamically view them when creating a report.

app/subscriber/src/features/my-reports/edit/ReportEditContext.tsx
```diff
@@ -126,6 +135,20 @@ export const ReportEditContextProvider: React.FC<IReportEditContextProviderProps
   const instance = values.instances.length ? values.instances[0] : undefined;
   const [activeInstance, setActiveInstance] = React.useState<IReportInstanceModel>();
   const [previewLastUpdatedOn, setPreviewLastUpdatedOn] = React.useState<string | undefined>();
+  const [showReportData, setShowReportData] = React.useState(
+    values.instances.length > 0 && values.instances[0].content.length > 0,
+  );
+  const [testData, setTestData] = React.useState(
+    generateReportInstanceContents({
```
Fosol, line 142: I want to make this configurable by the user later.  Then they can create custom data that more reflects what their report might look like.

app/subscriber/src/features/my-reports/edit/settings/template/utils/generateChartOptions.ts
```diff
@@ -0,0 +1,116 @@
+import { IChartSectionSettingsModel } from 'tno-core';
+
+import { IChartData } from './IChartData';
+
+export const generateChartOptions = (
```
Fosol, line 5: There is a similar function in tno-core called `mergeChartSettings` that I would like to combine so that we don't have similar logic in two places.

libs/net/models/Settings/ChartSectionSettingsModel.cs
```diff
@@ -53,17 +63,17 @@ public class ChartSectionSettingsModel
     /// <summary>
     /// get/set - An array of colour to use for this chart datasets.
     /// </summary>
-    public string[] DatasetColors { get; set; } = Array.Empty<string>();
+    public string[]? DatasetColors { get; set; }
```
Fosol, line 66: Nullable colours allows us to use default colours.

libs/net/models/Settings/ChartSectionSettingsModel.cs
```diff
@@ -15,6 +15,16 @@ public class ChartSectionSettingsModel
...
 
+    /// <summary>
+    /// get/set - Whether width and height should force aspect ratio.
+    /// </summary>
+    public bool? MaintainAspectRatio { get; set; }
+
+    /// <summary>
+    /// get/set - The chart aspect ratio.
+    /// </summary>
+    public int? AspectRatio { get; set; }
```
Fosol, line 26: I need to add this to the chart api at some point.

openshift/scripts/deploy.sh
```diff
@@ -30,13 +30,23 @@ podsEditor=$(getPods editor dc $env)
...
 
 # podsCapture=$(getPods capture-service dc $env)
-podsContentMigration=$(getPods contentmigration-service dc $env)
-podsContentMigrationRecent=$(getPods contentmigration-recent-service dc $env)
-podsContentMigrationHistoric=$(getPods contentmigration-historic-service dc $env)
 podsFileMonitor=$(getPods filemonitor-service dc $env)
 podsSyndication=$(getPods syndication-service dc $env)
 podsImage=$(getPods image-service dc $env)
 
+if [ $env = "prod" ]; then
```
Fosol, line 37: Fixed deployment script

PR #2212, opened 2024-08-26
Fix multiple charts

Fosol:
When adding charts to reports, this will allow multiple charts to be displayed at the same time.

## Subscriber Report Chart Editor

image

### DIFF
--- a/app/subscriber/src/features/my-reports/edit/ReportEditContext.tsx
+++ b/app/subscriber/src/features/my-reports/edit/ReportEditContext.tsx
@@ -60,8 +60,6 @@ export interface IReportEditContext {
   setPreviewLastUpdatedOn: React.Dispatch<React.SetStateAction<string | undefined>>;
   testData: IReportInstanceContentModel[];
   regenerateTestData: () => void;
-  showReportData: boolean;
-  toggleReportData: () => void;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


app/subscriber/src/features/my-reports/edit/ReportEditContext.tsx
```diff
@@ -60,8 +60,6 @@ export interface IReportEditContext {
   setPreviewLastUpdatedOn: React.Dispatch<React.SetStateAction<string | undefined>>;
   testData: IReportInstanceContentModel[];
   regenerateTestData: () => void;
-  showReportData: boolean;
```
Fosol, line 63: Using the test data is now controlled by the component so that changing the view doesn't change for all charts.

app/subscriber/src/features/my-reports/edit/settings/template/charts/ChartViewer.tsx
```diff
@@ -34,7 +34,7 @@ ChartJS.register(
   ChartDataLabels,
 );
 
-let myChart: ChartJS;
+let myChart: Record<string, ChartJS> = {};
```
Fosol, line 37: Need to maintain a link to multiple charts based on the uid

PR #2215, opened 2024-08-26
[Ready] TNO-2977: Expand WYSIWYG

jdtoombs:
Updates since draft:

- reverted state fix raised by @Fosol - will create a new PR for that
- now have to keep two separate states and sync them together as Quill has some focus and other issues when two instances of an editor are open
- updated editor and removed the now unnecessary `Modals`

Summary:
- added `<dialog>` to the WYSIWYG component itself so we do not need to create a modal everywhere we use WYSIWYG (can expand by default everywhere WYSIWYG is called)
- will be more responsive with new element 

image


libs/npm/core/src/components/form/wysiwyg/Wysiwyg.tsx
```diff
@@ -194,19 +190,25 @@ export const Wysiwyg: React.FC<IWysiwygProps> = ({
...
+            id={props.id}
+            name={props.name}
+            disabled={props.disabled}
             className="raw-editor"
             onChange={(e) => setState({ ...state, text: e.target.value })}
             value={state.text}
           />
         </>
       )}
+      {expand && !!toolBarNode && (
```
jdtoombs, line 206: All functionality now in the same component

libs/npm/core/src/components/form/wysiwyg/styled/Wysiwyg.tsx
```diff
@@ -3,6 +3,27 @@ import styled from 'styled-components';
...
+    }
+  }
+  dialog {
+    border: none;
+    width: 100%;
+    height: 85%;
+    .frm-in {
+      height: 85%;
+      .expand-button {
+        display: none;
```
jdtoombs, line 20: do not display the expand option when we have already expanded (otherwise user could open infinite instances and then have to close x instances)

app/editor/src/features/content/form/ContentStoryForm.tsx
```diff
@@ -77,53 +63,10 @@ export const ContentStoryForm: React.FC<IContentStoryFormProps> = ({
               : 'Summary'
           }
           name="body"
-          expandModal={setShowExpandModal}
           tags={tags}
           onChange={(text) => getTags('body', text)}
         />
       </Show>
-      <Modal
```
jdtoombs, line 85: No longer need these

jdtoombs: (For expanded wysiwyg)

Fosol (approved):

PR #2233, opened 2024-08-29
Fix report charts

Fosol:
Fixing various issues with resize and highlight keywords in charts.

image

### DIFF
--- a/api/node/src/routes.ts
+++ b/api/node/src/routes.ts
@@ -1,5 +1,14 @@
 import { Router } from 'express';
-import { CategoryScale, Colors } from 'chart.js';
+import {
+  CategoryScale,
+  Colors,
+  Legend,
+  LinearScale,
+  LineElement,
+  PointElement,
+  Title,
+  Tooltip,
+} from 'chart.js';
 import Chart from 'chart.js/auto';
 import { getString, convertChartJsConfigToBase64String, generateBase64, sendImage } from './utils';
 import { ChartTypes } from './charts';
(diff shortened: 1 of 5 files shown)


api/node/src/routes.ts
```diff
@@ -1,5 +1,14 @@
 import { Router } from 'express';
-import { CategoryScale, Colors } from 'chart.js';
+import {
```
Fosol, line 2: Make it the same registered features as the Subscriber app UI.

libs/net/template/ReportEngine.cs
```diff
@@ -217,15 +217,6 @@ public async Task<string> GenerateBase64ImageAsync(
         {
             // If the chart should be resized, calculate the new size and update the options.
             UpdateChartSize(dataModel, model.ChartTemplate.SectionSettings, 30);
-            var sectionJsonText = model.ChartTemplate.SectionSettings.Options.ToJson();
```
Fosol, line 220: Doesn't do anything

libs/net/template/ReportExtensions.cs
```diff
@@ -245,7 +245,7 @@ public static string GetContentGroupByPropertyValue(this ContentModel content, s
             "contentType" => content.ContentType.ToString(),
             "topicType" => content.Topics.FirstOrDefault()?.TopicType.ToString() ?? "",
             "topicName" => content.Topics.FirstOrDefault()?.Name ?? "",
-            "byline" => content.Byline,
+            "byline" => content.Byline.Replace("<mark>", "").Replace("</mark>", ""),
```
Fosol, line 248: Resolve highlight keywords bug

PR #2236, opened 2024-08-29
TNO-3016 Fix exclude content filters

fbarreta:
There are two fixes:
On line 830 there's a check on the section setting OverrideExcludeHistorical, when there's nothing to do with the exclude report content. So the feature is controlled only by the content setting ExcludeReports.
This fixes the reports using folders as data source.

On the line 854, there was a check that I believe was include by mistake on sectionSettings.RemoveDuplicates, that should be sectionSettings.OverrideExcludeHistorical (since the array is called excludeAboveAndHistorical), and also another problem that the on false was returning an empty array, and was overwriting the content that should be excluded by the excludeOnlyTheseContentIds.
This fixes the reports using filters as data source.

image
Report used to test content to be excluded.

image
Option to exclude report empty
image
Content

image
Option to exclude report filled
image
Content

### DIFF
--- a/libs/net/dal/Services/ReportService.cs
+++ b/libs/net/dal/Services/ReportService.cs
@@ -827,7 +827,7 @@ public class ReportService : BaseService<Report, int>, IReportService
                 if (sectionSettings.RemoveDuplicates)
                     query = query.Where(fc => !excludeAboveSectionContentIds.Contains(fc.ContentId));
 
-                if (excludeContentIds.Any() && !sectionSettings.OverrideExcludeHistorical)
+                if (excludeContentIds.Any())
                     query = query.Where(fc => !excludeContentIds.Contains(fc.ContentId));
 
                 var content = query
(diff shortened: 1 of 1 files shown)


libs/net/dal/Services/ReportService.cs
```diff
@@ -851,9 +851,9 @@ public IEnumerable<ReportInstance> GetLatestInstances(int id, int? ownerId = nul
 
                 // Modify the query to exclude content.
                 var excludeOnlyTheseContentIds = excludeContentIds.Any() && !sectionSettings.OverrideExcludeHistorical ? excludeContentIds : Array.Empty<long>();
-                var excludeAboveAndHistorical = sectionSettings.RemoveDuplicates
+                var excludeAboveAndHistorical = sectionSettings.OverrideExcludeHistorical
```
Fosol, line 854: If a user selects the "Include all content from linked folder even if in prior report" in a section does this work?  This is where the `overrideExcludeHistorical` setting comes from.  It's a per-section configuration option.  It ensures that that specific section will still include content that was also in a linked report.  The reason for this is because Scott wants the folder content to show up no matter what.

image

fbarreta: It would, because the variable excludeOnlyTheseContentIds on line 853 is assigned like that:
var excludeOnlyTheseContentIds = excludeContentIds.Any() && !sectionSettings.OverrideExcludeHistorical ? excludeContentIds : Array.Empty<long>();
It would only be populated if the Override is checked.
In this case excludeOnlyTheseContentIds is empty and would not exclude those contents.

Fosol (approved):

PR #2239, opened 2024-08-30
Fix report charts

Fosol:
Fixed issue with charts that should only display the content in their own section.

Fixed issue with regenerate report and charts.

Fixed modal disable confirm button.

### DIFF
--- a/app/subscriber/src/components/modal/Modal.tsx
+++ b/app/subscriber/src/components/modal/Modal.tsx
@@ -90,7 +90,7 @@ export const Modal: React.FC<IModalProps> = ({
                       <Button
                         variant={type === 'delete' ? 'error' : 'primary'}
                         onClick={onConfirm}
-                        disabled={isSubmitting && !enableConfirm}
+                        disabled={isSubmitting || !enableConfirm}
                       >
                         {confirmText ?? 'Continue'}
                       </Button>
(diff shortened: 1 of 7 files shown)


app/subscriber/src/components/modal/Modal.tsx
```diff
@@ -90,7 +90,7 @@ export const Modal: React.FC<IModalProps> = ({
                       <Button
                         variant={type === 'delete' ? 'error' : 'primary'}
                         onClick={onConfirm}
-                        disabled={isSubmitting && !enableConfirm}
+                        disabled={isSubmitting || !enableConfirm}
```
Fosol, line 93: If either then disable.  The way this was implemented didn't make sense.  I hope where ever else it's used it doesn't break.

app/subscriber/src/features/my-reports/edit/content/ReportEditContentForm.tsx
```diff
@@ -156,6 +156,9 @@ export const ReportEditContentForm = React.forwardRef<
                 }}
               />
             </Show>
+            <Show visible={isSubmitting}>
+              <Loading />
```
Fosol, line 160: The regenerate process takes a bit.

app/subscriber/src/features/my-reports/edit/settings/template/ReportSectionMediaAnalyticsChart.tsx
```diff
@@ -55,8 +55,14 @@ export const ReportSectionMediaAnalyticsChart = React.forwardRef<
   }, [values.instances]);
 
   React.useEffect(() => {
-    setData(convertToChart(chart, showReportData ? reportContent : testData, values.sections));
-  }, [chart, reportContent, showReportData, testData, values.sections]);
+    const sectionContent =
+      section.filterId || section.folderId || section.linkedReportId
```
Fosol, line 59: Only include content in the section when appropriate.

PR #2247, opened 2024-09-03
TNO-2970 sync to s3

kkwangsir:
setup local docker-compose service

main script is tools/objects-backup/entrypoint.sh

curl the minIO tool using it to sync files to s3 bucket incrementally.

we could add the script into the openshift cornjob later

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -347,7 +347,20 @@ services:
       - host.docker.internal:host-gateway
     networks:
       - tno
-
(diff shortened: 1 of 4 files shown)


Fosol (approved):
Am I understanding this correctly - This syncs files in the container with S3 bucket?

kkwangsir:
> Am I understanding this correctly - This syncs files in the container with S3 bucket?

Yes. I use my local papers data to sync.

PR #2251, opened 2024-09-04
TNO-1136 Set tone by user

areyeslo:
**Feature Update: User-Defined Tone Assessment for Articles**

This update introduces the ability for subscribers to assess and set a personal tone for an article, independent of the original tone. The custom tone selected by the user will be saved and associated with their account, ensuring that their personal assessment is retained without altering the article's original tone data.

**User Tone:**
image

**Editor Tone:**
image

In DB, a record is created on `tone_pool` table which is used to tone any content. When the user set a tone for a content, it is saved in `content_tone` table.

**Story**: https://citz-gdx.atlassian.net/browse/TNO-1136


api/net/Areas/Subscriber/Controllers/TonePoolController.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// <param name="model"></param>
+    /// <returns></returns>
+    ///
+    [HttpPost]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(TonePoolModel), (int)HttpStatusCode.Created)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "TonePool" })]
+    [ETagCacheTableFilter("tone_pools")]
+    [ResponseCache(Duration = 5 * 60)]
```
Fosol, line 99: Doesn't make sense on a POST endpoint

api/net/Areas/Subscriber/Controllers/TonePoolController.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// </summary>
+    /// <param name="model"></param>
+    /// <returns></returns>
+    ///
+    [HttpPost]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(TonePoolModel), (int)HttpStatusCode.Created)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "TonePool" })]
+    [ETagCacheTableFilter("tone_pools")]
```
Fosol, line 98: Doesn't make sense on a POST endpoint

api/net/Areas/Subscriber/Controllers/TonePoolController.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// <summary>
+    /// Return an array of TonePool.
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet, HttpHead]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(IEnumerable<TonePoolModel>), (int)HttpStatusCode.OK)]
+    [ProducesResponseType((int)HttpStatusCode.NotModified)]
+    [SwaggerOperation(Tags = new[] { "TonePool" })]
+    [ETagCacheTableFilter("tone_pools")]
```
Fosol, line 59: Doesn't make sense on an endpoint that returns one item.

api/net/Areas/Subscriber/Controllers/TonePoolController.cs
```diff
@@ -0,0 +1,118 @@
...
+    [ResponseCache(Duration = 5 * 60)]
```
Fosol, line 60: Have you tested this?

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
```diff
@@ -27,17 +30,67 @@ export interface IContentEditFormProps {
...
+      };
+
+      const createTonePool = async (userId: number) => {
+        try {
+          await addMyTonePool({
+            ...defaultTonePool,
+            name: `${userId}`,
+            ownerId: userId,
+          });
+          const newTonePool = await getMyTonePool(userId);
```
Fosol, line 74: Extra request isn't needed, just use the returned tonepool from the `addMyTonePool`.

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
```diff
@@ -215,7 +268,13 @@ export const ContentEditForm = React.forwardRef<HTMLDivElement | null, IContentE
             </Col>
             <Col className="sentiment">
               <Sentiment
-                value={form.content.tonePools.length ? form.content.tonePools[0].value : undefined}
+                value={
+                  userFormTonePool
```
Fosol, line 272: This will result in always using their own tonepool.  The user needs to choose if they want to use their own tonepool.  In most cases users will not change the sentiment.

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
```diff
@@ -27,17 +30,67 @@ export interface IContentEditFormProps {
...
+    React.useEffect(() => {
+      const getTonePool = async () => {
+        try {
+          if (!init.myTonePool && userId !== 0) {
+            const response = await getMyTonePool(userId);
+            if (response?.id) {
+              storeMyTonePool(response);
+            } else {
+              // If no valid tone pool exists, proceed to create one
+              await createTonePool(userId);
```
Fosol, line 58: This will result in creating a tonepool for each user that edits content in a report.  Many users will never want to use their own tonepool.  They only need to be created if they choose to change the sentiment.

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
```diff
@@ -42,22 +47,101 @@ export const ContentForm: React.FC<IContentFormProps> = ({
...
+      } catch (ex) {
+        console.error('Error updating tone pools:', ex);
+      }
+    },
+    [content, onContentChange, userId],
+  );
+
+  React.useEffect(() => {
+    const updatedTonePools =
+      content?.tonePools && content?.tonePools.length > 0
```
Fosol, line 126: Finding the correct tonepool is repeated code.  Make it a utility function to improve maintainability.

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
```diff
@@ -71,12 +155,17 @@ export const ContentForm: React.FC<IContentFormProps> = ({
           rows={1}
           disabled={disabled}
           onChange={(e) => {
+            if (!content || typeof content.id !== 'number') {
```
Fosol, line 158: What's the value of this error message that is repeated throughout this page?  If content doesn't exist why is the user editing the form?

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
```diff
@@ -152,6 +253,26 @@ export const ContentForm: React.FC<IContentFormProps> = ({
             onContentChange?.(values);
           }}
         />
+        <Formik
```
Fosol, line 256: Change the `FormikSentiment` to work in a way that is intuitive.  Using a sub-form just to get around this issue is confusing and difficult to maintain.

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
```diff
@@ -152,6 +253,26 @@ export const ContentForm: React.FC<IContentFormProps> = ({
...
+          {() => (
+            <Form>
+              <FormikSentiment
+                name="tonePools"
+                options={[...tonePools, myTonePool]}
+                coloredIcon={true}
+                defaultTonePoolId={myTonePool.id}
+                defaultTonePoolName={myTonePool.name}
+                onSentimentChange={handleSentimentChange}
+                required
```
Fosol, line 271: Sentiment isn't required for subscribers

app/subscriber/src/features/my-reports/edit/validation/SentimentFormSchema.ts
```diff
@@ -0,0 +1,9 @@
+import { array, number, object } from 'yup';
+
+export const sentimentFormSchema = object().shape({
+  tonePools: array().of(
+    object().shape({
+      value: number().required('Tone value is required'),
```
Fosol, line 6: It isn't required

app/subscriber/src/store/hooks/lookup/useLookup.ts
```diff
@@ -156,13 +158,27 @@ export const useLookup = (): [ILookupState, ILookupController] => {
           'lookup',
         );
       },
+      getTonePools: async () => {
```
Fosol, line 161: This isn't really something we need.  We're not currently allowing users to choose someone else's tonepool.  Also the way you have currently implemented this could result in hundreds of records returned in this request when at most the user would only need their own tonepool and the default tonepool.

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
```diff
@@ -27,17 +30,67 @@ export interface IContentEditFormProps {
...
     const instance = values.instances.length ? values.instances[0] : undefined;
     const userId = userInfo?.id ?? 0;
 
+    React.useEffect(() => {
+      const getTonePool = async () => {
+        try {
+          if (!init.myTonePool && userId !== 0) {
+            const response = await getMyTonePool(userId);
+            if (response?.id) {
+              storeMyTonePool(response);
```
Fosol, line 55: You are already doing this in your hook.  No need to do it again.

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
```diff
@@ -27,17 +30,67 @@ export interface IContentEditFormProps {
...
+
+      const createTonePool = async (userId: number) => {
+        try {
+          await addMyTonePool({
+            ...defaultTonePool,
+            name: `${userId}`,
+            ownerId: userId,
+          });
+          const newTonePool = await getMyTonePool(userId);
+          storeMyTonePool(newTonePool);
```
Fosol, line 75: You are already doing this in your hook.  No need to do it again.

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
```diff
@@ -27,17 +30,67 @@ export interface IContentEditFormProps {
...
+    >(undefined);
 
     const instance = values.instances.length ? values.instances[0] : undefined;
     const userId = userInfo?.id ?? 0;
 
+    React.useEffect(() => {
+      const getTonePool = async () => {
+        try {
+          if (!init.myTonePool && userId !== 0) {
+            const response = await getMyTonePool(userId);
```
Fosol, line 53: You are using redux to store their tonepool, yet you are going to make a request every time hit this page to get their tonepool from the API.

app/subscriber/src/store/slices/profile/interfaces/IProfileState.ts
```diff
@@ -23,10 +24,12 @@ export interface IProfileState {
...
     myReports: boolean;
     myColleagues: boolean;
+    myTonePool: boolean;
   };
   reportsFilter: string;
   reportOutput?: IReportResultForm;
   reportContent: { [reportId: number]: number[] };
   contributors: IContributorModel[];
   messages: ISystemMessageModel[];
+  myTonePool: ITonePoolModel;
```
Fosol, line 34: This should be undefined initially.

app/subscriber/src/store/slices/profile/profileSlice.ts
```diff
@@ -26,6 +27,16 @@ export const initialProfileState: IProfileState = {
     myFolders: false,
     myReports: false,
     myColleagues: false,
+    myTonePool: false,
+  },
+  myTonePool: {
```
Fosol, line 32: There is no value in initializing like this.

libs/net/dal/Extensions/ContentExtensions.cs
```diff
@@ -171,14 +173,23 @@ public static TNOContext UpdateContext(this TNOContext context, Content original
         {
             var current = a.TonePoolId != 0 ? oTonePools.FirstOrDefault(o => o.TonePoolId == a.TonePoolId) : null;
             if (current == null)
-                original.TonePoolsManyToMany.Add(a);
+            {
+                a.TonePool ??= context.TonePools.FirstOrDefault(tp => tp.Id == a.TonePoolId);
+                tonePoolsToAdd.Add(a);
```
Fosol, line 178: Why create a separate list that you then iterate over again?  Just do the logic here.

libs/net/dal/Services/TonePoolService.cs
```diff
@@ -23,5 +24,19 @@ public IEnumerable<TonePool> FindAll()
             .AsNoTracking()
             .OrderBy(a => a.SortOrder).ThenBy(a => a.Name).ToArray();
     }
+
+    public TonePool? FindByUserId(int userId)
+    {
+        return this.Context.TonePools.FirstOrDefault(tp => tp.OwnerId == userId && tp.IsPublic == true);
```
Fosol, line 30: We're presently only going to create one tonepool for each user.  It doesn't matter if it's public.

libs/net/entities/Models/ContentVersion.cs
```diff
@@ -56,6 +56,10 @@ public class ContentVersion
     /// </summary>
     public bool IsPrivate { get; set; } = false;
 
+    /// <summary>
+    /// get/set - Tone for the content
+    /// </summary>
+    public int? Tone {get;set;}
```
Fosol, line 62: Are you even using this property?

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
```diff
@@ -152,6 +253,26 @@ export const ContentForm: React.FC<IContentFormProps> = ({
...
+        <Formik
+          initialValues={initialValues}
+          validationSchema={sentimentFormSchema}
+          onSubmit={() => {}}
+          enableReinitialize={true}
+        >
+          {() => (
+            <Form>
+              <FormikSentiment
+                name="tonePools"
```
Fosol, line 265: If you correctly named this component then the Formik `setFieldValue` would work correctly. and you wouldn't need also include the `onSentimentChange` logic.

libs/npm/core/src/components/formik/sentiment/FormikSentiment.tsx
```diff
@@ -64,40 +75,65 @@ export const FormikSentiment = <T extends object>({
...
+      // Replace the existing TonePool value
+      updatedTonePools = tonePools?.map((pool, index) =>
+        index === existingTonePoolIndex ? { ...pool, value: option } : pool,
+      );
+    } else {
+      // Add a new TonePool entry
+      updatedTonePools = [...(tonePools || []), { ...defaultTonePool, value: option }];
+    }
+
+    setFieldValue(name.toString(), updatedTonePools, true);
```
Fosol, line 122: Having both of these can result in odd state.  Formik will update the form context and then another separate set of state is being updated next.  It is easier to maintain if you make the `setFieldValue` work correctly.

app/subscriber/src/store/hooks/lookup/useLookup.ts
Fosol: When reviewing the `LookupController` I see it is also currently configured to return all tonepools.  This isn't ideal and should be changed to at most, only those which are public.

libs/npm/core/src/hooks/api/subscriber/useApiSubscriberTonePools.ts
```diff
@@ -0,0 +1,37 @@
...
+    lifecycleToasts?: ILifecycleToasts;
+    selector?: Function;
+    envelope?: typeof defaultEnvelope;
+    baseURL?: string;
+  } = {},
+) => {
+  const api = useApi(options);
+
+  return React.useRef({
+    getTonePools: (etag: string | undefined = undefined) => {
```
Fosol, line 17: This endpoint doesn't appear to exist.

libs/npm/core/tsconfig.json
```diff
@@ -7,11 +7,10 @@
...
     "declarationMap": true,
     "emitDeclarationOnly": false,
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "paths": {
+      "@assets/*": ["../../../../app/subscriber/public/assets/*"]
```
Fosol, line 12: I don't understand the need for this?  Shouldn't we keep tno-core assets locally?  If it's the shared library then the apps that use the shared library should be getting their components/assets from tno-core.  It is also easier to maintain.

Fosol (changes requested):

api/net/Areas/Subscriber/Controllers/TonePoolController.cs
areyeslo: I have not tested findbyId, just findByUserId. Going to test it.

api/net/Areas/Subscriber/Controllers/TonePoolController.cs
areyeslo: Thank you!

libs/net/entities/Models/ContentVersion.cs
areyeslo: No, this is from the previous implementation. Thank you!

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
areyeslo: I'd need to check this by chat. I'd love to implement this way :)

libs/npm/core/tsconfig.json
areyeslo: Thank you for bringing this, I wanted to use the proper icons instead of applying style:
`<FaSmile className="tone-icon" color="#59E9BE" />`
The icons are under `app/subscriber/public/assets/reports/`, so I created an alias to be used in the import. Then I could use `import @reports/face-negative@2x.png`

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
areyeslo: So the creation of the tonePool should be until they click on the sentiment icon at the bottom.

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
areyeslo: Not using FormikSentiment component.

libs/npm/core/src/components/formik/sentiment/FormikSentiment.tsx
areyeslo: Not using FormikSentiment.

app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
areyeslo: created a toggle to change between user tone and editor tone.

app/subscriber/src/features/my-reports/edit/content/stories/ContentForm.tsx
areyeslo: Implemented another component.

app/subscriber/src/store/hooks/lookup/useLookup.ts
areyeslo: Removed to avoid retriving hundred of records.

Fosol (approved):

PR #2252, opened 2024-09-04
Prevent multiple notification in indexing service.

kkwangsir:
In 5 seconds not send a same notification.
Test that it works according to the logic that was designed.

some logs:

```
2024-09-04 00:11:48 dbug: TNO.Core.Http.OpenIdConnectRequestClient[0]
2024-09-04 00:11:48       HTTP request made: POST:http://host.docker.internal:40010/api/kafka/producers/notification (User-Agent: TNO)
2024-09-04 00:11:48 info: TNO.Services.Indexing.IndexingManager[0]
2024-09-04 00:11:48       Notification for Content 1061 was sent, added to cache.
.....
2024-09-04 00:11:52 info: TNO.Services.Indexing.IndexingManager[0]
2024-09-04 00:11:52       Notification for Content 1061 was recently sent, skipping this notification request.
2024-09-04 00:11:52 dbug: TNO.Services.Indexing.IndexingManager[0]
```

### DIFF
--- a/services/net/indexing/IndexingManager.cs
+++ b/services/net/indexing/IndexingManager.cs
@@ -46,7 +46,7 @@ public class IndexingManager : ServiceManager<IndexingOptions>
     protected ElasticsearchClient Client { get; private set; }
 
     private readonly IMemoryCache _cache;
-    private const string CacheKeyPrefix = "IndexedContent_";
+    private const string CacheKeyPrefix = "NotificationContent_";
     private const int CacheExpirationSecond = 5;
     #endregion
 
--- a/services/net/indexing/IndexingManager.cs
+++ b/services/net/indexing/IndexingManager.cs
@@ -1,8 +1,10 @@
 using Confluent.Kafka;
 using Elastic.Clients.Elasticsearch;
 using Elastic.Transport;
+using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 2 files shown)


services/net/indexing/IndexingManager.cs
```diff
@@ -457,17 +467,31 @@ private async Task SendNotifications(IndexRequestModel request, ContentModel con
...
+                // TODO: Generate appropriate notification request.
+                var notification = new NotificationRequestModel(NotificationDestination.SignalR | NotificationDestination.NotificationService, content.Id)
+                {
+                    RequestorId = request.RequestorId
+                };
+                _ = await this.Api.SendMessageAsync(notification)
+                    ?? throw new HttpClientRequestException($"Failed to receive result from Kafka when sending message.  Topic: {this.Options.NotificationTopic}, Content ID: {content.Id}");
+
+                // add to cache
+                _cache.Set(cacheKey, true, TimeSpan.FromSeconds(CacheExpirationSecond));
```
Fosol, line 486: The problem is this is a memory sink bug.  It'll continue to grow slowly until the pod crashes.

Fosol: You will need to find a solution where these variables don't keep increasing, but rather are cleared.

kkwangsir: yes, you are right. could has a better way.

Fosol (approved):

PR #2254, opened 2024-09-04
override script to backup to s3 in dev

kkwangsir:
It is similar to local process, but happens in dev namespace 

I have `s3-backup-credentials-dev` in dev secrete.

One thing I would like to double check, 

` 
SOURCE_DIR="/mnt/data"
`

and I saw 
```
  - name: api-storage
     mountPath: /mnt/data
```
Is this one correct that we want to backup?


Fosol:
We don't technically want to backup anything.  What we need is another location to store older files.
After X days we need a process to move files from `api-storage` to `S3 storage bucket`.  When this occurs it needs to update the content so that the file references point to S3.

PR #2256, opened 2024-09-04
TNO-2033 Merge user versions on content

fbarreta:
- Checks if there's an updated version of the content
- If that's true, update the content version, but keeps the user "versions" with the user ID.

### DIFF
--- a/app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
+++ b/app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
@@ -28,7 +28,7 @@ export const ContentEditForm = React.forwardRef<HTMLDivElement | null, IContentE
   ({ disabled }, ref) => {
     const [{ userInfo }] = useApp();
     const [, { updateReport }] = useReports();
-    const [, { addContent, updateContent, getContent }] = useContent();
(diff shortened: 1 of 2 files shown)


app/subscriber/src/features/my-reports/edit/content/ContentEditForm.tsx
```diff
@@ -60,18 +60,47 @@ export const ContentEditForm = React.forwardRef<HTMLDivElement | null, IContentE
...
+          if (!content || content === undefined) return null;
+          const originalId = content.id;
           const err = validate(content);
           if (err.hasErrors) return null;
-          const originalId = content.id;
-          const contentResult = !content.id
-            ? await addContent(content)
-            : await updateContent(content);
+          // gets the edited content to check if the version was changed by another user
+          const updatedContent = await getContent(originalId);
```
Fosol, line 93: Making a request to get latest here isn't ideal as it's an extra request which isn't required most of the time.  Also it doesn't actually stop the issue from occurring as the content could be updated in the time it takes to send another request.  Instead attempt to save first, if the failure is for concurrency, then get the latest and merge.

Fosol (changes requested):

Fosol (approved):

PR #2275, opened 2024-09-10
TNO-2973 Send out Top Stories fixes

laidaoyu:
- only send out top stories that are not sent out previously
- remembers the last send out time in setting table
- added ES query to check the content posted on time / content action as Top Story with value "true"


Fosol (approved):
Looks good

PR #2277, opened 2024-09-11
Fix schedule concurrency errors

Fosol:
A common error is related to updated the schedule event.  It will often throw an optimistic concurrency error due to various race conditions.  This new code will attempt to resolve that type of issue by fetch the latest, reapplying changes and then saving.

## Summary

- Updated Scheduler Service
- Updated Event Handler Service (for clearing folders)
- Updated Reporting Service

### DIFF
--- a/libs/net/services/Helpers/ApiService.cs
+++ b/libs/net/services/Helpers/ApiService.cs
@@ -118,6 +118,41 @@ public class ApiService : IApiService
             return await RetryRequestAsync<T>(callbackDelegate);
         }
     }
+
+    /// <summary>
+    /// Keep trying a request if the failure is caused by an optimistic concurrency error.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="callbackDelegate"></param>
+    /// <returns></returns>
+    public async Task<T> HandleConcurrencyAsync<T>(Func<Task<T>> callbackDelegate)
+    {
+        // Keep trying to update the record and handle concurrency errors.
+        while (true)
+        {
+            try
+            {
+                return await callbackDelegate();
+            }
+            catch (HttpClientRequestException ex)
+            {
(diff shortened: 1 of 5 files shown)


libs/net/services/Helpers/ApiService.cs
```diff
@@ -118,6 +118,41 @@ protected async Task<T> RetryRequestAsync<T>(Func<Task<T>> callbackDelegate)
...
+    /// <summary>
+    /// Keep trying a request if the failure is caused by an optimistic concurrency error.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="callbackDelegate"></param>
+    /// <returns></returns>
+    public async Task<T> HandleConcurrencyAsync<T>(Func<Task<T>> callbackDelegate)
+    {
+        // Keep trying to update the record and handle concurrency errors.
+        while (true)
```
Fosol, line 131: Essentially keep trying until successful.  It will only error out if the exception is not a optimistic concurrency error.

libs/net/services/Helpers/ApiService.cs
```diff
@@ -118,6 +118,41 @@ protected async Task<T> RetryRequestAsync<T>(Func<Task<T>> callbackDelegate)
...
+        {
+            try
+            {
+                return await callbackDelegate();
+            }
+            catch (HttpClientRequestException ex)
+            {
+                // If it's a concurrency error, keep trying.  Otherwise throw the error.
+                this.Logger.LogError(ex, "Failed to complete request.  Determining if this is a concurrency error.");
+                var data = ex.Data["Body"] as string;
```
Fosol, line 141: I couldn't find a safe way to determine if the key exists  in the dictionary.  Regrettably .net uses a IDictionary object that doesn't have those features.

PR #2283, opened 2024-09-12
TNO-3062 minor updates to the logs

laidaoyu:
- should re-throw exception after logging


libs/net/dal/Services/ProductService.cs
```diff
@@ -444,6 +444,7 @@ public void AddAndSave(UserProduct subscription)
         } catch (Exception ex)
         {
             this.Logger.LogError(ex, $"ProductService - AddAndSave ProductId: {subscription.ProductId}, UserId: {subscription.UserId} throws exception.");
+            throw;
```
Fosol, line 447: Is there a reason you have added all these try+catch statements in this service library?  All of these would bubble their exception out to the `ErrorHandlingMiddleware` class and you could deal with logging there.  What data are you missing that needs a log statement here?

laidaoyu: I try to log more information about the exception, such as which product/xml statement caused the exceptions.

Fosol (approved):

PR #2287, opened 2024-09-13
TNO-3025 remove today's date from search result if offset greater than 1

nehalaggarwal-bcgov:
TNO-3025 change to not show date picker on more than 24 hr period


JacobWang-bc (approved):
looks good for me

PR #2289, opened 2024-09-17
Tweak API support S3 bucket stream

kkwangsir:
-  add new fields `IsSyncedToS3`  `last_synced_to_s3_on`  `S3Path ` to `FileReference`
-  add `force`  param to `GetFiles` method , and `upload-files-to-s3` endpoint
-  Update Editor/Subscriber/Content and Storage Controller to support priority use of S3 files. 
-  add `delete-old-local-files` api
image

To make this PR small, I make another PR to refactor transcription service.

### DIFF
--- a/api/net/Areas/Editor/Controllers/StorageController.cs
+++ b/api/net/Areas/Editor/Controllers/StorageController.cs
@@ -614,5 +614,20 @@ public class StorageController : ControllerBase
             FailedUploads = failedUploads,
         });
     }
+
+    /// <summary>
+    /// Delete old local files
+    /// </summary>
+    /// <param name="beforeDate">only delete files updated before the specified date</param>
+    /// <returns>deleted file count</returns>
+    [HttpDelete("delete-old-local-files")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(int), (int)HttpStatusCode.OK)]
+    [SwaggerOperation(Tags = new[] { "Storage" })]
+    public async Task<IActionResult> DeleteOldLocalFiles([FromQuery] DateTime beforeDate)
+    {
+        var deletedCount = await _fileReferenceService.DeleteOldLocalFilesAsync(beforeDate);
+        return Ok(deletedCount);
+    }
     #endregion
 }
--- a/api/net/Areas/Editor/Controllers/StorageController.cs
+++ b/api/net/Areas/Editor/Controllers/StorageController.cs
@@ -42,6 +42,7 @@ public class StorageController : ControllerBase
     private readonly ApiOptions _apiOptions;
     private readonly IFileReferenceService _fileReferenceService;
     private readonly ILogger<StorageController> _logger;
+    private readonly S3Options _s3Options;
     #endregion
 
     #region Constructors
@@ -53,13 +54,16 @@ public class StorageController : ControllerBase
     /// <param name="apiOptions"></param>
     /// <param name="fileReferenceService"></param>
     /// <param name="logger"></param>
-    public StorageController(IConnectionHelper connection, IOptions<StorageOptions> storageOptions, IOptions<ApiOptions> apiOptions, IFileReferenceService fileReferenceService, ILogger<StorageController> logger)
+    /// <param name="s3Options"></param>
+    public StorageController(IConnectionHelper connection, IOptions<StorageOptions> storageOptions, IOptions<ApiOptions> apiOptions,
+        IFileReferenceService fileReferenceService, ILogger<StorageController> logger, IOptions<S3Options> s3Options)
     {
         _connection = connection;
         _storageOptions = storageOptions.Value;
         _apiOptions = apiOptions.Value;
         _fileReferenceService = fileReferenceService;
         _logger = logger;
+        _s3Options = s3Options.Value;
     }
     #endregion
 
@@ -533,10 +537,11 @@ public class StorageController : ControllerBase
         var uploadedFiles = new List<string>();
         var failedUploads = new List<string>();
         // check if s3 credentials are set
-        if (!S3Options.IsS3Enabled)
+        if (!_s3Options.IsS3Enabled )
         {
             return BadRequest("S3 is not enabled or credentials are not set");
         }
+
         foreach (var fileReference in fileReferences)
         {
             try
--- a/api/net/Areas/Editor/Controllers/StorageController.cs
+++ b/api/net/Areas/Editor/Controllers/StorageController.cs
@@ -517,7 +517,7 @@ public class StorageController : ControllerBase
     /// <summary>
     /// upload files to s3
     /// </summary>
-    /// <param name="updatedBefore">optional, only upload files updated before the specified date</param>
+    /// <param name="createdBefore">optional, only upload files created before the specified date</param>
     /// <param name="limit">optional, only upload limit files</param>
     /// <param name="force">optional, force upload files</param>
     /// <returns>uploaded files and failed uploads</returns>
@@ -526,10 +526,10 @@ public class StorageController : ControllerBase
     [ProducesResponseType(typeof(Dictionary<string, List<string>>), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Storage" })]
-    public async Task<IActionResult> UploadFileToS3Async([FromQuery] DateTime? updatedBefore = null, [FromQuery] int? limit = null, [FromQuery] bool force = false)
+    public async Task<IActionResult> UploadFileToS3Async([FromQuery] DateTime? createdBefore = null, [FromQuery] int? limit = null, [FromQuery] bool force = false)
     {
         _logger.LogInformation("upload-files-to-s3");
-        var fileReferences = await _fileReferenceService.GetFiles(updatedBefore, limit ?? 100, force);
+        var fileReferences = await _fileReferenceService.GetFiles(createdBefore, limit ?? 100, force);
         var uploadedFiles = new List<string>();
         var failedUploads = new List<string>();
         // check if s3 credentials are set
--- a/api/net/Areas/Editor/Controllers/StorageController.cs
+++ b/api/net/Areas/Editor/Controllers/StorageController.cs
@@ -521,8 +521,8 @@ public class StorageController : ControllerBase
     /// <summary>
     /// upload files to s3
     /// </summary>
-    /// <param name="createdAfter">optional, only upload files created after the specified date</param>
-    /// <param name="createdBefore">optional, only upload files created before the specified date</param>
+    /// <param name="publishedAfter">optional, only upload files's content published after the specified date</param>
+    /// <param name="publishedBefore">optional, only upload files's content published before the specified date</param>
     /// <param name="limit">optional, only upload limit files</param>
     /// <param name="force">optional, force upload files</param>
     /// <returns>uploaded files and failed uploads</returns>
@@ -531,10 +531,10 @@ public class StorageController : ControllerBase
     [ProducesResponseType(typeof(Dictionary<string, List<string>>), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Storage" })]
-    public async Task<IActionResult> UploadFileToS3Async([FromQuery] DateTime? createdAfter = null, [FromQuery] DateTime? createdBefore = null, [FromQuery] int? limit = null, [FromQuery] bool force = false)
+    public async Task<IActionResult> UploadFileToS3Async([FromQuery] DateTime? publishedAfter = null, [FromQuery] DateTime? publishedBefore = null, [FromQuery] int? limit = null, [FromQuery] bool force = false)
     {
         _logger.LogInformation("upload-files-to-s3");
-        var fileReferences = await _fileReferenceService.GetFiles(createdAfter, createdBefore, limit ?? 100, force);
+        var fileReferences = await _fileReferenceService.GetFiles(publishedAfter, publishedBefore, limit ?? 100, force);
         var uploadedFiles = new List<string>();
         var failedUploads = new List<string>();
         // check if s3 credentials are set
--- a/api/net/Areas/Editor/Controllers/StorageController.cs
+++ b/api/net/Areas/Editor/Controllers/StorageController.cs
@@ -521,6 +521,7 @@ public class StorageController : ControllerBase
     /// <summary>
     /// upload files to s3
     /// </summary>
+    /// <param name="createdAfter">optional, only upload files created after the specified date</param>
     /// <param name="createdBefore">optional, only upload files created before the specified date</param>
     /// <param name="limit">optional, only upload limit files</param>
     /// <param name="force">optional, force upload files</param>
@@ -530,10 +531,10 @@ public class StorageController : ControllerBase
     [ProducesResponseType(typeof(Dictionary<string, List<string>>), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Storage" })]
-    public async Task<IActionResult> UploadFileToS3Async([FromQuery] DateTime? createdBefore = null, [FromQuery] int? limit = null, [FromQuery] bool force = false)
+    public async Task<IActionResult> UploadFileToS3Async([FromQuery] DateTime? createdAfter = null, [FromQuery] DateTime? createdBefore = null, [FromQuery] int? limit = null, [FromQuery] bool force = false)
     {
         _logger.LogInformation("upload-files-to-s3");
-        var fileReferences = await _fileReferenceService.GetFiles(createdBefore, limit ?? 100, force);
+        var fileReferences = await _fileReferenceService.GetFiles(createdAfter, createdBefore, limit ?? 100, force);
         var uploadedFiles = new List<string>();
         var failedUploads = new List<string>();
         // check if s3 credentials are set
--- a/api/net/Areas/Editor/Controllers/StorageController.cs
+++ b/api/net/Areas/Editor/Controllers/StorageController.cs
@@ -590,6 +590,16 @@ public class StorageController : ControllerBase
                         await _fileReferenceService.UpdateAsync(fileReference);
 
                         uploadedFiles.Add(s3Key);
+
+                        try
+                        {
+                            System.IO.File.Delete(filePath);
+                            _logger.LogInformation("deleted local file: {FilePath}", filePath);
+                        }
+                        catch (Exception ex)
+                        {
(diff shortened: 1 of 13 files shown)


api/net/Areas/Editor/Controllers/StorageController.cs
```diff
@@ -249,21 +249,45 @@ public IActionResult Stream([FromRoute] int? locationId, [FromQuery] string path
...
+        if (fileReference.IsSyncedToS3)
+        {
+            try
+            {
+                var stream = await _fileReferenceService.DownloadFromS3Async(path);
+                return File(stream, "application/octet-stream", fileReference.FileName);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "error on stream file from s3: {Path}", path);
```
laidaoyu, line 285: do you want to throw an exception here? otherwise, it will continue to the next statement 290.

JacobWang-bc: > do you want to throw an exception here? otherwise, it will continue to the next statement 290.

Hi Dao, this method try to find the path in s3 and then if s3 doesn't have, it will go to next step to check the local storage.

```        
if (!safePath.FileExists()) throw new NoContentException($"Stream does not exist: '{path}'");
```
both of them are not exist it will throw error, if it goes correct then it return file

 `return File(fileStream, info.MimeType!);`

api/net/Areas/Subscriber/Controllers/ContentController.cs
```diff
@@ -145,11 +146,14 @@ public IActionResult FindById(long id)
     [ProducesResponseType(typeof(FileStreamResult), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Content" })]
-    public IActionResult DownloadFile(long id)
+    public async Task<IActionResult> DownloadFile(long id)
```
Fosol, line 149: Standard convention for async functions are to rename to `DownloadFileAsync`

api/net/Areas/Subscriber/Controllers/ContentController.cs
```diff
@@ -162,15 +166,15 @@ public IActionResult DownloadFile(long id)
     [ProducesResponseType(typeof(FileStreamResult), (int)HttpStatusCode.PartialContent)]
     [ProducesResponseType((int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Content" })]
-    public IActionResult Stream([FromQuery] string path)
+    public async Task<IActionResult> Stream([FromQuery] string path)
```
Fosol, line 169: Standard convention for async functions are to rename to `StreamAsync`

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -225,7 +225,7 @@ public async Task<bool> UploadToS3Async(string s3Key, Stream fileStream)
         }
     }
 
-    public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore = null, int limit = 100)
+    public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore = null, int limit = 100, bool force = false)
```
Fosol, line 228: Standard convention for async functions are to rename to `GetFilesAsync`

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -254,5 +258,155 @@ public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore =
             return Enumerable.Empty<FileReference>();
         }
     }
-    #endregion
+
+    public async Task<Stream> DownloadFromS3Async(string s3Key)
+    {
+        var accessKey = Environment.GetEnvironmentVariable("S3_ACCESS_KEY") ?? throw new InvalidOperationException("S3_ACCESS_KEY environment variable is not set");
```
Fosol, line 264: Move all of these into dependency injected configuration options.  Look into the `/api/net/Config` folder for examples

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -254,5 +258,155 @@ public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore =
...
+    }
+
+    public async Task<FileReference?> GetByPathAsync(string path)
+    {
+        return await this.Context.FileReferences
+            .FirstOrDefaultAsync(fr => fr.Path == path);
+    }
+
+
+    public async Task<(Stream? Stream, string? FileName, string? ContentType)> GetFileStreamAsync(string path)
```
Fosol, line 325: Why are we searching for FileReferences by their path?  We should only be searching by their primary key.

api/net/Areas/Editor/Controllers/StorageController.cs
```diff
@@ -575,5 +614,20 @@ public async Task<IActionResult> UploadFilesToS3([FromQuery] DateTime? updatedBe
...
+    /// <summary>
+    /// Delete old local files
+    /// </summary>
+    /// <param name="beforeDate">only delete files updated before the specified date</param>
+    /// <returns>deleted file count</returns>
+    [HttpDelete("delete-old-local-files")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(int), (int)HttpStatusCode.OK)]
+    [SwaggerOperation(Tags = new[] { "Storage" })]
+    public async Task<IActionResult> DeleteOldLocalFiles([FromQuery] DateTime beforeDate)
```
Fosol, line 627: Standard convention for async functions are to rename to `DeleteOldLocalFilesAsync`

api/net/Areas/Editor/Controllers/StorageController.cs
```diff
@@ -249,21 +249,45 @@ public IActionResult Stream([FromRoute] int? locationId, [FromQuery] string path
...
-        if (!safePath.FileExists()) throw new NoContentException($"Stream does not exist: '{path}'");
+        _logger.LogInformation("Getting stream for path: {Path}", path);
+        var fileReference = await _fileReferenceService.GetByS3PathAsync(path);
+        _logger.LogInformation("File reference: {FileReference}", fileReference);
+        if (fileReference == null)
+        {
+            return NotFound($"Stream does not exist: '{path}'");
+        }
+
+        if (fileReference.IsSyncedToS3)
```
Fosol, line 276: Is there a reason this logic appears to be identical to the code in the FileReferenceService?  It appears like it could be centralized.

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -254,5 +258,155 @@ public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore =
...
+        if (fileReference.IsSyncedToS3)
+        {
+            try
+            {
+                var stream = await DownloadFromS3Async(path);
+                return (stream, fileReference.FileName, fileReference.ContentType);
+            }
+            catch (Exception ex)
+            {
+                Logger.LogError(ex, "Error on stream file from S3: {Path}", path);
```
Fosol, line 347: Why does it try to get the file at this location if the file doesn't exist? Logging an error indicates an issue.

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -254,5 +258,155 @@ public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore =
...
+    }
+
+    public async Task<FileReference> UpdateAsync(FileReference entity)
+    {
+        this.Context.Update(entity);
+        await this.Context.SaveChangesAsync();
+        return entity;
+    }
+
+    public async Task<int> DeleteOldLocalFilesAsync(DateTime beforeDate)
```
Fosol, line 369: I don't think this function makes business sense.  We would never want to delete old local files as a separate process.  We want to migrate files to S3 and at the same time and only after successfully copying the file we would delete the old reference.  This requires an update to the database and indexing to occur too.

libs/net/dal/Services/FileReferenceService.cs
JacobWang-bc: In fact, path is actually the key s3 path. However, that we don't have path defined as unique in our references, which could be a hidden problem. I've modified it to ignore the reference step and access s3 directly with path(key).

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -254,5 +258,155 @@ public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore =
...
+        catch (Exception ex)
+        {
+            Logger.LogError(ex, "Unexpected error when downloading file from S3: {S3Key}", s3Key);
+            throw;
+        }
+    }
+
+    public async Task<FileReference?> GetByS3PathAsync(string s3Path)
+    {
+        return await this.Context.FileReferences
```
Fosol, line 314: You shouldn't be searching the database for a file reference by path.  You should only be searching for file reference by its primary key.

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -254,5 +258,155 @@ public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore =
...
+            .FirstOrDefaultAsync(fr => fr.S3Path == s3Path);
+    }
+
+    public async Task<FileReference?> GetByPathAsync(string path)
```
Fosol, line 318: You shouldn't be searching the database for a file reference by path.  You should only be searching for file reference by its primary key.

JacobWang-bc: I am going to remove those. not search them. `path` come from endpoint, I will let them goes to s3 file query or local file finder directly.

libs/net/dal/Config/S3Options.cs
```diff
@@ -0,0 +1,95 @@
+using Amazon.Runtime;
+using Amazon.S3;
+
+namespace TNO.DAL.Config;
+/// <summary>
+/// S3 config
+/// </summary>
+public static class S3Options
```
Fosol, line 8: Generally using static classes this way isn't ideal.  It makes it difficult or impossible to use dependency injection and then later unit tests.

Is there a reason you cannot use the default configuration options like everything else?  Then dependency injection works out of the box correctly.

If you need a helper class, then it should also not be a static class.  If you only need a single instance of a shared helper class then configure dependency injection so that it only uses that one class.

libs/net/dal/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -9,6 +9,7 @@
 using TNO.DAL.Config;
 using TNO.DAL.Services;
 using TNO.Elastic;
+using Amazon.S3;
```
Fosol, line 12: Is this required?

libs/net/dal/Services/FileReferenceService.cs
Fosol: I still don't know why we would need this specific function.  We only ever want to delete a file when we migrate it from Openshift to S3.

api/net/Areas/Editor/Controllers/StorageController.cs
```diff
@@ -510,38 +519,24 @@ public async Task<IActionResult> JoinClipsAsync([FromRoute] int? locationId, [Fr
...
     /// <param name="limit">optional, only upload limit files</param>
+    /// <param name="force">optional, force upload files</param>
     /// <returns>uploaded files and failed uploads</returns>
     [HttpPost("upload-files-to-s3")]
     [Produces(MediaTypeNames.Application.Json)]
     [ProducesResponseType(typeof(Dictionary<string, List<string>>), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [SwaggerOperation(Tags = new[] { "Storage" })]
-    public async Task<IActionResult> UploadFilesToS3([FromQuery] DateTime? updatedBefore = null, [FromQuery] int? limit = null)
+    public async Task<IActionResult> UploadFileToS3Async([FromQuery] DateTime? updatedBefore = null, [FromQuery] int? limit = null, [FromQuery] bool force = false)
```
Fosol, line 529: When a file was updated last has little meaning or value to MMI.  We use those for audit fields, but generally nothing else.  Uploading to S3 probably should be based on the `Content.PublishedOn` date, as that has a relationship to the license.  Also there will be a lot of historical content that was recently updated, but it's old and should be moved to S3.

JacobWang-bc: Wrong word, it's a filter base on `PublishedOn` ,  I will change to a proper one like `PublishBefore`.

libs/net/dal/Config/S3Options.cs
JacobWang-bc: Sure. let me follow other options styles.

libs/net/dal/Config/S3Options.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.ComponentModel.DataAnnotations;
+
+namespace TNO.DAL.Config;
+
+/// <summary>
+/// S3 config
+/// </summary>
+public class S3Options
+{
+    public string BucketName => Environment.GetEnvironmentVariable("S3_BUCKET_NAME") ?? "";
```
Fosol, line 12: This isn't how configuration options should be set.  This will result in always using the environment variable value `S3_BUCKET_NAME`, where it should be following the dependency injection configuration you set below `S3.BucketName`.  `Expression-bodied member` properties are not useful for configuration options.

You can review the existing code to see how all this works.  Also the following are links to information to understand it.

https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection
https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration

libs/net/dal/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -118,4 +119,23 @@ public static IServiceCollection AddStorageConfig(this IServiceCollection servic
...
+
+    /// <summary>
+    /// Add the S3 configuration to the service collection.
+    /// </summary>
+    /// <param name="services"></param>
+    /// <param name="config"></param>
+    /// <returns></returns>
+    public static IServiceCollection AddS3Config(this IServiceCollection services, IConfiguration config)
+    {
+        services.Configure<S3Options>(config.GetSection("S3"));
```
Fosol, line 131: Dependency injection allows the implementation to control the configuration details.  This means it will populate the S3Options with configuration values found at the `S3` path. 

It is very odd that you are adding the configuration to memory and also creating a singleton.  Not only this, but you are creating a singleton from the copy in memory.

What are you attempting to achieve by this?  Is there a reason you need a singleton?  If so, why are you creating one this way?  Are you mixing configuration options with a helper class?

api/net/Areas/Editor/Controllers/StorageController.cs
```diff
@@ -53,13 +54,16 @@ public class StorageController : ControllerBase
...
+    /// <param name="s3Options"></param>
+    public StorageController(IConnectionHelper connection, IOptions<StorageOptions> storageOptions, IOptions<ApiOptions> apiOptions,
+        IFileReferenceService fileReferenceService, ILogger<StorageController> logger, IOptions<S3Options> s3Options)
     {
         _connection = connection;
         _storageOptions = storageOptions.Value;
         _apiOptions = apiOptions.Value;
         _fileReferenceService = fileReferenceService;
         _logger = logger;
+        _s3Options = s3Options.Value;
```
Fosol, line 66: Here you have chosen to use the scoped version of the configuration options.  Which is the standard and in my opinion the correct way to do this.

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -23,9 +24,11 @@ public FileReferenceService(
         ClaimsPrincipal principal,
         IServiceProvider serviceProvider,
         StorageOptions options,
+        S3Options s3Options,
```
Fosol, line 27: Here you are using the singleton.  Is there a reason for this?

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -225,20 +218,24 @@ public async Task<bool> UploadToS3Async(string s3Key, Stream fileStream)
...
         {
             IQueryable<FileReference> query = this.Context.FileReferences;
 
-            if (updatedBefore.HasValue)
+            if (createdBefore.HasValue)
             {
-                updatedBefore = updatedBefore.Value.ToUniversalTime();
-                query = query.Where(fr => fr.UpdatedOn < updatedBefore.Value);
+                createdBefore = createdBefore.Value.ToUniversalTime();
+                query = query.Where(fr => fr.CreatedOn < createdBefore.Value);
```
Fosol, line 230: Why are we searching for files created before a date?

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -254,5 +251,91 @@ public async Task<IEnumerable<FileReference>> GetFiles(DateTime? updatedBefore =
...
+    }
+
+    public async Task<FileReference> UpdateAsync(FileReference entity)
+    {
+        this.Context.Update(entity);
+        await this.Context.SaveChangesAsync();
+        return entity;
+    }
+
+    public async Task<int> DeleteOldLocalFilesAsync(DateTime beforeDate)
```
Fosol, line 298: I still don't understand the purpose of this method related to the business processes.  What is the requirement that requires this function?

JacobWang-bc: I will remove it as well. Endpoint was removed.

libs/net/dal/Services/FileReferenceService.cs
JacobWang-bc: yes, we get a file before a date

JacobWang-bc: > Why are we searching for files created before a date?

We don't want to move the latest files to the S3 bucket, might be move files before 30 days. I also add a parameter to let us select a date range.

libs/net/dal/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -118,4 +119,23 @@ public static IServiceCollection AddStorageConfig(this IServiceCollection servic
...
+        services.AddSingleton(sp => sp.GetRequiredService<IOptions<S3Options>>().Value);
```
Fosol, line 132: You don't need a singleton of the configuration values.  You may only need a singleton of the helper class.

libs/net/dal/Services/FileReferenceService.cs
Fosol: This is for configuration only.  Use the standard `IOptions<S3Options>` approach.  If you need a singleton for connecting to S3, then you create a helper class and configure it through dependency injection as a singleton.

libs/net/dal/Services/FileReferenceService.cs
Fosol: We should be using the `Content.PublishedOn` date, not the `FileReference.CreatedOn` date.  The `CreatedOn` date is an audit field that tells us when a row in the database was created.  We  are currently importing historical content, which means all the `CreatedOn` dates will be at the same time.  I suppose for testing purposes this will be okay, but the implementation could be problematic if the performance of S3 isn't ideal.

Fosol (approved):

PR #2295, opened 2024-09-26
Fix add to report

Fosol:
An initial fix for add content to report.  A more complex fix needs to be implemented to sync multiple tabs, but this should resolve a few errors.

### DIFF
--- a/app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
+++ b/app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
@@ -55,13 +55,14 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content, onCl
 
         if (!report.instances.length) {
           // Fetch the report because the one in memory has no instances.
-          currentReport = await getReport(report.id);
+          currentReport = await getReport(report.id, true);
 
           // The report does not have an instance, we must create one to add content.
           // This should only occur the first time after a new report is created.
           if (!currentReport?.instances.length) {
             currentReport = await generateReport(report.id, true);
           }
+          instance = currentReport.instances.length ? currentReport.instances[0] : undefined;
         }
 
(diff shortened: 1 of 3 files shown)


app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -55,13 +55,14 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content, onCl
 
         if (!report.instances.length) {
           // Fetch the report because the one in memory has no instances.
-          currentReport = await getReport(report.id);
+          currentReport = await getReport(report.id, true);
```
Fosol, line 58: This was causing issues on the first time adding content to a report.

openshift/kustomize/tekton/base/tasks/git-conditions.yaml
```diff
@@ -72,7 +72,7 @@ spec:
           buildVar="${1/\-/\_}"
           # convert to uppercase
           buildVar="${buildVar^^}"
-          BUILD_${buildVar}=true
+          BUILD=BUILD_${buildVar}=true
```
Fosol, line 75: Fixed bug I introduced.  However, the test pipeline still doesn't work for some reason.

tools/scripts/gen-env-files.sh
```diff
@@ -410,6 +410,7 @@ Auth__Keycloak__Secret={YOU WILL NEED TO GET THIS FROM KEYCLOAK}
 Auth__OIDC__Token=/realms/mmi/protocol/openid-connect/token
 
 Service__ApiUrl=http://host.docker.internal:$portApi/api
+Service__DataLocation=Openshift
```
Fosol, line 413: Makes it easier for new resources to get the syndication service running

PR #2297, opened 2024-09-27
Fix papers filter

Fosol:
Change filter on papers page to include content from yesterday if published after 10PM.

This is only a workaround fix.  I still need to figure out why content was being published in the prior day.


app/editor/src/features/content/papers/constants/defaultPaperFilter.ts
```diff
@@ -10,6 +11,11 @@ import { defaultSort } from './defaultSort';
  * @returns Paper filter.
  */
 export const defaultPaperFilter = (sources: ISourceModel[] = []): IContentListFilter => {
+  // Return stories that are also from yesterday if published after 10PM.
+  const yesterdayAt10PM = moment()
```
Fosol, line 15: Here is the magic.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -729,14 +729,7 @@ private string FixBlacksNewsgroupXml(string filePath, string xml)
             var match = _storyBody.Match(xml);
             if (match.Success)
             {
-                try
-                {
-                    xml = xml.Replace(match.Groups["body"].Value, this.Options.FailedStoryBody);
-                } catch (Exception ex)
-                {
-                    this.Logger.LogError(ex, $"FixBlacksNewsgroupXml - process XML throws exceptions. The XML is: {xml}");
```
Fosol, line 737: We don't need to push this kind of data to the logs.  All we need is the path to the file, then we can review the XML ourselves.

PR #2302, opened 2024-10-01
TNO-3079 report add back button

kkwangsir:
image

add a button to let user go back to the article of ToC


libs/net/dal/Migrations/1.3.5/Up/PostUp/00-ReportTemplate.sql
```diff
@@ -0,0 +1,365 @@
...
+    word-wrap: break-word;
+    overflow-wrap: break-word;
+  }
+
+</style>
+
+<center><div style="margin-bottom: 25px; color:#FFFFF6;"><img src="@($"{SubscriberAppUrl}assets/reports/MMI_logo_white.png")" width="600"></div></center>
+
+@* This is the Do Not Forward disclaimer *@
+<div><p style="background-color: #FFF7E1; color: #876503; text-align: center; font-size: 1em; font-weight: 700; line-height: 1.1em; letter-spacing: 0.08px; padding: 10px 0px; margin: 6px 0px 20px 0px;">Reading this report on your phone? Struggling with navigation? Email <a href="mailto:scott.ryckman@gov.bc.ca">Scott Ryckman</a> to switch to the iPhone friendly version. </p></div>
```
Fosol, line 28: I don't think this is the latest version of the template.

Fosol (approved):
If you received your original copy from production, it should be okay.  I must assume Scott removed his survey at some point since it is no longer there.

PR #2303, opened 2024-10-03
Report fix - use prod tempalte

kkwangsir:
image
Now, the survey is back.


Fosol (changes requested):
In this case we have not applied the prior migration, so you do not need another one.  You can simply update the SQL in the prior.

Fosol (approved):

PR #2305, opened 2024-10-09
TNO-3033 changes in custom report template to add transcript tag and …

nehalaggarwal-bcgov:
…Icon if showing transcript.


JacobWang-bc:
Make sure the template works, and go to `libs/net` to create a migration by `make db-add n=1.3.6`, also create 1.3.6 folder and sql file. you can simply copy the 1.3.5 folder and rename it, then replace the body in sql with your tempalte.
At the end you need to do `db-update` to test it, the custom report can be updated by your SQL. 
here is a example [https://github.com/bcgov/tno/pull/2302](url)

laidaoyu (approved):

JacobWang-bc (approved):

PR #2307, opened 2024-10-10
TNO-2980 Fixed the byline issue

laidaoyu:
- read ipolitics author from rss xml
- updated rss item convert method to read authors' names


JacobWang-bc (approved):
looks good for me

Fosol (approved):

PR #2310, opened 2024-10-18
TNO-3085 Transcription service process S3 Files

laidaoyu:
- Updated transcription service to process s3 files if the content files are on s3
- refactor file reference service, move S3 methods to a new file
- updated transcription service config
- added new email config for sending the notice emails
- updated deploy file


services/net/transcription/TranscriptionService.cs
```diff
@@ -37,10 +42,14 @@ public TranscriptionService(string[] args) : base(args)
...
         base.ConfigureServices(services);
+        // services.AddSingleton(sp => sp.GetRequiredService<IOptions<StorageOptions>>().Value);
         services
             .Configure<TranscriptionOptions>(this.Configuration.GetSection("Service"))
             .AddTransient<IKafkaListener<string, TranscriptRequestModel>, KafkaListener<string, TranscriptRequestModel>>()
-            .AddSingleton<IServiceManager, TranscriptionManager>();
+            .AddSingleton<IServiceManager, TranscriptionManager>()
+            .AddSingleton<IS3StorageService, S3StorageService>()
+            .AddS3Config(this.Configuration.GetSection("S3"))
+            .AddSingleton(sp => sp.GetRequiredService<IOptions<StorageOptions>>().Value);
```
Fosol, line 52: What's the reason for this singleton?

services/net/transcription/TranscriptionService.cs
```diff
@@ -37,10 +42,14 @@ public TranscriptionService(string[] args) : base(args)
     protected override IServiceCollection ConfigureServices(IServiceCollection services)
     {
         base.ConfigureServices(services);
+        // services.AddSingleton(sp => sp.GetRequiredService<IOptions<StorageOptions>>().Value);
         services
             .Configure<TranscriptionOptions>(this.Configuration.GetSection("Service"))
             .AddTransient<IKafkaListener<string, TranscriptRequestModel>, KafkaListener<string, TranscriptRequestModel>>()
-            .AddSingleton<IServiceManager, TranscriptionManager>();
+            .AddSingleton<IServiceManager, TranscriptionManager>()
+            .AddSingleton<IS3StorageService, S3StorageService>()
```
Fosol, line 50: Why a singleton?

services/net/transcription/appsettings.json
```diff
@@ -20,7 +20,8 @@
     "AzureRegion": "canadacentral",
     "AzureCognitiveServicesKey": "",
     "ConvertToAudio": ["mp4", "mov", "m4a"],
-    "SendEmailOnFailure": true
+    "SendEmailOnFailure": true,
+    "NoticeEmailTo": "${Service__NoticeEmailTo}"
```
Fosol, line 24: Why set these to invalid values?

services/net/transcription/appsettings.json
```diff
@@ -66,5 +67,11 @@
       "BatchSize": 16384,
       "LingerMs": 1
     }
+  },
+  "S3": {
+    "BucketName": "${S3__BucketName}",
```
Fosol, line 72: Why set these to invalid values?

laidaoyu: I think these values are from .env file locally, they are valid, I think.

services/net/transcription/appsettings.json
laidaoyu: these values are from .env file locally, they are valid, I think.

services/net/transcription/TranscriptionService.cs
laidaoyu: for transcription service, s3 storage service class can be singleton, that it has s3 settings from config, and the service is connecting to s3 storage like connecting to pvc on openshift.

services/net/transcription/appsettings.json
Fosol: This will set the S3.BucketName to equal `"${S3__BucketName}"`, which is not valid.  It isn't clear what you are doing here.  Are you adding a form of documentation?

services/net/transcription/TNO.Services.Transcription.csproj
```diff
@@ -13,6 +13,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\..\libs\net\services\TNO.Services.csproj" />
     <ProjectReference Include="..\..\..\libs\net\kafka\TNO.Kafka.csproj" />
+    <ProjectReference Include="..\..\..\libs\net\dal\TNO.DAL.csproj" />
```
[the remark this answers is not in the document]
laidaoyu: true, this is because the s3 helper class is in DAL. Transcript service is not really using database directly.
I will move the s3 class to another project.
thanks

api/net/Program.cs
```diff
@@ -81,6 +82,7 @@
 builder.Services.AddOptions<TNO.API.Config.KeycloakOptions>().Bind(config.GetSection("Keycloak"));
 builder.Services.AddSingleton<IAuthorizationHandler, KeycloakClientRoleHandler>();
 builder.Services.AddSingleton<IAuthorizationPolicyProvider, ClientRoleAuthorizationPolicyProvider>();
+builder.Services.AddSingleton<IS3StorageService, S3StorageService>();
```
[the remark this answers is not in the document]
laidaoyu: I will remove.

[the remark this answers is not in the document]
laidaoyu: second thought: the s3 service contains the methods to upload / download files. the config is already set up for api, so the service can be a singleton.

Fosol (approved):
Nice work!

PR #2311, opened 2024-10-22
Update pipeline tooling

Fosol:
Debugging and fixing issues related to the DevOps pipeline and git.  Not certain why we're now getting errors in this area as we've been running the same code for a long time.  It's either an Openshift upgrade issue, or bitnami.  At some point we should fix the bitnami version.

### DIFF
--- a/openshift/kustomize/tekton/base/tasks/git-conditions.yaml
+++ b/openshift/kustomize/tekton/base/tasks/git-conditions.yaml
@@ -53,7 +53,15 @@ spec:
         git config --global --add safe.directory '*'
         git fetch origin master
         git fetch origin dev
+
+        # Delete head file if it exists
+        rm -f .git/refs/heads/dev.lock
+        git status
+        git clean -fd
+        git clean -f
+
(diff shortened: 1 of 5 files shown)


openshift/kustomize/tekton/base/tasks/git-conditions.yaml
```diff
@@ -53,7 +53,15 @@ spec:
         git config --global --add safe.directory '*'
         git fetch origin master
         git fetch origin dev
+
+        # Delete head file if it exists
```
Fosol, line 57: This section helps resolve some of the git issues.  For some reason we cannot access these git files through a different container.  It appears to me that these containers are running as root somehow, which should be impossible in Openshift.

tools/git/DeploymentConfig.yaml
```diff
@@ -0,0 +1,75 @@
+kind: DeploymentConfig
```
Fosol, line 1: Spin up a pod to access the git repo.

PR #2313, opened 2024-10-28
MMI-3095 cleanup junk upload files and prevent stream old file in editor

kkwangsir:
The Editors continuously creates junk files, sometimes uploading 8-10 GB in a week. Currently, my s3 moving system cannot track these files because their fileReference has already been deleted. If we don't control it, disk space will still be consumed and the purge program will be triggered. 
They will be like this
image

Here’s an explanation of my modifications and the root cause of the issue:
When uploading a new file and clicking "publish", in ContentService we delete the previous fileReference during this process, 

    oFileReferences.Except(updated.FileReferences).ForEach(a =>
        {
            context.Entry(a).State = EntityState.Deleted;
        });

we’re unable to remove the associated attachments. However, at this point, we haven’t yet uploaded the new file, so there’s no new fileReference either.

Afterward, in the content controller, there’s a check (`content.FileReferences.Any()`), which obviously will never be true in this scenario. Therefore, we go to enter the else logic that creates a new fileReference and file. So, I updated the method there to clean up junk files here, restricting it to a specific prefix in `UploadCleanUpAsync`
After thorough testing, we’ve been able to prevent the creation of junk files, and now each content item only have one attachment, this process is like replacing the file each time.
now we only have one file for each content
image

Additionally, during front-end testing, I noticed that the old code would immediately stream AV file after uploading, which introduced a delay. After publishing, it would also send a stream request for the old `fileReference` make a file not found message. Now, this has been only stream the new file.

### DIFF
--- a/app/editor/src/features/content/form/hooks/useContentForm.ts
+++ b/app/editor/src/features/content/form/hooks/useContentForm.ts
@@ -209,7 +209,9 @@ export const useContentForm = ({
 
   const setAvStream = React.useCallback(() => {
     if (!!path) {
-      getStream(path)
+      const encodedPath = encodeURIComponent(path);
+
+      getStream(encodedPath)
         .then((result) => {
           setStream(
             !!result
@@ -223,6 +225,14 @@ export const useContentForm = ({
         .catch(() => {});
     }
   }, [getStream, fileReference?.contentType, path]);
+
+  React.useEffect(() => {
+    if (path && fileReference) {
+      setAvStream();
+    } else {
+      setStream(undefined);
+    }
+  }, [fileReference, path, setAvStream]);
 
   const handleSubmit = React.useCallback(
     async (
@@ -247,7 +257,6 @@ export const useContentForm = ({
           // TODO: Make it possible to upload on the initial save instead of a separate request.
           // Upload the file if one has been added.
           const content = await upload(contentResult, values.file);
-          setAvStream();
           result = toForm({ ...content, tonePools: values.tonePools });
         } else if (
           !originalId &&
--- a/app/editor/src/features/content/form/hooks/useContentForm.ts
+++ b/app/editor/src/features/content/form/hooks/useContentForm.ts
@@ -305,7 +305,6 @@ export const useContentForm = ({
       getSeries,
       navigate,
       series,
-      setAvStream,
       updateContent,
       upload,
       userId,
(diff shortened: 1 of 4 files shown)


libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -84,14 +84,53 @@ public async Task<FileReference> UploadAsync(Content content, IFormFile file, st
...
+            var filePrefix = Path.GetFileNameWithoutExtension(model.Path); // get file prefix like "AVN-1085"
+            var directoryPath = Path.GetDirectoryName(path); // get directory path like "/data/AVN"
+
+            if (directoryPath != null)
+            {
+                var filesToDelete = Directory.GetFiles(directoryPath, $"{filePrefix}*", SearchOption.TopDirectoryOnly)
+                .Where(file => Path.GetFileNameWithoutExtension(file) == filePrefix);// find all files with the exact same prefix in the same directory
+
+                // delete all files with the same prefix in the same directory
+                foreach (var file in filesToDelete)
```
Fosol, line 122: What is the desired outcome if one delete fails?  Will this cause issues downstream?

app/editor/src/features/content/form/hooks/useContentForm.ts
```diff
@@ -224,6 +226,14 @@ export const useContentForm = ({
     }
   }, [getStream, fileReference?.contentType, path]);
 
+  React.useEffect(() => {
```
Fosol, line 229: What's the difference between AvStream and Stream?

JacobWang-bc: confirmed, no need to have this, callback could handle changes.

libs/net/dal/Services/FileReferenceService.cs
JacobWang-bc: added a try except block preventing errors from impacting the entire process.

libs/net/dal/Services/FileReferenceService.cs
```diff
@@ -84,14 +84,68 @@ public async Task<FileReference> UploadAsync(Content content, IFormFile file, st
...
+                {
+                    try
+                    {
+                        File.Delete(file);
+                    }
+                    catch (Exception ex)
+                    {
+                        // record error but continue
+                        deleteErrors.Add($"Failed to delete file {file}: {ex.Message}");
+                        Logger.LogError(ex, "Failed to delete file {file}", file);
```
Fosol, line 134: If you're logging errors here why also log a warning below?

kkwangsir: Thanks. I simplified the log.

Fosol (approved):

PR #2317, opened 2024-10-31
Fix reporting service

Fosol:
The reporting service is throwing errors that appear related to code replacing images in the template.  This should resolve the error, but I don't know the implications of the change yet.  I don't understand why there would be the multiple copies of the image in a report.

### DIFF
--- a/libs/net/ches/ChesService.cs
+++ b/libs/net/ches/ChesService.cs
@@ -263,14 +263,14 @@ namespace TNO.Ches
         /// </summary>
         /// <param name="emailBody">email body as html markup - possibly containing base64 encoded images</param>
         /// <returns>dictionary of the images as attachments and the 'key' to use to search and replace them in the markup</returns>
-        private Dictionary<string, AttachmentModel> GetImagesFromEmailBody(string emailBody)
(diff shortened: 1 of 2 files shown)


libs/net/ches/ChesService.cs
```diff
@@ -281,7 +281,8 @@ private Dictionary<string, AttachmentModel> GetImagesFromEmailBody(string emailB
                         Filename = Guid.NewGuid().ToString(),
                         Content = base64Image
                     };
-                    imageDictionary.Add(m.Value, attachment);
+                    if (!imageDictionary.ContainsKey(m.Value))
```
Fosol, line 284: Only add the image to the dictionary if it hasn't already been added.

openshift/kustomize/tekton/base/triggers/template.yaml
```diff
@@ -66,7 +66,7 @@ spec:
         workspaces:
           - name: source
             persistentVolumeClaim:
-              claimName: pipelines
+              claimName: repo
```
Fosol, line 69: Change to the new pvc.

PR #2319, opened 2024-11-03
Config sticky sessions

Fosol:
Openshift routes will do their best to keep sticky sessions.  Nginx will also maintain sticky sessions.

There are some downsides that everyone should be aware of.

- If a pod dies there is nothing to handle the transition, so the client will throw an error
- I was unable to see the cookie to maintain the session, which could indicate another proxy layer involved
- I have limited ability to test and confirm all scenarios
- Nginx is hardcoded configuration to communicate with api-0 and api-1.  Which means we would need to update the config if we want to have more pods available.

The good that comes from this.

- We can have more than one api pod.  If the pod dies everything will automatically switch to the other pod.
- SignalR is getting a reliable connection as long as the api pods don't die

### DIFF
--- a/network/reverse-proxy/editor/nginx.conf
+++ b/network/reverse-proxy/editor/nginx.conf
@@ -1,0 +1,43 @@
+worker_processes auto;
+
+error_log /var/log/nginx/error.log notice;
+pid /tmp/nginx.pid;
+
+events {
+  worker_connections 1024;
+}
+
+http {
+  proxy_temp_path /tmp/proxy_temp;
+  client_body_temp_path /tmp/client_temp;
+  fastcgi_temp_path /tmp/fastcgi_temp;
+  uwsgi_temp_path /tmp/uwsgi_temp;
+  scgi_temp_path /tmp/scgi_temp;
+
+  include /etc/nginx/mime.types;
+  default_type application/octet-stream;
+
+  log_format main '$remote_addr - $remote_user [$time_local] "$request" '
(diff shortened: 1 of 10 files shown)


openshift/kustomize/api/base/route-editor.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
   namespace: default
   annotations:
     haproxy.router.openshift.io/rewrite-target: /
+    haproxy.router.openshift.io/balance: source
```
Fosol, line 9: Source is essentially by ip

openshift/kustomize/api/base/route-editor.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
...
+    haproxy.router.openshift.io/cookie_name: mmi
```
Fosol, line 10: This is supposed to use a cookie for the session, but I can't see it in the client

network/reverse-proxy/editor/nginx.conf
```diff
@@ -0,0 +1,43 @@
...
+  access_log /var/log/nginx/access.log main;
+
+  sendfile on;
+  #tcp_nopush     on;
+
+  keepalive_timeout 65;
+
+  upstream backend {
+    # App server 1
+    server api-1:8080;
```
Fosol, line 33: Hardcoded api pod connection.  It would be nice to configure this to handle when one of the pods is no longer available.

PR #2325, opened 2024-11-08
Update transcript logging

Fosol:
Attempting to figure out why transcripts are failing.

### DIFF
--- a/services/net/transcription/TranscriptionManager.cs
+++ b/services/net/transcription/TranscriptionManager.cs
@@ -8,15 +8,15 @@ using Microsoft.Extensions.Options;
 using TNO.API.Areas.Services.Models.Content;
 using TNO.Ches;
 using TNO.Ches.Configuration;
-using TNO.Services.Transcription.Exceptions;
 using TNO.Core.Exceptions;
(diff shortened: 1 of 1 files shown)


services/net/transcription/TranscriptionManager.cs
```diff
@@ -29,7 +29,7 @@ public class TranscriptionManager : ServiceManager<TranscriptionOptions>
     private CancellationTokenSource? _cancelToken;
     private Task? _consumer;
     private readonly TaskStatus[] _notRunning = new TaskStatus[] { TaskStatus.Canceled, TaskStatus.Faulted, TaskStatus.RanToCompletion };
-    private readonly WorkOrderStatus[] _ignoreWorkOrders = new WorkOrderStatus[] { WorkOrderStatus.Completed, WorkOrderStatus.Cancelled, WorkOrderStatus.Failed };
+    private readonly WorkOrderStatus[] _ignoreWorkOrders = new WorkOrderStatus[] { WorkOrderStatus.Completed };
```
Fosol, line 32: We only want to ignore requests if the workorder is already completed.

PR #2326, opened 2024-11-08
MMI-3071 Improve import TNO data

laidaoyu:
- in content service, use memory cache for the common lists, including source codes, lookups, ingest services, and settings. set the memory expiration time as 30 minutes


services/net/content/ContentManager.cs
```diff
@@ -294,6 +314,88 @@ private string FixHTMLIssues(string? text)
...
+    /// </summary>
+    /// <returns></returns>
+    private IEnumerable<API.Areas.Services.Models.Setting.SettingModel>? GetSettings()
+    {
+        lock (_settingsLock)
+        {       
+            if (!_memoryCache.TryGetValue(SettingsListCacheKey, out SettingsList))
+            {
+                SettingsList = this.Api.GetSettings().Result;
+                _memoryCache.Set(SettingsListCacheKey, SettingsList, TimeSpan.FromMinutes(30));
```
Fosol, line 329: How do you plan to deal with changes made in the application?  With this it would ignore all changes for at least 30 minutes.  One of the areas that might change more often is topics when importing content from TNO.

laidaoyu: you can restart the service, which will clear the memory cache, if you need the changes in effect right away.

Fosol: Not exactly an ideal solution, and one that will result in future oddities when content is not imported due to failures of out of sync data.

PR #2332, opened 2024-11-14
fix DB migrations 1.3.7 

kkwangsir:
replace designer file, update SQL file, please test locally.


Fosol (approved):
The auto generated files look correct.  I've tested locally successfully.

PR #2334, opened 2024-11-19
Add CRON job to restart pods

Fosol:
This CRON job will restart the api-services and reporting-services at 2 AM every night.  This is to reduce the chance of the Out of Memory leak we have.

The restart is a rolling release, so there should be no down time.


openshift/kustomize/services/reporting/base/config-map.yaml
```diff
@@ -21,3 +21,4 @@ data:
   CHES_EMAIL_AUTHORIZED: "true"
   CHARTS_URL: http://charts-api:8080
   RESEND_ON_FAILURE: "true"
+  SEND_TO_ALL_BEFORE_FAILING: "false"
```
Fosol, line 24: We don't want it to try sending to everyone before failing.  It locks the service up for a long time.

PR #2338, opened 2024-11-21
Enable local nuget packages

Fosol:
Oracle and NLP do not currently support .NET 9.0.  I've downgraded those to .NET 8.0.  To do this it requires publishing local nuget packages and versioning.  This means to update these services developers will need to publish new nuget package versions appropriately.

Testing building the NLP Service on OSX isn't going well.  .NET 9.0 failed due to no ARM64 support.  I am not certain if .NET 8.0 will work.

## Summary

- Support local nuget packages
- Downgrade TNO Migration to net8.0
- Downgrade NLP Service to net8.0


NuGet.Config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="local-packages" value="./libs/net/packages" />
```
Fosol, line 4: This should allow all our local environments build the projects with the local packages.  I am not certain this will work in other environments (Github, Openshift) yet.

services/net/contentmigration/TNO.Services.ContentMigration.csproj
```diff
@@ -11,14 +11,15 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\..\libs\net\services\TNO.Services.csproj" />
+    <!-- <ProjectReference Include="..\..\..\libs\net\services\TNO.Services.csproj" /> -->
```
laidaoyu, line 14: why comment this out?

services/net/contentmigration/TNO.Services.ContentMigration.csproj
```diff
@@ -11,14 +11,15 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="LinqKit" Version="1.3.7" />
     <PackageReference Include="System.CommandLine.DragonFruit" Version="0.3.0-alpha.20070.2" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="9.0.0" />
-    <PackageReference Include="Aspire.Oracle.EntityFrameworkCore" Version="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="8.0.1" />
+    <PackageReference Include="Aspire.Oracle.EntityFrameworkCore" Version="8.2.2" />
+    <PackageReference Include="TNO.Services" Version="1.0.0" />
```
laidaoyu, line 22: did you mean version 2.0.0? the above TNO.Services.csproj has update the version to 2.0.0.

NuGet.Config
laidaoyu: why do we change to local nuget? project reference doesn't work in .net 8.0?

services/net/nlp/TNO.Services.NLP.csproj
```diff
@@ -23,6 +23,10 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.ML" Version="4.0.0" />
     <PackageReference Include="OpenNLP.NET" Version="1.9.4.1" />
+    <PackageReference Include="System.Formats.Asn1" Version="8.0.1" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" Version="8.0.1" />
+    <PackageReference Include="System.Text.Json" Version="8.0.5" />
+    <PackageReference Include="TNO.Services" Version="1.0.0" />
```
laidaoyu, line 29: same question as above, why use version 1.0.0?

services/net/contentmigration/TNO.Services.ContentMigration.csproj
Fosol: I left it in because if you want to test updates before building a new package it makes it easy to switch back to the project.  But if it causes confusion I can remove.

services/net/contentmigration/TNO.Services.ContentMigration.csproj
Fosol: No.  This points to version 1.0.0 because that version is .NET8.0.  Version 2.0.0 would be .NET9.0 if we built new packages.

services/net/nlp/TNO.Services.NLP.csproj
Fosol: A project reference will use the latest code changes, and that would result in .NET9.0.

PR #2339, opened 2024-11-21
Add local nuget to docker

Fosol:
The docker images that require local nuget packages need a few teaks so that they work.

### DIFF
--- a/services/net/contentmigration/Dockerfile
+++ b/services/net/contentmigration/Dockerfile
@@ -11,6 +11,7 @@ USER 0
 
 WORKDIR /src
 
+COPY ../../../libs/net/packages /root/.nuget/packages
 COPY services/net/contentmigration services/net/contentmigration
 COPY libs/net libs/net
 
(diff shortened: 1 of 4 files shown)


services/net/contentmigration/Dockerfile
```diff
@@ -11,13 +11,15 @@ USER 0
 
 WORKDIR /src
 
+COPY ../../../libs/net/packages /root/.nuget/packages
```
Fosol, line 14: Copy packages locallly

services/net/contentmigration/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="local-packages" value="/root/.nuget/packages" />
```
Fosol, line 4: Point the nuget source to the local packages so that the build works in a docker image

laidaoyu (approved):

PR #2341, opened 2024-11-21
Fix select content

Fosol:
The page would crash when selecting a lot of content.  This would occur because the implementation would store all the content information into local storage.

This update only stores the content Id values.  However, this does request adding a way to fetch content that was selected but is not currently in memory.

image

### DIFF
--- a/app/subscriber/src/components/content-list/ContentList.tsx
+++ b/app/subscriber/src/components/content-list/ContentList.tsx
@@ -1,11 +1,14 @@
 import { AxiosError } from 'axios';
+import { castToSearchResult } from 'features/utils';
 import { IContentSearchResult } from 'features/utils/interfaces';
+import _ from 'lodash';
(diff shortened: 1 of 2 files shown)


app/subscriber/src/components/content-list/ContentList.tsx
```diff
@@ -101,16 +128,18 @@ export const ContentList: React.FC<IContentListProps> = ({
...
+    let array: number[];
     if (existing) {
       array = JSON.parse(existing);
     } else {
       array = [];
     }
     if (array.length !== selected?.length) {
-      array = selected;
+      array = _.uniq([...array, ...selected.map((i) => i.id)]);
+      localStorage.setItem('selected', JSON.stringify(array));
```
Fosol, line 139: Reduce the number of times writing to local storage to only when it needs to.

PR #2343, opened 2024-11-25
Removed tno services from deployment

laidaoyu:
- removed content-current-service and contentmigration-recent-service


JacobWang-bc (approved):
looks good for me

openshift/kustomize/tekton/base/tasks/deploy-all.yaml
```diff
@@ -117,16 +117,6 @@ spec:
           [env]="dev test prod"
         )
 
-        declare -A COMPONENT5=(
```
Fosol, line 120: You will need to resync these values due to the way shell scripts work.

Fosol (changes requested):

Fosol:
You will also need to update the deployment scripts for deployment

openshift/kustomize/tekton/base/tasks/deploy-all.yaml
laidaoyu: I have updated the script. thanks!

nehalaggarwal-bcgov (commented):
Looks good

PR #2344, opened 2024-11-25
Merge content migration services into one

laidaoyu:
- removed content-migration historic image, AV, online, and print services
- created content-migration historic service, this service should handle image, AV, online and print contents


Fosol (commented):
Looks good.

I think you should also include an update to the Tekton deploy-all task `/openshift/kustomize/tekton/base/tasks/deploy.all.yaml`

laidaoyu:
> deploy.all.yaml

I don't see the code needs to be changed in deploy-all.yaml file

Fosol:
I think I was looking at an old copy.  Looks like you've updated it to `content-historic-service` already.

laidaoyu:
> I think I was looking at an old copy. Looks like you've updated it to `content-historic-service` already.

Thank you!

PR #2345, opened 2024-11-26
Update indexing process

Fosol:
Currently there is a race condition between the Indexing Service and the Folder Collection Service.  Any folder with a filter that relies on an Elasticsearch query can result in missing content the first time it is indexed.

This new workflow will publish the indexing request to a new `folder` topic when the content has successfully been indexed.  The Folder Collection Service will now consume the `folder` topic instead of the `index` topic.

### DIFF
--- a/api/net/Areas/Kafka/Controllers/ProducerController.cs
+++ b/api/net/Areas/Kafka/Controllers/ProducerController.cs
@@ -122,5 +122,24 @@ public class ProducerController : ControllerBase
             StatusCode = 201
         };
     }
+
+    /// <summary>
+    /// Add the index request to the folder queue.
+    /// </summary>
+    /// <param name="model"></param>
+    /// <returns></returns>
+    [HttpPost("folder")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(DeliveryResultModel<IndexRequestModel>), (int)HttpStatusCode.Created)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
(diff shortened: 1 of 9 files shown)


api/net/Areas/Kafka/Controllers/ProducerController.cs
```diff
@@ -122,5 +122,24 @@ public async Task<IActionResult> SendEventAsync([FromBody] EventScheduleRequestM
...
+    /// Add the index request to the folder queue.
+    /// </summary>
+    /// <param name="model"></param>
+    /// <returns></returns>
+    [HttpPost("folder")]
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(DeliveryResultModel<IndexRequestModel>), (int)HttpStatusCode.Created)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Kafka" })]
+    public async Task<IActionResult> SendToFolderAsync([FromBody] IndexRequestModel model)
```
Fosol, line 136: Simple endpoint to essentially pass the indexing message to the folder topic

openshift/kustomize/services/folder-collection/base/config-map.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
 data:
   KAFKA_CLIENT_ID: FolderCollection
   MAX_FAIL_LIMIT: "5"
-  TOPICS: index
+  TOPICS: folder
```
Fosol, line 20: This will need to be manually applied in Openshift

PR #2347, opened 2024-11-26
MMI 1815 option omit BCupdates

kkwangsir:
Add a ES search filter to excludes `BC updates` (headline) and `BC colander` (Body)

### `before switching on the filter`
image

### `checked on`
image

### `Omit BC updates and BC Calendar`
image

### DIFF
--- a/libs/net/dal/Extensions/JsonDocumentExtensions.cs
+++ b/libs/net/dal/Extensions/JsonDocumentExtensions.cs
@@ -254,23 +254,47 @@ public static class JsonDocumentExtensions
             ((JsonObject)boolNode!)["must_not"] = mustNotNode;
         }
 
-        // Add first query_string to exclude "BC Update"
+        // Add a bool query to combine mediaType and content conditions
         ((JsonArray)mustNotNode!).Add(new JsonObject
         {
-            ["query_string"] = new JsonObject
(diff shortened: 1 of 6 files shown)


Fosol (approved):
I think this will work!

libs/net/dal/Extensions/JsonDocumentExtensions.cs
```diff
@@ -222,4 +222,82 @@ public static JsonDocument AddExcludeMediaTypes(this JsonDocument query, IEnumer
...
+
+        // Get or create the must_not array
+        if (!((JsonObject)boolNode!).TryGetPropertyValue("must_not", out var mustNotNode))
+        {
+            mustNotNode = new JsonArray();
+            ((JsonObject)boolNode!)["must_not"] = mustNotNode;
+        }
+
+        // Add a bool query to combine mediaType and content conditions
+        ((JsonArray)mustNotNode!).Add(new JsonObject
```
laidaoyu, line 258: we can use NEST library methods to manipulate ES queries, instead of parsing json like this.

JacobWang-bc: Tested it, this one involves too many changes to use it for now.

PR #2348, opened 2024-11-27
File copy didn't exist. Fix broken pipeline.

kkwangsir:
I saw this from pipeline. 

I think scale filecopy cause this erorr.
```

step-deploy

Stopping extract-quotes
Warning: apps.openshift.io/v1 DeploymentConfig is deprecated in v4.14+, unavailable in v4.10000+
Warning: extensions/v1beta1 Scale is deprecated in v1.2+, unavailable in v1.16+
deploymentconfig.apps.openshift.io/extract-quotes-service scaled
Stopping filecopy
Warning: apps.openshift.io/v1 DeploymentConfig is deprecated in v4.14+, unavailable in v4.10000+
error: no objects passed to scale
```

### DIFF
--- a/openshift/kustomize/tekton/base/tasks/deploy-all.yaml
+++ b/openshift/kustomize/tekton/base/tasks/deploy-all.yaml
@@ -168,16 +168,6 @@ spec:
         )
 
         declare -A COMPONENT10=(
-          [id]="filecopy"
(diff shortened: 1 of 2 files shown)


Fosol (approved):
Thanks for fixing this.  It caused issues in deployment.

PR #2353, opened 2024-11-28
Fix highlight keywords

Fosol:
My fix yesterday to resolve the HTML issue introduced a weird duplication error.  This should be resolved now.

### DIFF
--- a/libs/net/elastic/Extensions/StringExtensions.cs
+++ b/libs/net/elastic/Extensions/StringExtensions.cs
@@ -166,22 +166,21 @@ public static class StringExtensions
     /// <returns></returns>
     public static string MarkKeywords(this string text, IEnumerable<string> keywords, string tagName = "mark")
     {
-        var result = new StringBuilder(text);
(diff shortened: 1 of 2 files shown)


libs/net/elastic/Extensions/StringExtensions.cs
```diff
@@ -166,22 +166,21 @@ public static IEnumerable<string> ExtractKeywords(this string search, string que
...
             {
                 values.Add(match.Groups[i].Value);
             }
-            values = values.Distinct().Where(v => !String.IsNullOrWhiteSpace(v)).ToList(); // Not clear why each match results in multiple groups even when only a single value is found.
+            // Not clear why each match results in multiple groups even when only a single value is found.
+            values = values.Distinct().Where(v => !String.IsNullOrWhiteSpace(v)).ToList();
+            var result = String.Join("", values);
             if (values.Count != 0)
-                return $"<{tagName}>{String.Join("", values)}</{tagName}>";
-            return result.ToString();
```
Fosol, line 181: This was the cause of the duplication.

PR #2357, opened 2024-12-05
Crunchy upgrade + Chart Sorting

Fosol:
Charts now have a way to sort their dataset and index axis grouping.  Also documentation and Infrastructure as Code updated to support upgrade of CrunchyDB.

> Note that when sorting a dataset it is an alphabet sort, which means dates don't work correctly.   Dates do sort correctly in the index axis group by.

## Summary

- Add DB Migration `1.3.9` to update Chart Template
- Published new `tno-core:1.0.0`
- Added sort to chart dataset
- Added sort to chart index axis group by
- Added documentation for Crunchy DB upgrade
- Updated Editor app
- Updated Subscriber app
- Fixed DevOps scripts

## Subscriber Chart Admin

image


libs/net/dal/Migrations/1.3.9/Up/PostUp/00-ChartTemplate.sql
```diff
@@ -0,0 +1,40 @@
+DO $$
+BEGIN
```
Fosol, line 2: Refer to the cshtml file to see what changed.

libs/net/template/ReportExtensions.cs
```diff
@@ -671,13 +681,35 @@ public static string[] GetLabels(IEnumerable<IGrouping<string, ContentModel>> da
     /// <param name="dataset"></param>
     /// <param name="settings"></param>
     /// <param name="sections"></param>
+    /// <param name="isDataset"></param>
     /// <returns></returns>
-    public static string GetLabel(IGrouping<string, ContentModel> dataset, API.Models.Settings.ChartSectionSettingsModel settings, Dictionary<string, ReportSectionModel> sections)
+    public static string GetLabel(IGrouping<string, ContentModel> dataset, API.Models.Settings.ChartSectionSettingsModel settings, Dictionary<string, ReportSectionModel> sections, bool isDataset)
```
Fosol, line 686: I don't like adding this new boolean argument, but it is the simplest solution without needing to rewrite a lot of code.

PR #2368, opened 2024-12-18
MMI-244 Updated CBRA reports

laidaoyu:
- added new cbra report postgres functions
- added migrations
- updated cbra report excel format


Fosol (approved):
I like that you moved a lot of the work to the database.  I wonder though if temporary tables are ideal and performant.  I'm sure they're a lot more performant than pulling all that data to the API.  Good work.

PR #2372, opened 2024-12-23
Fix chart options

Fosol:
For some reason `maintainAspectRatio` was failing to deserialize boolean values.  However, this has not occurred in the past.  I'm not confident this will work either 100%.


libs/net/template/Models/Charts/Options/ChartOptionsModel.cs
```diff
@@ -36,6 +36,6 @@ public class ChartOptionsModel
     /// get/set -
     /// </summary>
     [JsonPropertyName("maintainAspectRatio")]
-    public string? MaintainAspectRatio { get; set; }
+    public bool? MaintainAspectRatio { get; set; }
```
Fosol, line 39: Deserialization was failing here because it was expecting a string instead of a boolean value.

PR #2373, opened 2024-12-26
Fix chart labels

Fosol:
For some reason some charts will display their data labels even when they are turned off.  I've defaulted the labels to `false` in the template and added additional logging.

### DIFF
--- a/libs/net/template/ReportEngine.cs
+++ b/libs/net/template/ReportEngine.cs
@@ -196,7 +196,7 @@ public class ReportEngine : IReportEngine
         if (model.ChartTemplate.SectionSettings.ScaleCalcMax.HasValue)
         {
             // Determine the maximum scale and add the auto max to it.
-            var max = dataModel?.Datasets.Any() == true ? dataModel?.Datasets.Max(ds => ds.Data.Any() ? ds.Data.Max(v => v ?? 0) : 0) : null;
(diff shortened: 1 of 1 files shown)


libs/net/template/ReportEngine.cs
```diff
@@ -196,7 +196,7 @@ public async Task<string> GenerateBase64ImageAsync(
         if (model.ChartTemplate.SectionSettings.ScaleCalcMax.HasValue)
         {
             // Determine the maximum scale and add the auto max to it.
-            var max = dataModel?.Datasets.Any() == true ? dataModel?.Datasets.Max(ds => ds.Data.Any() ? ds.Data.Max(v => v ?? 0) : 0) : null;
+            var max = dataModel?.Datasets.Length > 0 ? dataModel?.Datasets.Max(ds => ds.Data.Length > 0 ? ds.Data.Max(v => v ?? 0) : 0) : null;
```
Fosol, line 199: Linting recommendation

libs/net/template/ReportEngine.cs
```diff
@@ -440,41 +443,33 @@ private JsonDocument MergeChartOptions(API.Models.Settings.ChartTemplateSettings
 
         try
         {
-            // There appears to be no way to modify a value...
             if (sectionJson.TryGetPropertyValue("indexAxis", out JsonNode? sectionIndexAxis))
             {
-                if (defaultJson.ContainsKey("indexAxis"))
-                    defaultJson.Remove("indexAxis");
+                defaultJson.Remove("indexAxis");
```
Fosol, line 448: Linting recommendation

PR #2374, opened 2024-12-27
Fix chart labels

Fosol:
Update the charts api to default to hide data label values.  The problem appears to be how undefined values are handled.  It's difficult to test or duplicate the issue however.

### DIFF
--- a/api/node/src/utils/applyChartJsOptions.ts
+++ b/api/node/src/utils/applyChartJsOptions.ts
@@ -1,0 +1,29 @@
+export const applyChartJsOptions = (values: any) => {
+  const options = { ...values };
+
+  // Ensure default configuration properties exist.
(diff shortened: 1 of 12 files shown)


api/node/src/utils/applyChartJsOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+        if (options.plugins.datalabels.labels.title === undefined) {
+          options.plugins.datalabels.labels.title = {};
+        }
+      }
+    }
+  }
+
+  if (options.plugins.datalabels.labels.title.display === undefined) {
+    // Default this property.
+    options.plugins.datalabels.labels.title.display = false;
```
Fosol, line 20: Make it false because `undefined` results in the labels being displayed.

PR #2375, opened 2024-12-30
MMI-244 Fixed CBRA report issues

laidaoyu:
- fixed the time zone issues that the publish on time in db has time zone info, updated the view to convert it to local time
- updated staff summary time to show in hh:mm:ss format

### DIFF
--- a/libs/net/entities/CBRAReportStaffSummary.cs
+++ b/libs/net/entities/CBRAReportStaffSummary.cs
@@ -13,5 +13,5 @@ public class CBRAReportStaffSummary
     public string Staff { get; set; } = "";
 
     [Column("cbra_hours")]
-    public decimal CbraHours { get; set; }
+    public string CbraHours { get; set; }
 }
--- a/libs/net/dal/Migrations/1.3.14/Down/PostDown/00-CbraReportFunctions.sql
+++ b/libs/net/dal/Migrations/1.3.14/Down/PostDown/00-CbraReportFunctions.sql
@@ -1,0 +1,334 @@
+DO $$
+BEGIN
+
+-- CBRA report view --
+CREATE OR REPLACE VIEW public.vw_cbra_published_contents AS
+SELECT DISTINCT
+  c.id
+  , c.source
+  , c.source_id
+  , c.series_id
+  , c.media_type_id
+  , c.published_on
+  ,	COALESCE(ss.name, '') AS series_name
+FROM content c
+LEFT JOIN source s ON c.source_id = s.id
+LEFT JOIN series ss ON c.series_id = ss.id
+JOIN (
+    SELECT unnest(string_to_array(setting.value::TEXT, ','::TEXT)::INTEGER[]) AS media_type_id
+    FROM setting
+    WHERE setting.name::text = 'CBRAReportMediaTypeIds'::TEXT
+  ) m ON c.media_type_id = m.media_type_id
+WHERE ((ss.id IS NOT NULL AND ss.is_cbra_source = true)
+  OR (ss.id IS NULL AND s.is_cbra_source = true))
+AND c.status = 2;
+
+DROP FUNCTION IF EXISTS public.fn_cbra_report_total_entries(date, date);
+DROP FUNCTION IF EXISTS public.fn_cbra_report_totals_by_broadcaster(date, date);
+DROP FUNCTION IF EXISTS public.fn_cbra_report_totals_by_program(date, date);
+DROP FUNCTION IF EXISTS public.fn_cbra_report_staff_summary(date, date);
+DROP FUNCTION IF EXISTS public.fn_cbra_report_total_excerpts(date, date);
+
+-- Function: fn_cbra_report_total_excerpts
+CREATE OR REPLACE FUNCTION public.fn_cbra_report_total_excerpts (
+    f_from_date DATE
+    , f_to_date DATE
+  )
+  RETURNS TABLE(category CHARACTER varying, totals numeric)
+  LANGUAGE 'plpgsql'
+AS $BODY$
+DECLARE cbra_unqualified integer;
+DECLARE cbra_totals integer;
+DECLARE cbra_action_id integer;
+BEGIN
+
+	CREATE TEMPORARY TABLE IF NOT EXISTS temp_table_content_total_excerpts (
+    content_id NUMERIC
+  );
+	INSERT INTO temp_table_content_total_excerpts
+	SELECT DISTINCT c.id
+	FROM public.vw_cbra_published_contents c
+	WHERE c.published_on >= f_from_date
+	  AND c.published_on <= f_to_date;
+
+	SELECT
+    CAST("value" AS INTEGER) INTO cbra_action_id
+	FROM setting
+	WHERE "name" = 'CBRAUnqualifiedActionId'
+	LIMIT 1;
+
+	SELECT COUNT(*) INTO cbra_unqualified
+	FROM content_action ca
+	JOIN temp_table_content_total_excerpts c ON ca.content_id = c.content_id
+	WHERE ca.action_id = cbra_action_id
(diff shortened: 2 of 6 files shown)


libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -569,12 +569,15 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.Entity("TNO.Entities.CBRAReportStaffSummary", b =>
                 {
-                    b.Property<decimal>("cbra_hours")
-                        .HasColumnType("numeric");
+                    b.Property<string>("CbraHours")
```
Fosol, line 572: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -569,12 +569,15 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                        .IsRequired()
+                        .HasColumnType("text")
+                        .HasColumnName("cbra_hours");
 
-                    b.Property<string>("staff")
+                    b.Property<string>("Staff")
```
Fosol, line 577: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -583,15 +586,18 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.Entity("TNO.Entities.CBRAReportTotalEntries", b =>
                 {
-                    b.Property<string>("dayofweek")
+                    b.Property<string>("DayOfWeek")
```
Fosol, line 589: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -600,12 +606,14 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.Entity("TNO.Entities.CBRAReportTotalExcerpts", b =>
                 {
-                    b.Property<string>("category")
+                    b.Property<string>("Category")
```
Fosol, line 609: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -600,12 +606,14 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
                         .IsRequired()
-                        .HasColumnType("text");
+                        .HasColumnType("text")
+                        .HasColumnName("category");
 
-                    b.Property<int>("totals")
-                        .HasColumnType("integer");
+                    b.Property<int>("Totals")
```
Fosol, line 614: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -614,16 +622,19 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
             modelBuilder.Entity("TNO.Entities.CBRAReportTotalsByBroadcaster", b =>
                 {
-                    b.Property<decimal>("percentageoftotalrunningtime")
-                        .HasColumnType("numeric");
+                    b.Property<decimal>("PercentageOfTotalRunningTime")
+                        .HasColumnType("numeric")
+                        .HasColumnName("percentage_of_total_running_time");
 
-                    b.Property<string>("sourcetype")
+                    b.Property<string>("SourceType")
```
Fosol, line 629: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -614,16 +622,19 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
                         .IsRequired()
-                        .HasColumnType("text");
+                        .HasColumnType("text")
+                        .HasColumnName("source_type");
 
-                    b.Property<string>("totalrunningtime")
+                    b.Property<string>("TotalRunningTime")
```
Fosol, line 634: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -632,27 +643,33 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.Entity("TNO.Entities.CBRAReportTotalsByProgram", b =>
                 {
-                    b.Property<string>("mediatype")
+                    b.Property<string>("MediaType")
```
Fosol, line 646: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -632,27 +643,33 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                        .HasColumnName("media_type");
 
-                    b.Property<decimal>("percentageoftotalrunningtime")
-                        .HasColumnType("numeric");
+                    b.Property<decimal>("PercentageOfTotalRunningTime")
+                        .HasColumnType("numeric")
+                        .HasColumnName("percentage_of_total_running_time");
 
-                    b.Property<string>("series")
+                    b.Property<string>("Series")
```
Fosol, line 655: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -632,27 +643,33 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
                         .IsRequired()
-                        .HasColumnType("text");
+                        .HasColumnType("text")
+                        .HasColumnName("series");
 
-                    b.Property<string>("sourcetype")
+                    b.Property<string>("SourceType")
```
Fosol, line 660: This looks like an issue.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
```diff
@@ -632,27 +643,33 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                        .HasColumnName("source_type");
 
-                    b.Property<decimal>("totalcount")
-                        .HasColumnType("numeric");
+                    b.Property<decimal>("TotalCount")
+                        .HasColumnType("numeric")
+                        .HasColumnName("total_count");
 
-                    b.Property<string>("totalrunningtime")
+                    b.Property<string>("TotalRunningTime")
```
Fosol, line 669: This looks like an issue.

libs/net/entities/CBRAReportStaffSummary.cs
```diff
@@ -13,5 +13,5 @@ public class CBRAReportStaffSummary
     public string Staff { get; set; } = "";
 
     [Column("cbra_hours")]
-    public decimal CbraHours { get; set; }
+    public string CbraHours { get; set; }
```
Fosol, line 16: Are you changing the type because it is actually a string value, or are you changing it because Excel expects a string value.  If the second, then don't change the type.

libs/net/dal/Migrations/TNOContextModelSnapshot.cs
laidaoyu: I will take a look. these names are generated automatically.

libs/net/entities/CBRAReportStaffSummary.cs
laidaoyu: the type is changed in the function, that it returns string not numbers.

Fosol (approved):
I think this is actually correct when comparing to other entities that have been mapped.  It looks like perhaps it was not correctly generated in your previous migration.  I'm not certain why, but most likely an order of events issue.

PR #2376, opened 2025-01-01
MMI-244 Create index on content table

laidaoyu:
- created index on status field and published on field


Fosol (approved):

libs/net/dal/Migrations/1.3.15/Up/PostUp/00-CbraCreateIndex.sql
```diff
@@ -0,0 +1,11 @@
+DO $$
+BEGIN
+
+CREATE INDEX IF NOT EXISTS "IX_content_published_on"
+   ON public.content USING btree
+   (published_on DESC);
+CREATE INDEX IF NOT EXISTS "IX_content_status"
```
Fosol, line 7: You should add this to the ContentConfiguration.cs

laidaoyu: updated. thanks!

PR #2377, opened 2025-01-01
MMI-244 Fixed migration issue

laidaoyu:
- regenerated and updated migration files
- removed manual create index scripts


Fosol (commented):
I thought you already created these indexes.  Won't this fail if your prior scripts in the migration have already run?

Fosol (approved):
I noticed you're changing an existing migration.  Have you tested what happens when the original migration already has been applied to an environment?

laidaoyu:
> I noticed you're changing an existing migration. Have you tested what happens when the original migration already has been applied to an environment?

the migration failed and won't execute in any of the environment, that is why I changed the existing migration.

PR #2378, opened 2025-01-02
MMI-244 Updated cbra functions

laidaoyu:
- updated the cbra functions to convert input dates to utc date time


Fosol (approved):
Hard to determine what was changed

laidaoyu:
> Hard to determine what was changed

the functions from date and to date are converted to utc time, and used in the where queries, instead of converting published on field from content table.

PR #2379, opened 2025-01-07
Replace CrunchyDB

Fosol:
Due to the performance issue of the CrunchyDB v15 in Openshift, I had to migrate to a single Postgres database in Openshift.  I've updated our Infrastructure as Code to reflect this change.  This solution is temporary and puts us at risk due to the single database does not have HA or backups.

### DIFF
--- a/db/postgres/docker/Dockerfile.prod
+++ b/db/postgres/docker/Dockerfile.prod
@@ -1,0 +1,3 @@
+FROM postgres:15.6
+
+COPY ./scripts /docker-entrypoint-initdb.d/
--- a/openshift/kustomize/api-services/base/deploy.yaml
+++ b/openshift/kustomize/api-services/base/deploy.yaml
@@ -70,9 +70,6 @@ spec:
             - name: ingest-storage
               mountPath: /mnt/ingest
           resources:
-            limits:
-              cpu: 500m
-              memory: 1536Mi
             requests:
               cpu: 100m
               memory: 350Mi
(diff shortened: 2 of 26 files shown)


db/postgres/docker/Dockerfile.prod
```diff
@@ -0,0 +1,3 @@
+FROM postgres:15.6
```
Fosol, line 1: To test the version.  We should upgrade our local database at some point.

openshift/kustomize/api-services/base/deploy.yaml
```diff
@@ -140,13 +137,13 @@ spec:
             - name: DB_POSTGRES_USERNAME
               valueFrom:
                 secretKeyRef:
-                  name: crunchy-pguser-admin
-                  key: user
+                  name: database
```
Fosol, line 140: New secret information as I'll be deleting CrunchyDB

openshift/kustomize/elastic/base/statefulset.yaml
```diff
@@ -63,9 +63,6 @@ spec:
             requests:
               cpu: 100m
               memory: 850Mi
-            limits:
```
Fosol, line 66: Removing limits because Openshift has given everyone a ton more headroom.

openshift/kustomize/postgres/standalone/README.md
```diff
@@ -0,0 +1,49 @@
...
+psql -h localhost -p 5432 -U postgres -f full-backup.sql -d tno
+psql -h localhost -p 5432 -U admin -f full-backup.sql -d tno
+psql -h mmi-prod.postgres.database.azure.com -p 5432 -U mmi -f full-backup.sql -d mmi-prod
+```
+
+```bash
+#!/bin/bash
+OUTPUT_FILE="./trace.out"
+
+nohup psql -h localhost -p 5432 -U admin -d tno -a -f full-backup.sql >> ${OUTPUT_FILE} 2>&1 < /dev/null &
```
Fosol, line 35: Run restore as a background process so that a network disconnect doesn't result in having to start over.

openshift/kustomize/postgres/standalone/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+```bash
+#!/bin/sh
+while true
+do
+  > trace.out
```
Fosol, line 42: Clear the log every 10 minutes or the file gets huge.

PR #2383, opened 2025-01-15
Update Syndication Service

Fosol:
This update enables the Syndication Ingest Service to read the Indiginews RSS feed.  Specifically it is able to extract the `content:encoded` and `dc:creator` item values.

I've also updated the `ConvertTextToParagraphs(...)` function because it currently breaks HTML when adding spaces to punctuation.  I however do not know if this change will break prior content being ingested.  If it does we'll need to update it again so that it iterates over each HTML element and only updates appropriate text.  Which will be much more difficult.


libs/net/core/Extensions/StringExtensions.cs
```diff
@@ -8,10 +8,24 @@ namespace TNO.Core.Extensions;
 /// <summary>
 /// StringExtensions static class, provides extension methods for strings.
 /// </summary>
-public static class StringExtensions
+public static partial class StringExtensions
 {
     #region Variables
-    private static readonly Regex TimeZone = new("[A-Z]{3}$");
+    [GeneratedRegex(@"[A-Z]{3}$")]
```
Fosol, line 14: This will compile the regex once and speed up the processing.

libs/net/core/Extensions/StringExtensions.cs
```diff
@@ -272,41 +286,41 @@ public static bool DirectoryExists(this string path, bool ignoreCase = false)
...
 
-            // Add space after the punctuation mark, except there is space/new line/quote after the punctuation mark
-            sanitizedString = Regex.Replace(sanitizedString, @"[\?\.\!\;\,\:](?![\n|""|'|\r|\r\n|\s+])", "$0 ");
-
             // if the input string is not markup, sanitize it
-            if (!tagRegex.IsMatch(sanitizedString))
+            if (!tagRegex.IsMatch(value))
             {
+                // Add space after the punctuation mark, except there is space/new line/quote after the punctuation mark
+                var sanitizedString = PunctuationRegex().Replace(value, "$0 ");
```
Fosol, line 301: We now will only sanitize non-html text.  Which could be problematic.  But currently it will corrupt valid URL values if we use it the way were were.

services/net/syndication/SyndicationAction.cs
```diff
@@ -260,6 +260,18 @@ private SourceContent CreateSourceContent(IngestModel ingest, SyndicationItem it
         var contentType = ingest.IngestType?.ContentType ?? throw new InvalidOperationException($"Ingest '{ingest.Name}' is missing ingest content type.");
         var publishedOn = item.PublishDate.UtcDateTime != DateTime.MinValue ? item.PublishDate.UtcDateTime : (DateTime?)null;
         var uid = Runners.BaseService.GetContentHash(source, title, publishedOn);
+        string? media = null;
+
+        // Extract values from namespaces.  I don't know a better way to do this.
+        foreach (var ext in item.ElementExtensions)
```
Fosol, line 266: I really don't understand why the default Microsoft SyndicationFeed doesn't know how to use namespaces.  I looked for a few hours for some example code and couldn't find any.

PR #2384, opened 2025-01-16
Fix Paper Ingest Service

Fosol:
The Paper Ingest Service would fail if someone entered a duplicate source mapping.  It will now ignore the a bad configuration and log an error.

### DIFF
--- a/services/net/filemonitor/FilemonitorAction.cs
+++ b/services/net/filemonitor/FilemonitorAction.cs
@@ -643,7 +643,7 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
     /// <param name="ingest"></param>
     /// <returns></returns>
     /// <exception cref="ConfigurationException"></exception>
-    private static Dictionary<string, string> GetSourcesForIngest(IngestModel ingest)
+    private Dictionary<string, string> GetSourcesForIngest(IngestModel ingest)
     {
         var sources = new Dictionary<string, string>();
         var sourcesStr = ingest.GetConfigurationValue(Fields.Sources);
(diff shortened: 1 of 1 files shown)


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -653,7 +653,7 @@ private static Dictionary<string, string> GetSourcesForIngest(IngestModel ingest
             foreach (var source in sourcesArr)
             {
                 var pair = source.Split('=');
-                if (pair.Length == 2)
+                if (pair.Length == 2 && !sources.ContainsKey(pair[0]))
```
Fosol, line 656: Don't allow duplicates to be added.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -653,13 +653,13 @@ private static Dictionary<string, string> GetSourcesForIngest(IngestModel ingest
...
                 {
                     sources.Add(pair[0], pair[1]);
                 }
                 else
                 {
-                    throw new ConfigurationException($"Invalid source value in ingest configuration. Source '{source}' for ingest '{ingest.Name}'");
+                    this.Logger.LogError("Invalid source value in ingest configuration. Source '{source}' for ingest '{ingestName}'", source, ingest.Name);
```
Fosol, line 662: Only log an error.

services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -754,7 +754,7 @@ private string FixBlacksNewsgroupXml(string filePath, string xml)
             var contents = sr.ReadToEnd();
             if (String.IsNullOrWhiteSpace(contents))
             {
-                this.Logger.LogWarning("Missing file contents at '{path}'", filePath);
+                this.Logger.LogError("Missing file contents at '{path}'", filePath);
```
Fosol, line 757: Any empty file is an error.

PR #2385, opened 2025-01-16
MMI-3111

kkwangsir:
image

If we can't find the OwnerId, we use the content summary,
which allows the report template to display an external story summary in the ToC.


app/editor/src/features/admin/report-templates/templates/body/CustomReport.cshtml
```diff
@@ -95,7 +95,23 @@
...
                 {
-                  var summary = OwnerId.HasValue ? content.Versions.ContainsKey(OwnerId.Value) ? content.Versions[OwnerId.Value].Summary : "" : "";
+                                  /**
+                * Check if content has versions and contains the current owner version.
+                * If owner version exists, use its summary.
+                * If no version exists, use the content default summary.
+                * If content.Summary is null, use empty string as fallback.
+                */
+
+                  string summary;
```
Fosol, line 105: No need to change, but an extension method would have been probably cleaner.

Fosol (approved):

laidaoyu (approved):

PR #2388, opened 2025-01-20
Fix report issues

Fosol:
Some charts would throw errors if certain boolean values were `undefined`.  They will now default to false in the chart api.

When unlocking a report it will now remove the saved `report.instance.body` value so that changes can be detected and a new body can be generated.

### DIFF
--- a/api/node/src/utils/applyChartJsOptions.ts
+++ b/api/node/src/utils/applyChartJsOptions.ts
@@ -13,11 +13,38 @@ export const applyChartJsOptions = (values: any) => {
         }
       }
     }
+    if (options.plugins.legend === undefined) {
+      options.plugins.legend = {};
+      if (options.plugins.legend.labels === undefined) {
+        options.plugins.legend.labels = {};
+      }
+      if (options.plugins.legend.title === undefined) {
+        options.plugins.legend.title = {};
+      }
+    }
+    if (options.plugins.title === undefined) {
+      options.plugins.title = {};
+    }
+    if (options.plugins.subtitle === undefined) {
+      options.plugins.subtitle = {};
+    }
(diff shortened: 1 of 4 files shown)


api/node/src/utils/applyChartJsOptions.ts
```diff
@@ -13,12 +13,39 @@ export const applyChartJsOptions = (values: any) => {
         }
       }
     }
+    if (options.plugins.legend === undefined) {
```
Fosol, line 16: Handle undefined values and create default objects.

api/node/src/utils/applyChartJsOptions.ts
```diff
@@ -13,12 +13,39 @@ export const applyChartJsOptions = (values: any) => {
...
+    }
   }
 
   if (options.plugins.datalabels.labels.title.display === undefined) {
     // Default this property.
     options.plugins.datalabels.labels.title.display = false;
   }
+  if (options.plugins.legend.title.display === undefined) {
+    // Default this property.
+    options.plugins.legend.title.display = false;
```
Fosol, line 39: Default to false if `undefined`

app/subscriber/src/features/my-reports/edit/ReportEditForm.tsx
```diff
@@ -257,6 +257,7 @@ export const ReportEditForm = React.forwardRef<HTMLDivElement | null, IReportEdi
               if (instance) {
                 const result = await updateReportInstance({
                   ...instance,
+                  body: '',
```
Fosol, line 260: Empty the `report.instance.body` so that it can be regenerated when a report is unlocked.

PR #2391, opened 2025-01-22
MMI-3063 better UX for concurrency errors

kkwangsir:
In the past, it was monitored that there were some concurrency errors in the API during content editing. After a long period of comparison, I found that most of it was actually harmless. 
Some of it was caused by the request for a transcript. I tried automatic merging, but it turned out to be unsafe, as there were too many factors to consider. 
So, I changed it to display the modified fields on the frontend, giving users a prompt, and logging some detailed content on the backend for debugging purposes.

image

### DIFF
--- a/libs/net/dal/Services/ContentService.cs
+++ b/libs/net/dal/Services/ContentService.cs
@@ -299,13 +299,13 @@ public class ContentService : BaseService<Content, long>, IContentService
     {
         // Fields to ignore: system fields 
         var ignoreFields = new HashSet<string> {
-            "Versions",
-            "UpdatedBy",
-            "UpdatedOn",
-            "Version",
-            "CreatedBy",
-            "CreatedOn",
-            "PostedOn",
+            nameof(Content.Versions),
+            nameof(Content.UpdatedBy),
+            nameof(Content.UpdatedOn),
+            nameof(Content.Version),
+            nameof(Content.CreatedBy),
+            nameof(Content.CreatedOn),
+            nameof(Content.PostedOn),
         };
 
         var propertyNames = currentValues.Properties
--- a/libs/net/dal/Services/ContentService.cs
+++ b/libs/net/dal/Services/ContentService.cs
@@ -384,22 +384,18 @@ public class ContentService : BaseService<Content, long>, IContentService
 
             if (databaseValues == null)
             {
-                var errorMessage = "Unable to retrieve latest content. Please refresh the page and try again. If the issue persists, contact support.";
-                this.Logger.LogError(errorMessage);
-                throw new DbUpdateConcurrencyException(errorMessage);
+                var errorMessage = "Unable to retrieve latest content.";
+                this.Logger.LogError(ex, errorMessage);
+                throw new NoContentException(errorMessage, ex);
             }
 
             // Get changed properties and log details
             var changedProperties = GetChangedProperties(currentValues, databaseValues);
             LogFieldChanges(changedProperties);
 
-            var changedFields = string.Join("', '", changedProperties.Keys);
+            var changedFields = string.Join(",", changedProperties.Keys);
 
-            // Throw user-friendly message with better formatting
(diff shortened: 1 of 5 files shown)


libs/net/dal/Services/ContentService.cs
```diff
@@ -288,17 +288,119 @@ public override Content AddAndSave(Content entity)
...
+
+            // Get changed properties and log details
+            var changedProperties = GetChangedProperties(currentValues, databaseValues);
+            LogFieldChanges(changedProperties);
+
+            var changedFields = string.Join("', '", changedProperties.Keys);
+
+            // Throw user-friendly message with better formatting
+            var userMessage = $"Content has been modified by another user. Modified fields: '{changedFields}'. " +
+                            "Please save your changes, refresh page, and reapply your updates.";
```
Fosol, line 400: It's impossible for the user to save their changes before refreshing the page.  It will result in the same error.

A better user experience would be to inform the user which fields are different.  Then they could choose whether they want to override the values in the database with their own.  You just need to be careful here, because many users will just say override.  Ideally we would have a way to determine what was changed, rather than what is different.  For example if the first save changes fields A and B, and the second save changes fields B and C.  Technically we can automatically resolve changes to fields A and C, but need to ask the user what to do with field B because they both edited it.

Fosol: Ideally if there is no difference other than the version number, or the status, it should be able to merge and save correctly without returning an error.

JacobWang-bc:
image

It's been a while since we've found with this issue. Based on the designer's feedback, let's keep the current toast display approach but make it clearer for the user. If there's a content conflict, show the modified fields. For the ContentConflictException error specifically, manually close the toast to give users enough time to read and act.

libs/net/dal/Services/ContentService.cs
```diff
@@ -288,17 +288,118 @@ public override Content AddAndSave(Content entity)
...
+        try
+        {
+            var original = FindById(entity.Id) ?? throw new NoContentException("Entity does not exist");
+            this.Context.UpdateContext(original, entity);
+            if (entity.GuaranteeUid() && original.Uid != entity.Uid) original.Uid = entity.Uid;
+            return base.UpdateAndSave(original);
+        }
+        catch (DbUpdateConcurrencyException ex)
+        {
+            this.Logger.LogInformation("Concurrency conflict detected - ContentId: {ContentId}", entity.Id);
```
nehalaggarwal-bcgov, line 382: Do we need this as error log too?

nehalaggarwal-bcgov (approved):
It looks good to me. Good work on adding comprehensive Exception handling code.

laidaoyu (approved):

libs/net/dal/Services/ContentService.cs
JacobWang-bc: thanks, it is a good idea.

libs/net/dal/Services/ContentService.cs
```diff
@@ -288,17 +288,118 @@ public override Content AddAndSave(Content entity)
...
-    /// <exception cref="NoContentException"></exception>
+    /// <param name="currentValues">Current property values from the entity being saved</param>
+    /// <param name="databaseValues">Current property values from the database</param>
+    /// <returns>Dictionary of changed properties with their current and database values</returns>
+    private Dictionary<string, object> GetChangedProperties(Microsoft.EntityFrameworkCore.ChangeTracking.PropertyValues currentValues,
+    Microsoft.EntityFrameworkCore.ChangeTracking.PropertyValues? databaseValues)
+    {
+        // Fields to ignore: system fields 
+        var ignoreFields = new HashSet<string> {
+            "Versions",
```
Fosol, line 302: When you need the name of a property you should use `nameof(Content.Versions)` so that it is compile time safe.  Magic strings are pain to maintain.

PR #2393, opened 2025-01-23
MMI-3103 Updated content commentary timeout filter

laidaoyu:
- updated content commentary timeout filter that it is now based on the commentary action created on time, which is the time when the user set the content commentary action

### DIFF
--- a/app/subscriber/src/features/commentary/utils/filterCommentaryResults.ts
+++ b/app/subscriber/src/features/commentary/utils/filterCommentaryResults.ts
@@ -16,15 +16,15 @@ export const filterCommentaryResults = (
   return results.filter((content) => {
     const postedCutoff = determineCommentaryTime(commentaryActionId, content.actions);
     const publishedCutoff = moment().subtract(1, 'month');
-    let commentaryCreatedOn = null;
-    // use the commentary action created on time to decide if the content should be shown in the commentary list
+    let commentaryUpdatedOn = null;
+    // use the commentary action updated on time to decide if the content should be shown in the commentary list
     const commentaryAction = content.actions.find((a) => a.id === commentaryActionId);
     if (commentaryAction) {
-      commentaryCreatedOn = commentaryAction.createdOn;
+      commentaryUpdatedOn = commentaryAction.updatedOn;
     }
     return (
-      commentaryCreatedOn &&
-      moment(commentaryCreatedOn) >= moment(postedCutoff) &&
+      commentaryUpdatedOn &&
+      moment(commentaryUpdatedOn) >= moment(postedCutoff) &&
       moment(content.publishedOn) >= publishedCutoff
     );
   });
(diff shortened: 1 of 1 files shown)


app/subscriber/src/features/commentary/utils/filterCommentaryResults.ts
```diff
@@ -16,9 +16,15 @@ export const filterCommentaryResults = (
   return results.filter((content) => {
     const postedCutoff = determineCommentaryTime(commentaryActionId, content.actions);
     const publishedCutoff = moment().subtract(1, 'month');
+    let commentaryCreatedOn = null;
+    // use the commentary action created on time to decide if the content should be shown in the commentary list
+    const commentaryAction = content.actions.find((a) => a.id === commentaryActionId);
+    if (commentaryAction) {
+      commentaryCreatedOn = commentaryAction.createdOn;
```
Fosol, line 23: I don't think it should be the `createdOn`.  That field is an audit field and has little to do with the story.  I suspect you should be using the `publishedOn` value.  The `publishedOn` value is when the story was published to the world.  In paper news it is the time when the story was sent to the newspaper, or the website.  In radio/TV it is the time the show was broadcasted.

I am also confused why the `postedOn` date isn't working.  The `postedOn` datetime should end up being when the  content was published (when the request to publish occurred).  This allows editors to place content that was published in the past but only recently posted in MMI to show up in the commentary and other searches.

laidaoyu: Scott wants the field expires at number of hours "It should be based on the actual time the item was tagged with the Commentary checkbox." The commentary created on is the time when the user tagged the content as commentary.
The published on is the time when the content is published, if it was published at 3am in the morning, the editor (or whoever user) work after 6am won't be able to view the contents from the list.
^ this is what I understand.

Fosol: The `action.createdOn` will not represent when the editor tags the content with the Commentary checkbox.  It will represent the time the action was saved to the database.  Which will essentially always be the same as the `content.createdOn` value.

laidaoyu: On the UI, the commentary list search API request returns a list of contents that has commentary actions (the contents that should have commentary action), and the above code filters the contents based on the commentary action created on time, and show on the UI.

Fosol: If the business requirement is to return content based on when it was added to the commentary action, then your implementation is not correct.  When content is updated by Editors it will automatically add all actions (including commentary).  Which means the `action.createdOn` will have nothing to do with whether the Editor checked the commentary action.  Which means content that is added to commentary after it has been published will not be returned correctly.

laidaoyu: should I use the action.updatedOn then? if the action is updated, this field will be updated only for that action.

Fosol: In theory that would represent the last time it was updated.  However, you would need to test that these values are not constantly updated every time the content is.  Otherwise it will be a moving target.

laidaoyu: I tested, the action updated on would not be changed when the content is saved. so it would be good to use updated on?

Fosol: We can give it a try

Fosol (approved):

nehalaggarwal-bcgov (approved):
Changes looks good.

PR #2397, opened 2025-01-29
MMI-3105 Add unique key and default value for media type select in ContentForm

JacobWang-bc:
Add unique key and default value for media type select in ContentForm.
fix missing mediaTypeId after click "x" button in this field , and causes error message in the form valiadtion.

### DIFF
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -272,6 +272,13 @@ const ContentForm: React.FC<IContentFormProps> = ({
                                   width={FieldSize.Small}
                                   options={mediaTypeOptions}
                                   required
+                                  // Set a unique key for the component
+                                  // This ensures that the component is re-rendered when the mediaTypeId changes
+                                  key={`mediaType-${props.values.mediaTypeId}`}
+                                  // Set the default value for the media type select
+                                  defaultValue={mediaTypeOptions.find(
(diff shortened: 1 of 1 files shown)


Fosol (approved):

app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -272,6 +272,13 @@ const ContentForm: React.FC<IContentFormProps> = ({
                                   width={FieldSize.Small}
                                   options={mediaTypeOptions}
                                   required
+                                  // Set a unique key for the component
+                                  // This ensures that the component is re-rendered when the mediaTypeId changes
+                                  key={`mediaType-${props.values.mediaTypeId}`}
+                                  // Set the default value for the media type select
+                                  defaultValue={mediaTypeOptions.find(
+                                    (mt) => mt.value === props.values.mediaTypeId,
+                                  )}
```
laidaoyu, line 281: from the video Scott posted, I think the problem is in the onChange event of Source. I wonder if the above changes would solve the Source onChange problem?

JacobWang-bc: Actually, I tested about five possibilities earlier and narrowed them down one by one. The issue you mentioned was tested as well, but it wasn't the cause. I also added some async operations in some places and other tests, and they weren't the main reasons either. Ultimately, I'm left with this default solution. It's quite dramatic

PR #2398, opened 2025-01-31
Update CHES library

Fosol:
Updated the CHES library to remove all the linting errors showing up in every PR.  Also updated the configuration for notifications and reports so that they always BCC an email account.

## Summary

- Resolve nullability issues
- BCC all report/notification


libs/net/ches/TNO.Ches.csproj
```diff
@@ -25,6 +25,7 @@
     <PackageReference Include="System.Text.Json" Version="9.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Server.IIS" Version="2.2.6" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
```
Fosol, line 28: Upgraded based on security warning.

PR #2404, opened 2025-02-12
MMI-3100 image tag email alert, migration 1.3.19

nehalaggarwal-bcgov:
Added code to include file reference in the notification alert when it is available


libs/net/dal/Migrations/1.3.19/Up/PostUp/00-Notification_Template.sql
```diff
@@ -0,0 +1,70 @@
...
+  <br />
+}
+<br />
+<div style="font-size: 10px;">
+  This e-mail is a service provided by Government Communications and Public Engagement and is only intended
+  for the original addressee. All content is the copyrighted property of a third party creator of the material.
+  Copying, retransmitting, redistributing, selling, licensing, or emailing the material to any third party or
+  any employee of the Province who is not authorized to access the material is prohibited.
+</div>'
+WHERE "id" = (SELECT CAST(COALESCE("value", '0') AS INT) FROM public."setting" WHERE "name" = 'AlertId')
```
laidaoyu, line 67: there are settings for "BasicAlertTemplateId" and "AlertId".
I think you are updating template table here, maybe you should the template id instead of alert id?

libs/net/dal/Migrations/1.3.19/Up/PostUp/00-Notification_Template.sql
```diff
@@ -0,0 +1,70 @@
...
+{
+  <div>@Content.Byline</div>
+}
+@if (!string.IsNullOrEmpty(@Content.Page))
+{
+  <div>Page #@Content.Page</div>
+}
+@if (!string.IsNullOrEmpty(@Content.ImageContent))
+{
+  var src = $"data:{@Content.ContentType};base64," + Content.ImageContent;
```
laidaoyu, line 44: I wonder if "@Content.ContentType" has the correct data

libs/net/dal/Migrations/1.3.19/Up/PostUp/00-Notification_Template.sql
```diff
@@ -0,0 +1,70 @@
+DO $$
```
Fosol, line 1: These templates are difficult too review.  Please explain what was changed.

We also keep a file in sync so that it's easier to review here `/app/editor/src/features/admin/notifications/constants/defaultRazorTemplate.ts`.  Ideally you should apply your changes to this file too.

laidaoyu: I think the difference is in the block of line 42-46. they are added for showing images.

libs/net/dal/Migrations/1.3.19/Up/PostUp/00-Notification_Template.sql
nehalaggarwal-bcgov: I did not add that change. It was from the previous Notification Template change that I copied from.

libs/net/dal/Migrations/1.3.19/Up/PostUp/00-Notification_Template.sql
nehalaggarwal-bcgov, line 44: This is what I added lines 42-46

nehalaggarwal-bcgov: I took this from the template that sends out AV files. It seemed to work for the Preview in Dev (I tested it in Editor preview in Dev)

libs/net/dal/Migrations/1.3.19/Up/PostUp/00-Notification_Template.sql
nehalaggarwal-bcgov: Yes. I have confirmed that this is the current one in Prod.

laidaoyu (approved):

PR #2405, opened 2025-02-13
Create CHES retry service

Fosol:
This is a new service that will review all reports sent to CHES.  It will make a request to CHES to know if the report was actually emailed out.  If it is stuck in `Accepted` then it will promote the request (ask it to be sent again).  Testing is very difficult due to the only environment where these issues are occurring is in production.  While testing there does appear to be temporary authorization issues when communicating with CHES.  The service should retry, but we'll need to watch the logs.

> Note this does not yet handle notifications that are stuck

## Summary

- Add DB migration 1.3.20
- Add new CHES Retry Service
- Update API
- Update Editor
- Update Subscriber


api/net/Areas/Services/Controllers/AVOverviewController.cs
```diff
@@ -100,7 +100,7 @@ public IActionResult Update([FromBody] AVOverviewInstanceModel model)
     [Produces(MediaTypeNames.Application.Json)]
     [ProducesResponseType(typeof(IEnumerable<UserAVOverviewInstanceModel>), (int)HttpStatusCode.OK)]
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
-    [SwaggerOperation(Tags = new[] { "Report" })]
+    [SwaggerOperation(Tags = new[] { "Evening Overview" })]
```
Fosol, line 103: Correcting labels

PR #2406, opened 2025-02-17
Enhance Executive Summary

Fosol:
Updating the database so that the Executive Summaries in reports can have unlimited text.

## Summary

- Add DB migration 1.3.21

image


libs/net/dal/Configuration/ReportSectionConfiguration.cs
```diff
@@ -4,11 +4,15 @@ namespace TNO.DAL.Configuration;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using TNO.Entities;
 
-public class ReportSectionConfiguration : BaseTypeConfiguration<ReportSection, int>
+public class ReportSectionConfiguration : AuditColumnsConfiguration<ReportSection>
```
Fosol, line 7: Need to override the Description column which requires removing the inheritance and applying the configuration directly on this object.

PR #2409, opened 2025-02-20
Fix accumulate reports

Fosol:
Reports that accumulate are not working correctly in all cases.  Specifically a report that has never been sent before.  This update should resolve that issue.  It's difficult to test all possible regression issues with these changes.

image

### DIFF
--- a/api/net/Areas/Services/Controllers/ReportController.cs
+++ b/api/net/Areas/Services/Controllers/ReportController.cs
@@ -85,6 +85,7 @@ public class ReportController : ControllerBase
     /// Make a request to Elasticsearch for content that matches the specified report 'id' filter.
     /// </summary>
     /// <param name="id"></param>
+    /// <param name="instanceId"></param>
     /// <param name="requestorId"></param>
     /// <returns></returns>
     [HttpGet("{id}/content")]
@@ -93,11 +94,11 @@ public class ReportController : ControllerBase
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [ProducesResponseType((int)HttpStatusCode.NoContent)]
     [SwaggerOperation(Tags = new[] { "Report" })]
-    public async Task<IActionResult> FindContentForReportIdAsync(int id, int? requestorId)
+    public async Task<IActionResult> FindContentForReportIdAsync(int id, long? instanceId, int? requestorId)
     {
         var report = _service.FindById(id);
         if (report == null) return NoContent();
-        var results = await _service.FindContentWithElasticsearchAsync(report, null, requestorId);
+        var results = await _service.FindContentWithElasticsearchAsync(report, instanceId, requestorId);
         return new JsonResult(results);
     }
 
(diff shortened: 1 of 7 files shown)


api/net/Areas/Services/Controllers/ReportController.cs
```diff
@@ -93,11 +94,11 @@ public IActionResult FindById(int id)
     [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
     [ProducesResponseType((int)HttpStatusCode.NoContent)]
     [SwaggerOperation(Tags = new[] { "Report" })]
-    public async Task<IActionResult> FindContentForReportIdAsync(int id, int? requestorId)
+    public async Task<IActionResult> FindContentForReportIdAsync(int id, long? instanceId, int? requestorId)
```
Fosol, line 97: Need to pass the instance in some cases so that it is used to exclude content when making a search for more.

app/subscriber/src/features/my-reports/edit/view/ReportHistoryForm.tsx
```diff
@@ -90,9 +87,15 @@ export const ReportHistoryForm = () => {
 
               <div className="col-3">
                 <Row gap="0.2em">
-                  {formatDate(instance.sentOn ?? '', false)}
-                  <FaRegClock size={18} className="" />
-                  {formatTime(instance.sentOn ?? '')}
+                  {instance.sentOn ? (
```
Fosol, line 90: Only display if there is a date.

app/subscriber/src/features/my-reports/edit/view/ReportHistoryForm.tsx
```diff
@@ -28,10 +28,7 @@ export const ReportHistoryForm = () => {
         setInstances(
           instances.filter(
             (i, index) =>
-              !(
-                index === 0 &&
-                [ReportStatusName.Pending, ReportStatusName.Reopen].includes(i.status)
-              ),
+              ![ReportStatusName.Reopen, ReportStatusName.Completed].includes(i.status) && i.body,
```
Fosol, line 31: Only include if there is a body.

libs/net/dal/Services/ReportInstanceService.cs
```diff
@@ -161,6 +161,13 @@ public ReportInstance Update(ReportInstance entity, bool updateChildren = false)
                 var originalInstanceContent = originalInstanceContents.FirstOrDefault(o => o.ContentId == ric.ContentId && o.SectionName == ric.SectionName);
                 if (originalInstanceContent == null)
                 {
+                    // Fetch the original content from the database to compare.
+                    originalContent = this.Context.Contents
```
Fosol, line 165: This probably won't happen in production, but locally the Elasticsearch data is not always synced with the database.

libs/net/dal/Services/ReportService.cs
```diff
@@ -797,7 +797,8 @@ public IEnumerable<ReportInstance> GetLatestInstances(int id, int? ownerId = nul
             : Array.Empty<long>();
 
         // When an auto report runs it may need to exclude content in the currently unsent report.
-        if (currentInstance != null && currentInstance.SentOn.HasValue == false && reportSettings.Content.ExcludeContentInUnsentReport)
+        if (currentInstance != null && currentInstance.SentOn.HasValue == false
+            && (reportSettings.Content.ExcludeContentInUnsentReport || !reportSettings.Content.ClearOnStartNewReport))
```
Fosol, line 801: Add the current instance content to the exclude list, so that subsequent requests for content knows to look for new content.

PR #2410, opened 2025-02-21
Fix accumulation report

Fosol:
When a report has a section that does not have a folder or filter as a datasource it must keep the content from the prior instance when configured to accumulate.

image


libs/net/dal/Services/ReportService.cs
```diff
@@ -899,6 +899,42 @@ public IEnumerable<ReportInstance> GetLatestInstances(int id, int? ownerId = nul
                 searchResults.Add(section.Name, content);
                 excludeAboveSectionContentIds.AddRange(contentIds.Except(excludeAboveSectionContentIds));
             }
+            else if (!reportSettings.Content.ClearOnStartNewReport
```
Fosol, line 902: This condition wasn't considered in earlier iterations.  When a section doesn't have a datasource it needs to keep the prior data when configured to accumulate.

PR #2411, opened 2025-02-25
Fix update report

Fosol:
When the CHES Retry Service updates the status of a report instance it can result in automatic refreshes to the user currently editing the report.  This updates should only update the report instance status and leave everything else as it was.

Also added additional logging for the CHES Retry Service.  Currently it's difficult to know what it's working on.


app/subscriber/src/features/my-reports/MyReports.tsx
```diff
@@ -46,12 +46,21 @@ export const MyReports: React.FC = () => {
...
     if (report) {
       try {
-        const instance = await getReportInstance(message.id, false);
-        if (instance) {
+        if (message.message === 'status') {
           setReport({
             ...report,
-            instances: report.instances.map((i) => (i.id === message.id ? instance : i)),
+            instances: report.instances.map((i) =>
+              i.id === message.id ? { ...i, status: message.status } : i,
```
Fosol, line 53: Only update status

app/subscriber/src/features/my-reports/edit/ReportEditPage.tsx
```diff
@@ -161,17 +161,26 @@ export const ReportEditPage = () => {
...
-          // If there is a new instance, prepend it.
-          const add = !report.instances.some((i) => i.id === message.id);
+        if (message.message === 'status') {
           setReport({
             ...report,
-            instances: add
-              ? [instance, ...report.instances]
-              : report.instances.map((i) => (i.id === message.id ? instance : i)),
+            instances: report.instances.map((i) =>
+              i.id === message.id ? { ...i, status: message.status } : i,
```
Fosol, line 168: Only update status

PR #2412, opened 2025-02-27
MMI-3042 Fix text

JacobWang-bc:
## Issue 
In the Wysiwyg editor or web interface, certain Canadian French characters (such as é, è, ç) are displayed as question marks (?). This issue persists even though the HTML header is set to UTF-8 encoding. The problem is from character encoding handling during data ingestion through the filemonitor service.

```
        using var sr = new System.IO.StreamReader(filePath);
```
StreamReader first lookup Byte Order Mark (BOM) from files, if news site doesn't put it in the file, StreamReader use UTF8 as default. For content with French word, it may be using Windows-1252, ISO-8859-1, encoding standard, but StreamReader will consider the file use UTF8, normally will see '\uFFFD' in the content, because we could not find those characters.
image

## Modification 
Solution steps:

- Call the ·ReadFileContents· function to begin reading the file content
- Read the file's byte data.
- Call the ·GetFileContent· function to detect the file's encoding format
- Check if the character set was detected.
- convert to string or null or throw exception

## Processing Flow
```mermaid


graph TD;
    A[Start] --> B[ReadFileContents];
    B --> C[Read file bytes];
    C --> D[GetFileContent];
    D --> E{Charset detected?};
    E -- Yes --> F[ReadWithDetectedEncoding];
    E -- No --> G[ReadWithFallbackEncoding];
    F --> H[CheckContent];
    G --> H;
    H --> I[Return file content];
    I --> J[End];

```

### DIFF
--- a/services/net/filemonitor/FilemonitorAction.cs
+++ b/services/net/filemonitor/FilemonitorAction.cs
@@ -193,6 +193,7 @@ public class FileMonitorAction : IngestAction<FileMonitorOptions>
     /// <param name="ingest"></param>
     /// <param name="pathToFile"></param>
     /// <returns></returns>
+
     private async Task CopyFileAsync(SftpClient client, IngestModel ingest, string pathToFile)
     {
         var outputPath = GetOutputPath(ingest);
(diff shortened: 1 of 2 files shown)


services/net/filemonitor/FilemonitorAction.cs
```diff
@@ -209,6 +210,57 @@ private async Task CopyFileAsync(SftpClient client, IngestModel ingest, string p
             await Task.Factory.FromAsync(client.BeginDownloadFile(pathToFile, saveFile), client.EndDownloadFile);
 
             this.Logger.LogDebug("File copied '{file}'", pathToFile);
+            // try multiple encodings
+            var possibleEncodings = new[] {
```
Fosol, line 214: Good work.  This should be a separate utility function.

JacobWang-bc: Thanks. let me change it.

Fosol (approved):
Nice

laidaoyu (approved):

PR #2418, opened 2025-03-06
MMI-3142 AV content prep time issue

laidaoyu:
- updated AV content prep time component:
  * removed arrow icon
  * automatically update Total minutes when user click on publish

### DIFF
--- a/app/editor/src/features/content/form/ContentForm.tsx
+++ b/app/editor/src/features/content/form/ContentForm.tsx
@@ -29,6 +29,7 @@ import {
   hasErrors,
   IOptionItem,
   ISeriesModel,
+  ITimeTrackingModel,
   Modal,
   OptionItem,
   Row,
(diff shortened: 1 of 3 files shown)


app/editor/src/features/content/form/ContentForm.tsx
```diff
@@ -947,6 +954,33 @@ const ContentForm: React.FC<IContentFormProps> = ({
                               !props.values.file)
                           }
                           onClick={() => {
+                            if (contentPrepTime && Number.parseInt(contentPrepTime) > 0) {
```
Fosol, line 957: Does it already validate that the user entered time?

Fosol (approved):

app/editor/src/features/content/form/ContentForm.tsx
laidaoyu: contentPrepTime is a string here. This is to validate the string is a number and is greater than 0. I think user might enter 0 or some invalid number.

nehalaggarwal-bcgov (approved):
Looks great Dao! Just curious, do we need to put any upper limit to the prep time field?

PR #2419, opened 2025-03-06
Auto restart

Fosol:
All services will now restart automatically if configured after a critical failure.  This doesn't solve the issue where a service goes into a Failed status.

The CHES Retry Service will now auto restart after a sequence of failures.

### DIFF
--- a/libs/net/services/Config/ServiceOptions.cs
+++ b/libs/net/services/Config/ServiceOptions.cs
@@ -32,12 +32,12 @@ public class ServiceOptions
     /// <summary>
     /// get/set - Whether to restart a service after a critical sequence of failures.
     /// </summary>
-    public bool AutoRestartAfterFailure { get; set; } = true;
+    public bool AutoRestartAfterCriticalFailure { get; set; } = true;
 
     /// <summary>
     /// get/set - Number of millisecond delay before restarting service after a critical sequence of failures (default: 30 seconds).
     /// </summary>
-    public int RetryAfterFailedDelayMS { get; set; } = 30000;
+    public int RetryAfterCriticalFailureDelayMS { get; set; } = 30000;
 
     /// <summary>
     /// get/set - The URL to the API.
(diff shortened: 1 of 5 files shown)


libs/net/services/Config/ServiceOptions.cs
```diff
@@ -32,12 +32,12 @@ public class ServiceOptions
     /// <summary>
     /// get/set - Whether to restart a service after a critical sequence of failures.
     /// </summary>
-    public bool AutoRestartAfterFailure { get; set; } = true;
+    public bool AutoRestartAfterCriticalFailure { get; set; } = true;
```
Fosol, line 35: Renamed for clarity

libs/net/services/Runners/BaseService.cs
```diff
@@ -191,13 +192,33 @@ public async Task<int> RunAsync()
...
     {
-        using var scope = this.App.Services.CreateScope();
-        var service = scope.ServiceProvider.GetRequiredService<IServiceManager>();
-        await service.RunAsync();
+        var options = this.App.Services.GetRequiredService<IOptions<ServiceOptions>>().Value;
+        do
+        {
+            try
+            {
+                using var scope = this.App.Services.CreateScope();
```
Fosol, line 205: Keep looping if configured to do so, but with a delay.  Since it creates a new instance it will have updated state.  I should change all the services so that they don't need to have this logic within them anymore.  Another day...

services/net/ches-retry/ChesRetryManager.cs
```diff
@@ -66,25 +66,41 @@ public override async Task RunAsync()
         // Always keep looping until an unexpected failure occurs.
         while (true)
         {
-            if (this.State.Status == ServiceStatus.RequestSleep || this.State.Status == ServiceStatus.RequestPause || this.State.Status == ServiceStatus.RequestFailed)
+            try
```
Fosol, line 69: Moved repeated try+catch code to single location.

PR #2426, opened 2025-03-22
MMI-3212 Fix Scheduler Service Issue

Fosol:
It appears that the root issue with the Scheduler Service was that if the user had a stale version of the report and was editing it when an event fired, they would override the last run time of the report.  Which then would cause the Scheduler Service to request the report to regenerate again.  This fix should update the local copy of the report so that the next save does not overwrite the last run time.

### DIFF
--- a/app/subscriber/src/features/my-reports/MyReports.tsx
+++ b/app/subscriber/src/features/my-reports/MyReports.tsx
@@ -22,7 +22,7 @@ import { ReportPreview } from './ReportPreview';
 import * as styled from './styled';
 
 export const MyReports: React.FC = () => {
-  const [{ myReports, reportsFilter }, { findMyReports, deleteReport }] = useReports();
+  const [{ myReports, reportsFilter }, { findMyReports, deleteReport, getReport }] = useReports();
   const [{ getReportInstance }] = useReportInstances();
   const { toggle, isShowing } = useModal();
   const navigate = useNavigate();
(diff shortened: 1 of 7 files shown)


app/subscriber/src/features/my-reports/MyReports.tsx
```diff
@@ -53,6 +53,14 @@ export const MyReports: React.FC = () => {
               i.id === message.id ? { ...i, status: message.status, version: message.version } : i,
             ),
           });
+        } else if (message.message === 'event') {
```
Fosol, line 56: Only update the events.

app/subscriber/src/features/my-reports/edit/ReportEditPage.tsx
```diff
@@ -168,6 +168,14 @@ export const ReportEditPage = () => {
               i.id === message.id ? { ...i, status: message.status, version: message.version } : i,
             ),
           });
+        } else if (message.message === 'event') {
```
Fosol, line 171: Only update the events.

app/subscriber/src/store/hooks/subscriber/useReportSync.ts
```diff
@@ -38,6 +40,21 @@ export const useReportSync = () => {
               );
               return results;
             });
+          } else if (message.message === 'event') {
```
Fosol, line 43: Only update the events.

libs/net/dal/Services/ContentService.cs
```diff
@@ -402,8 +402,8 @@ public override Content UpdateAndSave(Content entity)
                 var contentConflictEx = new ContentConflictException($"FIELDS:{changedFields}");
                 throw new DbUpdateConcurrencyException(contentConflictEx.Message, contentConflictEx);
             }
-            
-            throw ex;
+
+            throw;
```
Fosol, line 406: Fixed warning.

libs/net/reports/TNO.Reports.csproj
```diff
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <PackageReference Include="NPOI" Version="2.7.2" />
+    <PackageReference Include="SixLabors.ImageSharp" Version="3.1.7" />
```
Fosol, line 18: Fixed critical security bug.

PR #2428, opened 2025-03-25
MMI-3213 Fix Ingest Services & Transcript Work Queue

Fosol:
Three changes made in this PR.

1. Default the retry timeout for all Ingest Services.  In theory they should pick backup where they left off after their timeout.
2. Apply proper dependency injection into the transcript workorder endpoint.
3. Rewrote LINQ statement to correct issues with generated SQL.

### DIFF
--- a/api/net/Areas/Editor/Controllers/WorkOrderController.cs
+++ b/api/net/Areas/Editor/Controllers/WorkOrderController.cs
@@ -100,7 +100,7 @@ public class WorkOrderController : ControllerBase
     {
         var uri = new Uri(this.Request.GetDisplayUrl());
         var query = Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery(uri.Query);
-        var result = _workOrderService.FindDistinctWorkOrders(new WorkOrderFilter(query), this._serializerOptions);
+        var result = _workOrderService.FindDistinctWorkOrders(new WorkOrderFilter(query));
         var page = new Paged<WorkOrderModel>(result.Items.Select(i => i), result.Page, result.Quantity, result.Total);
         return new JsonResult(page);
     }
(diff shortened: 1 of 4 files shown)


api/net/Areas/Editor/Controllers/WorkOrderController.cs
```diff
@@ -100,7 +100,7 @@ public IActionResult Find()
     {
         var uri = new Uri(this.Request.GetDisplayUrl());
         var query = Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery(uri.Query);
-        var result = _workOrderService.FindDistinctWorkOrders(new WorkOrderFilter(query), this._serializerOptions);
+        var result = _workOrderService.FindDistinctWorkOrders(new WorkOrderFilter(query));
```
Fosol, line 103: This was implemented incorrectly.  The serialization options should be applied through dependency injection, like it is every where else.

libs/net/dal/Services/Interfaces/IWorkOrderService.cs
```diff
@@ -21,7 +21,7 @@ public interface IWorkOrderService : IBaseService<WorkOrder, long>
     /// </summary>
     /// <param name="filter">Filter to apply to the query.</param>
     /// <returns>A page of work order items that match the filter.</returns>
-    IPaged<WorkOrderModel> FindDistinctWorkOrders(WorkOrderFilter filter, JsonSerializerOptions options);
+    IPaged<WorkOrderModel> FindDistinctWorkOrders(WorkOrderFilter filter);
```
Fosol, line 24: This was implemented incorrectly.  The serialization options should be applied through dependency injection, like it is every where else.

libs/net/dal/Services/WorkOrderService.cs
```diff
@@ -116,98 +125,13 @@ public IPaged<WorkOrder> Find(WorkOrderFilter filter)
     /// </summary>
     /// <param name="filter">Filter to apply to the query.</param>
     /// <returns>A page of work order items that match the filter.</returns>
-    public IPaged<WorkOrderModel> FindDistinctWorkOrders(WorkOrderFilter filter, JsonSerializerOptions options)
+    public IPaged<WorkOrderModel> FindDistinctWorkOrders(WorkOrderFilter filter)
     {
-        var query =
```
Fosol, line 121: This LINQ statement is incorrect.  I don't know why I didn't catch it when Felipe created it.

PR #2433, opened 2025-03-29
MMI-3174 add Byline field in the paper content table

JacobWang-bc:
Update Papers component to adjust column sizes and add Byline field

Because the byline field is not keywords type in Elasticsearch. I don't turn the sortable on.


nehalaggarwal-bcgov (approved):
Looks good!

PR #2438, opened 2025-04-17
MMI-3226 Sridhar code changes

nehalaggarwal-bcgov:
Made changes to fix build errors.


libs/net/dal/TNOContext.cs
```diff
@@ -140,6 +140,10 @@ public IQueryable<CBRAReportTotalEntries> GetCbraReportTotalEntries(DateOnly fro
 
     #endregion
 
+    #region MediaAnalytics
```
nehalaggarwal-bcgov, line 143: Added this change here, rest is Sridhar's code.

Fosol (changes requested):
This requires a DB migration or it will break things.

Fosol (changes requested):
You will need to change your migration version because there is already a 1.3.26.  You also need to do this correctly, or it'll mess up the snapshot.

```bash
# Roll your database back to the last good migration
make db-rollback n=1.3.25

# Remove the last migration code files
make db-rollback

# Reset the TNOContextModelSnapshot.cs to remove all your changes
# Git rebase and get the latest code into your branch

# Now make a new DB migration
make db-add n=1.3.27
make db-update

# Commit the the latest changes
```

libs/net/dal/Migrations/20250430200031_1.3.26.cs
laidaoyu, line 1: in dev branch, there is already 1.3.26 migration, so this PR will have conflicts

PR #2440, opened 2025-04-18
MMI-3224 Add Talkwalker to Reports

Fosol:
I've added a new generic way to include images to reports.

Talkwalker exports their widgets in several formats, one is a PNG.  The PNG is updated automatically every time the report is generated.

The downside to this implementation is that if someone views an old report, it will have a new image.  This is because the URL doesn't change every day.

Another downside to this implementation is that some of the Talkwalker image exports are not great.  We'll need a more advanced solution to handle the different exported widgets (JSON, HTML, etc).

The upside is we can add any type of image to a report.

If we need to keep history, it will require a more advanced solution which saves the image and refers to it instead of the Talkwalker link.

## Summary

- Create new DB migration 1.3.26
- Published tno-core:1.0.10
- Updated Editor app
- Updated Subscriber app
- Updated API

## Subscriber Image Section

image

## Subscriber Preview

image


libs/net/dal/Migrations/1.3.26/Up/PostUp/00-ReportTemplate.sql
```diff
@@ -0,0 +1,374 @@
...
+          if (!string.IsNullOrEmpty(content.ImageContent))
+          {
+            <div>
+              <img style="height:min-content" src="@src" alt="@fileName" />
+            </div>
+          }
+        }
+      }
+    }
+    else if (section.Value.SectionType == ReportSectionType.Image)
```
Fosol, line 327: View the accompanying CustomReport.cshtml file to see this addition.  We can now display Image sections.

app/editor/src/features/admin/report-templates/templates/body/CustomReport.cshtml
```diff
@@ -27,7 +28,7 @@
   <div style="font-size:22px; font-weight:500;"><p>@(Settings.Subject.ShowTodaysDate ? $" {ReportExtensions.GetTodaysDate():dd-MMM-yyyy}" : "")</p></div>
 </a>
 
-@if (Content.Count() == 0 && !ViewOnWebOnly)
+@if (Content.Count() == 0 && !ViewOnWebOnly && !hasImages)
```
Fosol, line 31: If the report has images, it's okay to not have content.

app/editor/src/features/admin/report-templates/templates/body/CustomReport.cshtml
```diff
@@ -319,7 +320,15 @@
         }
       }
     }
-
+    else if (section.Value.SectionType == ReportSectionType.Image)
```
Fosol, line 323: The new image section.

Fosol:
I have merged this to the `master` branch.  This PR will need to be squashed and merged to the `dev` branch before separate release is deployed to PROD.

JacobWang-bc (approved):

PR #2445, opened 2025-05-16
Quotes update

JacobWang-bc:
**Description:**

1.  Update openshift config
2.  Made the code better at handling what the LLM sends back (including empty results).
3.  Improved the logs to see what the LLM process is doing.
4.  Update prompt template


openshift/kustomize/services/extract-quotes/overlays/dev/config-map.yaml
```diff
@@ -19,3 +19,9 @@ data:
   TOPICS: index
   CHES_EMAIL_ENABLED: "true"
   CHES_EMAIL_AUTHORIZED: "true"
+  USE_LLM: "true"
```
Fosol, line 22: Often for overlays you won't have a new config-map.yaml, but rather one that overrides the values in the base.  It isn't critical to do this.

Fosol (approved):

PR #2446, opened 2025-05-16
MMI-3082 vacation mode label changes

nehalaggarwal-bcgov:
Vacation mode toggles when the vacation mode label [X] is pressed.


app/subscriber/src/features/settings/MyAccountSettings.tsx
```diff
@@ -7,6 +7,38 @@ import { ISubscriberUserModel, ToggleButton } from 'tno-core';
...
+    };
+  };
+  const user = createUser();
+
+  try {
+    await updateUser(user, !!impersonate);
+    toast.success('Vacation mode has successfully been updated.');
+  } catch (error) {
+    // Handle the error, if needed
+    console.error('Failed to update user:', error);
```
Fosol, line 38: Any reason a failure is hidden from the user?

PR #2447, opened 2025-05-16
MMI-3249 Update Syndication Service

Fosol:
We can now configure syndication feeds to be applied to different sources.

image


app/editor/src/features/admin/ingests/configurations/Newspaper.tsx
```diff
@@ -28,7 +28,7 @@ export const Newspaper: React.FC = (props) => {
   const sources = lookups.sources.map((s) => new OptionItem(s.name, s.code, !s.isEnabled));
   const source = sources.find((t) => t.value === values.configuration.defaultSource);
 
-  const initialItems = () => {
+  const initialItems = React.useCallback(() => {
```
Fosol, line 31: Functions that are created each execution result in unnecessary renders.

PR #2448, opened 2025-05-20
MMI-3248 Changes to More button visibility condition for CP wire

nehalaggarwal-bcgov:


app/subscriber/src/features/content/view-content/ViewContent.tsx
```diff
@@ -368,7 +368,7 @@ export const ViewContent: React.FC<IViewContentProps> = ({ setActiveContent, act
         <Show visible={isAV && !!content?.summary && isDifferent && !popout}>
           <Col className="summary-container">
             <span>{formattedSummary}</span>
-            <Show visible={!!content?.sourceUrl}>
+            <Show visible={!!content?.sourceUrl && content?.mediaType?.name !== 'CP Wire'}>
```
Fosol, line 371: I'm sure this is good enough for the Jira story, but magic string constants like this are problematic.  Ideally this would be a configurable option based on media type that would hide/show the more button.

Fosol (approved):

PR #2452, opened 2025-06-04
MMI-2918 Advanced Search improvements

laidaoyu:
- added Elasticsearch validate query route
- updated Advanced Search UI to show color coded query string
- added query validate error on the UI


app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
```diff
@@ -358,12 +438,23 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({ onSearch, setSe
...
-                  name="search"
-                  onChange={handleSearchTermsChanged}
-                />
+                <div className="text-area-editor">
+                  <Editor
+                    id="searchText"
+                    value={searchTerms}
+                    onValueChange={handleSearchTermsChanged}
+                    highlight={(code: any) => {
+                      return highlight(code, languages.sql, 'sql');
```
Fosol, line 447: why is it `sql`?

libs/net/dal/Services/ContentService.cs
Fosol, line 1: Need to fix all the linting errors.

Fosol:
I have an updated tno-core library in my PR.

app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
laidaoyu: search query is similar to sql query, which is using AND, OR and ||, &&.

PR #2458, opened 2025-06-13
MMI-3262 Fix user access

Fosol:
When a user has been blocked for specific sources and media types, they should not be able to see them in the Subscriber app.  Additionally, all request to search should filter out blocked content.

## User Management

image

## Advanced Search

image

## Filter by Media Type

image

### DIFF
--- a/app/subscriber/src/features/search-page/components/advanced-search/components/sections/MediaTypeSection.tsx
+++ b/app/subscriber/src/features/search-page/components/advanced-search/components/sections/MediaTypeSection.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { useContent, useLookupOptions } from 'store/hooks';
+import { useApp, useContent, useLookupOptions } from 'store/hooks';
 import { Button, ButtonVariant, Checkbox, FieldSize, Row, Select, Show } from 'tno-core';
 
 import { IFilterDisplayProps } from './IFilterDisplayProps';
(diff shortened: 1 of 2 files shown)


app/subscriber/src/features/search-page/components/advanced-search/components/sections/MediaTypeSection.tsx
```diff
@@ -35,24 +36,26 @@ export const MediaTypeSection: React.FC<IFilterDisplayProps> = ({ displayFilters
...
-                    mediaTypeIds: e.target.checked
-                      ? [...(filter.mediaTypeIds ?? []), +e.target.value] // add it
-                      : filter.mediaTypeIds?.filter((i) => i !== +e.target.value), // remove it
-                  });
-                }}
-              />
-            </div>
-          ))}
+          {mediaTypeOptions
+            .filter((mt) => !userInfo?.mediaTypes.includes(+mt.value!))
```
Fosol, line 40: Remove any media types that are blocked.

libs/net/dal/Services/UserService.cs
```diff
@@ -128,7 +128,10 @@ public IPaged<User> Find(UserFilter filter)
     public User? FindByUsername(string username)
     {
         return this.Context.Users
-            .Where(u => u.Username.ToLower() == username.ToLower()).FirstOrDefault();
+            .Include(u => u.MediaTypes)
```
Fosol, line 131: Include the blocked sources and media types so that they will be added to every search.

PR #2462, opened 2025-06-23
MMI-3171 Updated User forms

laidaoyu:
- updated direct/indirect user forms
- added new fields: organization and reports
- updated api routes and data models


app/editor/src/features/admin/users/UserForm.tsx
```diff
@@ -79,6 +82,58 @@ const UserForm: React.FC = () => {
     } catch {}
   };
 
+  const updateUserIsEnabled = React.useCallback(
```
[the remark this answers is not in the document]
laidaoyu: I will update. thanks!

app/editor/src/features/admin/users/utils/formatDate.ts
```diff
@@ -0,0 +1,17 @@
+import moment from 'moment';
+/**
+ * @param {string | Date} date - The date to format
+ * @param {boolean} includeTime - Whether to include the time in the formatted date
+ * @returns {string} - The formatted date
+ * @description - This function takes a date and returns a formatted date string
+ */
+export const formatDate = (date: string | Date | undefined, includeTime?: boolean): string => {
```
Fosol, line 8: I'm not against another formatDate function, but I think there are already two.

Fosol (approved):

app/editor/src/features/admin/users/utils/formatDate.ts
laidaoyu: This is a special date format only for the pop up modal.

PR #2464, opened 2025-06-25
MMI-3269 Override Report From Email Address

Fosol:
We are currently seeing significant delays in emails being sent out in the morning.  To attempt to work around these issues we will provide a way to configure the FROM address in each report.  This will allow us to spread the number of emails being sent out across multiple addresses, and perhaps allow us to avoid some of the bandwidth restrictions.

If a report has an override it will use it instead of the default configured email address.

## Editor Report Admin

image


libs/net/ches/ChesService.cs
```diff
@@ -228,7 +228,7 @@ public async Task<EmailResponseModel> SendEmailAsync(IEmail email)
         {
             ArgumentNullException.ThrowIfNull(email);
 
-            email.From = this.Options.From ?? email.From;
+            email.From = !String.IsNullOrWhiteSpace(this.Options.From) ? this.Options.From : email.From;
```
Fosol, line 231: Resolves a configuration bug.

libs/net/template/ReportEngine.cs
```diff
@@ -244,7 +243,7 @@ public async Task<string> GenerateBase64ImageAsync(
                 model.ChartTemplate.SectionSettings.ChartType ?? "bar",
                 $"?{width}{height}&options={optionsBase64}"),
             body);
-        this.Logger.LogDebug("Chart generated, chartTemplateId:{templateId} options:{options}", model.ChartTemplate.Id, optionsJson);
+        this.Logger.LogDebug("Chart generated, chartTemplateId:{templateId}", model.ChartTemplate.Id);
```
Fosol, line 246: Removed extra noisy logging.

openshift/kustomize/services/reporting/base/deploy.yaml
```diff
@@ -162,13 +162,13 @@ spec:
                 configMapKeyRef:
                   name: reporting-service
                   key: TOPICS
-
-            # CHES Configuration
-            - name: CHES__From
+            - name: Service__DefaultFrom
```
Fosol, line 165: Need to change this in Openshift or it will continue to use the default.

laidaoyu (approved):

PR #2466, opened 2025-07-10
MMI-2918 Advanced search updates

laidaoyu:
- added warning messages

### DIFF
--- a/app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
+++ b/app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
@@ -8,7 +8,13 @@ import { filterFormat } from 'features/search-page/utils';
 import { highlight, languages } from 'prismjs';
 import React from 'react';
 import { BsCalendarEvent, BsSun } from 'react-icons/bs';
-import { FaCaretSquareDown, FaCheckSquare, FaPlay, FaRegSmile } from 'react-icons/fa';
+import {
+  FaCaretSquareDown,
+  FaCheckSquare,
+  FaExclamationCircle,
+  FaPlay,
+  FaRegSmile,
+} from 'react-icons/fa';
 import { FaSave } from 'react-icons/fa';
 import {
   FaBookmark,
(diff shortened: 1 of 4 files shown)


app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
```diff
@@ -305,7 +321,31 @@ export const AdvancedSearch: React.FC<IAdvancedSearchProps> = ({
   }, [queryValidateResult]);
 
   const ParsedQueryErrorTextArea = React.useCallback(() => {
```
Fosol, line 323: Looks like this should be a separate component?

libs/net/dal/Services/ContentService.cs
```diff
@@ -277,7 +278,34 @@ private void UpdateFieldsArray(JsonArray? fieldsValues, string? arrayFieldNames)
     private void RemoveJsonObjectKey(string key, JsonObject? jobject)
     {
         if (jobject == null) return;
-        if (jobject.ContainsKey(key))
+        if (key.Contains("."))
+        {
+            if (jobject != null)
```
Fosol, line 283: Isn't this already covered by line 280?

libs/net/dal/Services/ContentService.cs
```diff
@@ -277,7 +278,34 @@ private void UpdateFieldsArray(JsonArray? fieldsValues, string? arrayFieldNames)
...
+            {
+                var keys = key.Split(".");
+                if (jobject.ContainsKey(keys[0]))
```
Fosol, line 286: This looks like a perfect time to use a recursive function?

laidaoyu: this method is only used within the class for one case, so there is no need to write a complicate recursive function to handle all the cases.

app/subscriber/src/features/search-page/components/advanced-search/AdvancedSearch.tsx
laidaoyu: it is an embedded component. It is not useful for other components, so I didn't make it as a separate one.

nehalaggarwal-bcgov (approved):
Looks good.

PR #2468, opened 2025-07-11
MMI-3261 Elasticsearch cloud

Fosol:
Updated Indexing Service so that we can run a second cluster of them which will index content to the Elasticsearch cloud cluster.

I've deployed this to PROD so that we can test indexing Elasticsearch in the cloud.

### DIFF
--- a/openshift/kustomize/services/indexing/base/config-map.yaml
+++ b/openshift/kustomize/services/indexing/base/config-map.yaml
@@ -24,3 +24,4 @@ data:
   NOTIFICATION_TOPIC: notify
   CHES_EMAIL_ENABLED: "true"
   CHES_EMAIL_AUTHORIZED: "true"
+  INDEX_ONLY: "false"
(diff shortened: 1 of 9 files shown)


openshift/kustomize/services/indexing/base/config-map.yaml
```diff
@@ -24,3 +24,4 @@ data:
   NOTIFICATION_TOPIC: notify
   CHES_EMAIL_ENABLED: "true"
   CHES_EMAIL_AUTHORIZED: "true"
+  INDEX_ONLY: "false"
```
Fosol, line 27: The Indexing Service can be configured to only index content.  By default it will update content status.

openshift/kustomize/services/indexing/base/deploy.yaml
```diff
@@ -108,17 +108,6 @@ spec:
                   name: services
                   key: KAFKA_BOOTSTRAP_SERVERS
 
-            - name: Kafka__Producer__ClientId
```
Fosol, line 111: It's not a producer

openshift/kustomize/services/indexing/base/deploy.yaml
```diff
@@ -135,6 +124,11 @@ spec:
                 secretKeyRef:
                   name: elastic
                   key: PASSWORD
+            - name: Service__ElasticsearchApiKey
```
Fosol, line 127: We can now support Api Key authentication

PR #2472, opened 2025-07-18
MMI-3270 Updated product subscription

laidaoyu:
- added email format options on the UI for report product subscription
- updated report service to add or update user report


Fosol (approved):
Thanks for closing the hole in the subscription process.

PR #2477, opened 2025-07-28
MMI-3278 Fix Syndication Service

Fosol:
There were critical issues related to fetching sources.  Which was hiding the issues related to CASTANET.

### DIFF
--- a/api/net/Areas/Services/Controllers/SourceController.cs
+++ b/api/net/Areas/Services/Controllers/SourceController.cs
@@ -5,10 +5,10 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
 using Swashbuckle.AspNetCore.Annotations;
 using TNO.API.Areas.Services.Models.Ingest;
+using TNO.API.Filters;
 using TNO.API.Models;
 using TNO.DAL.Services;
 using TNO.Keycloak;
-using TNO.API.Filters;
 
 namespace TNO.API.Areas.Services.Controllers;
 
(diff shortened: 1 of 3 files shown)


api/net/Areas/Services/Controllers/SourceController.cs
```diff
@@ -67,9 +67,10 @@ public IActionResult FindByCode(string code)
     /// Get an array of sources.
     /// </summary>
     /// <returns></returns>
-    [HttpGet]
+    [HttpGet, HttpHead]
```
Fosol, line 70: Head request is required for ETags

api/net/Areas/Services/Controllers/SourceController.cs
```diff
@@ -67,9 +67,10 @@ public IActionResult FindByCode(string code)
...
     [Produces(MediaTypeNames.Application.Json)]
     [ProducesResponseType(typeof(IEnumerable<SourceModel>), (int)HttpStatusCode.OK)]
+    [ProducesResponseType((int)HttpStatusCode.NotModified)]
```
Fosol, line 73: With ETags we return a 304 when the data hasn't changed.

libs/net/core/Http/HttpRequestClient.cs
```diff
@@ -78,20 +79,29 @@ public void Dispose()
...
             var contentType = response.Content.Headers.ContentType;
+            var body = "";
             try
             {
-                if (contentType?.MediaType?.Contains("json", StringComparison.InvariantCultureIgnoreCase) == true)
+                if (contentType?.MediaType?.Contains("json", StringComparison.InvariantCultureIgnoreCase) == true && response.IsSuccessStatusCode)
                     return JsonSerializer.Deserialize<TModel>(responseStream, _serializeOptions);
+                else if (response.IsSuccessStatusCode)
+                    return default;
+                else if (response.StatusCode == HttpStatusCode.NotModified)
```
Fosol, line 89: 304 is not considered successful, but it actually is successful.  We must return null.

libs/net/core/Http/HttpRequestClient.cs
```diff
@@ -78,20 +79,29 @@ public void Dispose()
...
             {
+                // Catch deserialization errors and log the response data.
                 var data = responseStream.ReadAllBytes();
-                var body = Encoding.Default.GetString(data);
+                body = Encoding.Default.GetString(data);
                 _logger.LogError(ex, "Failed to deserialize response: {body}", body);
                 throw;
             }
 
-            throw new HttpClientRequestException(response, $"Response must contain JSON but was '{contentType?.MediaType}'.");
```
Fosol, line 94: This was causing an exception due to attempting to read the body of the response twice.

PR #2478, opened 2025-07-29
Fix Indexer

Fosol:
The Indexer Service needs to include all the content data.

### DIFF
--- a/libs/net/dal/Services/ContentService.cs
+++ b/libs/net/dal/Services/ContentService.cs
@@ -20,7 +20,7 @@ public class ContentService : BaseService<Content, long>, IContentService
     #region Variables
     private readonly ITNOElasticClient _client;
     private static readonly ContentStatus[] _onlyPublished = new[] { ContentStatus.Publish, ContentStatus.Published };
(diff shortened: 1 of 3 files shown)


libs/net/dal/Services/ContentService.cs
```diff
@@ -63,6 +63,14 @@ public IPaged<Content> FindWithDatabase(ContentFilter filter, bool asNoTracking
             .Include(c => c.Owner)
             .AsQueryable();
 
+        if (filter.Includes != null)
+        {
+            foreach (var include in filter.Includes)
+            {
+                query = query.Include(include);
```
Fosol, line 70: Need to include all the relationship tables.

tools/indexer/IndexerManager.cs
```diff
@@ -125,6 +126,18 @@ private async Task<bool> IndexContentAsync()
                 SourceIds = this.Options.SourceIds,
                 Sort = ["id"],
             };
+            contentFilter.Includes = new[]
```
Fosol, line 129: We need to index everything

laidaoyu (approved):

PR #2480, opened 2025-08-01
MMI-3282 Indirect Users

Fosol:
Administrators can now add blocked sources and media types to Indirect users.  When an indirect user has blocked sources and/media types any search for that content will not return anything.  This includes the subscriber app searches and reports.

## Summary

- Updated API
- Updated Editor app

image

### DIFF
--- a/app/editor/src/features/admin/users/UserFormIndirectUser.tsx
+++ b/app/editor/src/features/admin/users/UserFormIndirectUser.tsx
@@ -1,3 +1,4 @@
+import { InputOption } from 'features/content/list-view/components/tool-bar/filter';
 import { useFormikContext } from 'formik';
 import React from 'react';
 import { useLookupOptions } from 'store/hooks';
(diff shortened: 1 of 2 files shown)


app/editor/src/features/admin/users/UserFormIndirectUser.tsx
```diff
@@ -138,6 +141,59 @@ export const UserFormIndirectUser: React.FC<IUserFormProps> = (props) => {
                 }}
               />
             </Section>
+            <Section>
```
Fosol, line 144: Add this to the Indirect User form so that sources and media types can be linked to the Indirect user.

libs/net/dal/Services/UserService.cs
```diff
@@ -134,6 +134,8 @@ public IEnumerable<User> GetUserUpdateHistory()
     public User? FindByUserKey(string key)
     {
         return this.Context.Users
+            .Include(u => u.MediaTypes)
```
Fosol, line 137: When impersonating an indirect user, it will now correctly apply the blocked content.

laidaoyu (approved):

PR #2481, opened 2025-08-05
Mmi 3243

JacobWang-bc:
Introduces a new, optimized method `FindByIdMinimal` to `UserService`. 

This method allows for efficient user retrieval by ID, returning only essential user data without tracking changes, becuase after got the user, we only send it to the kafka, this method can improves performance for sure.

The `ReportController` has been updated to utilize this new `FindByIdMinimal` method, replacing the previous `FindById` to enhance efficiency.

### DIFF
--- a/libs/net/dal/Services/UserService.cs
+++ b/libs/net/dal/Services/UserService.cs
@@ -123,6 +123,19 @@ public class UserService : BaseService<User, int>, IUserService
             .FirstOrDefault(u => u.Id == id);
     }
 
+    /// <summary>
+    /// Find user by ID with minimal data (optimized for performance).
+    /// Only returns basic user information without related entities.
+    /// </summary>
+    /// <param name="id">User ID</param>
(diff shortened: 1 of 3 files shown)


libs/net/dal/Services/UserService.cs
```diff
@@ -123,6 +123,19 @@ public IPaged<User> Find(UserFilter filter)
             .FirstOrDefault(u => u.Id == id);
     }
 
+    /// <summary>
+    /// Find user by ID with minimal data (optimized for performance).
+    /// Only returns basic user information without related entities.
+    /// </summary>
+    /// <param name="id">User ID</param>
+    /// <returns>User with minimal data or null if not found</returns>
+    public User? FindByIdMinimal(int id)
```
Fosol, line 132: Nice to see we're thinking about the impact of each request and what is actually needed in each case.

Fosol (approved):

PR #2485, opened 2025-08-13
MMI-3251 Cache Report Images

Fosol:
Report Image Sections now have an option to cache the images.  This improves the performance of the report when being viewed by subscribers, and it also reduces the number of request to the original source of the images (i.e. Talkwalker).

## Summary

- Add DB migration 1.3.32
- Published tno-core:1.0.20
- Updated API
- Updated Editor
- Updated Subscriber
- Updated Reporting Service
- Fixed security alert in Nuget package
- Updated Infrastructure as Code

### Subscriber Report Image Section - Cache Image

image

### Editor Report Image Section - Cache Image

image

### Report With Images

image


openshift/kustomize/services/reporting/base/deploy.yaml
```diff
@@ -36,13 +36,20 @@ spec:
         part-of: tno
         component: reporting-service
     spec:
+      volumes:
```
Fosol, line 39: Reporting Service needs access to the same share as the API.

PR #2486, opened 2025-08-21
Mmi 1815 fix-- report blank

JacobWang-bc:
Users reported seeing blank report when this option is enabled. However, I don't believe this option is the actual cause. After testing dozens of times, the issue only occurred once, and I've been receiving automated reports for months without seeing any blank reports caused by enabling this option.
I believe this is an error that occurs when Elasticsearch is under heavy load. Therefore, I've added a retry mechanism to give the system time to recover.

### DIFF
--- a/libs/net/dal/Services/ReportService.cs
+++ b/libs/net/dal/Services/ReportService.cs
@@ -885,7 +885,10 @@ public class ReportService : BaseService<Report, int>, IReportService
 
                 // Determine index.
                 var defaultIndex = filterSettings.SearchUnpublished ? _elasticOptions.ContentIndex : _elasticOptions.PublishedIndex;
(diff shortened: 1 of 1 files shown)


libs/net/dal/Services/ReportService.cs
```diff
@@ -940,6 +943,30 @@ public IEnumerable<ReportInstance> GetLatestInstances(int id, int? ownerId = nul
         return searchResults;
     }
 
+    /// <summary>
+    /// Retry Elasticsearch query once after 200ms to handle intermittent failures.
+    /// This helps resolve issues like JSON parsing failures in BC Updates filtering.
```
Fosol, line 948: It's not clear why retrying here would resolve the issue.  I like the retry.

Fosol (approved):

PR #2491, opened 2025-09-09
MMI-3346 Notification Service

Fosol:
The Notification Service is crashing throughout the day.  It's very difficult to debug but there appears to be a relationship with CHES failures.  At certain points it becomes unavailable.  

I believe our code is attempting to read the response body twice.  Which for some reason in .NET is not allowed.  I've modified all the handling of HTTP request errors so that the HttpClientRequestException does not read the response body.

### DIFF
--- a/libs/net/ches/ChesService.cs
+++ b/libs/net/ches/ChesService.cs
@@ -1,6 +1,7 @@
 using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
 using System.Text;
+using System.Text.Json;
 using System.Text.RegularExpressions;
 using System.Web;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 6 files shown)


libs/net/ches/ChesService.cs
```diff
@@ -182,6 +183,11 @@ private async Task<string> SendAsync(string endpoint, HttpMethod method)
                 _logger.LogError(ex, "Failed to send/receive request: {code} {url}.  Error: {error}", ex.StatusCode, url, ex.GetAllMessages());
                 throw new ChesException("Failed to send message to CHES", ex);
             }
+            catch (Exception ex)
```
Fosol, line 186: Handle all exception and wrap them in a ChesException.

libs/net/core/Exceptions/HttpClientRequestException.cs
```diff
@@ -49,7 +49,6 @@ public HttpClientRequestException(string message, Exception innerException, Http
         public HttpClientRequestException(HttpResponseMessage response) : base($"HTTP Request failed [{response?.RequestMessage?.Method}]:[{response?.StatusCode}]:{response?.RequestMessage?.RequestUri}", null, response?.StatusCode)
         {
             this.Response = response ?? throw new ArgumentNullException(nameof(response));
-            this.Data["Body"] = response.Content.ReadAsStringAsync().Result;
```
Fosol, line 52: Don't read the body here.  Let the handler decide how to handle the exception.  A body can only be read once.

libs/net/services/Helpers/ApiService.cs
```diff
@@ -105,6 +105,20 @@ protected async Task<T> RetryRequestAsync<T>(Func<Task<T>> callbackDelegate)
             this.FailureCount = 0;
             return response;
         }
+        catch (HttpClientRequestException ex)
```
Fosol, line 108: If the response is available then extract the response body.

PR #2494, opened 2025-09-22
Mmi 3243 update

AlessiaYChen:


api/net/Areas/Subscriber/Controllers/ReportController.cs
```diff
@@ -476,7 +476,56 @@ public async Task<IActionResult> AddContentToReportAsync(int id, [FromBody] IEnu
...
+    [Produces(MediaTypeNames.Application.Json)]
+    [ProducesResponseType(typeof(ReportModel), (int)HttpStatusCode.OK)]
+    [ProducesResponseType(typeof(ErrorResponseModel), (int)HttpStatusCode.BadRequest)]
+    [SwaggerOperation(Tags = new[] { "Report" })]
+    public async Task<IActionResult> FastAddContentToReportAsync(int id, [FromBody] IEnumerable<ReportInstanceContentModel> content)
+    {
+        var user = _impersonate.GetCurrentUser();
+        var report = _reportService.FindById(id) ?? throw new NoContentException("Report does not exist");
+        if (report.OwnerId != user.Id &&
+            !report.SubscribersManyToMany.Any(s => s.IsSubscribed && s.UserId == user.Id) &&
```
Fosol, line 517: This looks like a logic/bug in the original code.  I don't know why a subscriber of a report should be allowed to add content to someone else's report.

libs/npm/core/src/hooks/api/subscriber/useApiSubscriberReports.ts
```diff
@@ -101,7 +101,7 @@ export const useApiSubscriberReports = (
     },
     addContentToReport: (reportId: number, content: IReportInstanceContentModel[]) => {
       return api.post<never, AxiosResponse<IReportModel>, any>(
-        `/subscriber/reports/${reportId}/content`,
+        `/subscriber/reports/${reportId}/content/fast`,
```
Fosol, line 104: For this to work you will need to publish a new version of the tno-core package and add it to both webapps.

api/net/Areas/Subscriber/Controllers/ReportController.cs
```diff
@@ -476,7 +476,56 @@ public async Task<IActionResult> AddContentToReportAsync(int id, [FromBody] IEnu
...
+            !report.IsPublic) throw new NotAuthorizedException("Not authorized to review this report");
+
+        var addContent = content.Select(c => (Entities.ReportInstanceContent)c);
+        if (addContent.Any())
+        {
+            var result = await _reportService.FastAddContentToReportAsync(id, user.Id, addContent) ?? throw new NoContentException("Report does not exist");
+
+            var instances = _reportService.GetLatestInstances(id, user.Id);
```
Fosol, line 525: I wonder if we really need to return the latest.  Possibly we just return the items that were successfully added.  Then let the UI update its local memory based on this information.  I wouldn't do this until we know whether this update is still too slow.

api/net/Areas/Subscriber/Controllers/ReportController.cs
```diff
@@ -333,7 +334,7 @@ public IActionResult Delete([FromBody] ReportModel model)
     public async Task<IActionResult> Preview(int id)
     {
         var user = _impersonate.GetCurrentUser();
-        var report = _reportService.FindById(id) ?? throw new NoContentException("Report does not exist");
+        var report = _reportService.Find(new ReportFilter { Ids = new[] { id } }, false).FirstOrDefault() ?? throw new NoContentException("Report does not exist");
```
Fosol, line 337: What's the value in this change?

app/subscriber/src/store/hooks/subscriber/useReports.ts
```diff
@@ -8,10 +8,116 @@ import {
   IReportModel,
   IReportResultModel,
   IUserModel,
+  ReportStatusName,
   sortable,
   useApiSubscriberReports,
 } from 'tno-core';
 
+export interface IReportContentMutationModel {
```
Fosol, line 16: Move utility functions and interfaces into separate files.  When everything is in a single file it causes additional merge conflicts down the road.  It's also more difficult to find code when it's buried in a large file.

app/subscriber/src/components/tool-bar/AddToReportMenu.tsx
```diff
@@ -57,44 +55,14 @@ export const AddToReportMenu: React.FC<IAddToReportMenuProps> = ({ content, onCl
...
-        }
-
-        if (!currentReport || !instance) {
-          toast.error('Failed to generate a new report');
-          return;
-        }
-
-        const convertedContent = toInstanceContent(content, instance.id, sectionName, 0);
-        await addContentToReport(currentReport.id, convertedContent);
+        const payload = toInstanceContentPayload(content, 0, sectionName, 0);
```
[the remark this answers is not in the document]
AlessiaYChen: Yes, the logic is moved to the FastAddContentToReportAsync in the ReportController now.

PR #2588, opened 2026-04-13
Integrating Trivy container image vulnerability scanning

rodrigo-barraza-gov:


JacobWang-bc:
Can we let Trivy scan become a reused component?

rodrigo-barraza-gov:
> Can we let Trivy scan become a reused component?

Good suggestion, yes, ended up creating an action.yml as a reusable github action.

JacobWang-bc (approved):

PR #2589, opened 2026-04-20
integrated .trivyignore support into Trivy scan action

rodrigo-barraza-gov:


JacobWang-bc (approved):
Thanks for cleaning up.

PR #2661, opened 2026-09-23
fix(reports): never delete subscriptions on report save, save subscribers intentionally

Fosol:
## Problem

Users were being removed from reports they subscribed to without anyone removing them. The subscription-change alert never fired.

`ReportService.Update` treated the PUT body's subscriber list as authoritative and deleted every `user_report` row not in it. The subscriber app saves the whole report on every content edit from a form snapshot that can be hours old, so any subscription added elsewhere in the meantime (admin, product approval, self-subscribe) was deleted on the next save. The subscriber-area report PUT had no alert hook, and `Watch:IsEnabled` defaults to false.

## Changes

**DAL**
- `ReportService.Update` never deletes a subscription. A subscriber absent from the submitted list gets `isSubscribed=false`. New overload `Update(entity, updateSubscribers: false)` skips subscribers.
- New `ReportService.UpdateSubscribersAndSave(reportId, subscribers)`: merge-only, touches only the rows sent, validates users exist.
- `UserService.UpdateAndSave` applies the same no-delete rule (absent → unsubscribed, present → resubscribed). Admin `UserModel.Reports` lists active subscriptions only so a pending product-request row is not flipped by an admin user save.
- Editor `ReportModel.ToEntity` now carries `isSubscribed`/`format`/`sendTo` instead of resetting them.

**API (subscriber area)**
- `PUT subscriber/reports/{id}` ignores subscribers in the body.
- New `PUT subscriber/reports/{id}/subscribers` takes only changed rows, owner-only, with subscription-change alert coverage (`missingIsRemoved: false` so partial lists don't false-positive).

**Subscriber app**
- Report save diffs subscribers against the loaded report and sends only the changed rows to the new endpoint, then saves the report. Content-tab saves send no subscribers.
- Subscribers tab: trash sets `isSubscribed=false` instead of dropping the row; adding a previously unsubscribed user resubscribes them; select-all/copy/export use active subscribers.

**Tests**
- `ReportServiceTest`: stale list unsubscribes rather than deletes; `updateSubscribers=false` leaves rows untouched; merge endpoint only touches submitted rows; unknown user throws.

Also published new `tno-core: 1.0.39` version.

## Not done (deliberately)

Optimistic concurrency on report saves. The `report` row is only written by UI endpoints, but the report PUT also copies `event_schedule` fields that the scheduler and reporting services write on every run. A version check there would reject the UI save after each scheduled run.

## Verify

`make refresh n=api`, open a report, add/remove a subscriber on the Subscribers tab, save, confirm the `user_report` row flips `is_subscribed` rather than disappearing. Set `Watch__IsEnabled=true` and `Watch__SendTo` to receive alert emails.


.vscode/settings.json
```diff
@@ -31,11 +31,35 @@
   "editor.bracketPairColorization.enabled": false,
   "editor.semanticHighlighting.enabled": false,
   "editor.suggestSelection": "first",
+  // Import order is owned by eslint (simple-import-sort). VS Code's TypeScript "Organize Imports"
+  // sorts by module path ('../..' before '../../../utils'), which fights that rule on every save,
+  // so it is disabled here and eslint's fix runs instead.
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": "explicit" 
+    "source.organizeImports": "never",
```
Fosol: Fixed conflict with repo lint config and local config.

Fosol (commented):
Fixed many linting warnings

